PR #158, opened 2015-02-27
new bots, deduplicater changes

helderfernandes1279:


sebix:
Here's also a XMPP Collector in ca14b7a @dmth, not sure if it's the same.

PR #180, opened 2015-05-24
Simple ip parse

Th4nat0s:
Hello,

I did a generic simple IP list parser, It fill data (src, type, uri) from the parser properties, not a final good solution.. but it works.

I then added Abuse.CH sources.

Nota... I had dos2unixed some files.

Bye


SYNchroACK:
@Th4nat0s thank you very much. Cn you contact me synchroack@gmail.com, I would like to talk with you regarding abuse.ch feeds.

SYNchroACK:
@Th4nat0s let us know if there is any feature that you still need that is not currently implemented. As far I can see, everything is implemented. Thank you!

Cheers

PR #184, opened 2015-05-27
Collector Bot - Microsoft DCU (from an azure blob service)

Phantasus:
Hi,

I wrote a DCU collector for Microsoft DCU. It fetches remote data and feeds it into the IntelMQ
system. You need to specify an azure account to be able to fetch the data, as it does the fetching
through azure and to exact: through its blobservice.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 intelmq.egg-info
 build
 dist
+*.swp
--- a/intelmq/bots/BOTS
+++ b/intelmq/bots/BOTS
@@ -43,6 +43,14 @@
                 "rate_limit": "3600"
             }
         },
+        "DCU": {
+            "module": "intelmq.bots.collectors.dcu.collector",
+            "description": "Microsoft DCU collector fetches report from an azure account",
+            "parameters": {
+                "azure_account_name": "<account_name>",
+                "azure_account_key": "<account_key>"
+            }
+        },
         "VXVault": {
             "module": "intelmq.bots.collectors.url.collector",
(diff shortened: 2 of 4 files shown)


Phantasus:
Wouldn't that be a different collector? Like a file reader. As this is only a collector, not a parser.

SYNchroACK:
@Phantasus , since collector without parser is unusable, is it possible to add commit with the parser?
Thank you @Phantasus

Phantasus:
@SYNchroACK, sure a collector without a parser is almost unuseable (depends on the collector and parser). But I thought that making small pull requests which are itself a working unit is IMHO better than big pull requests which aren't that good testable, like a complete python2 -> python3 transition :-)

SYNchroACK:
**Quick note**

The parameters from configuration files are already been loaded automatically and are available in 
```
self.parameters.<key_from_config>
```

Example:
```
self.parameters.azure_account_key
```

aaronkaplan:
Hi,

So, there are a couple of things that we found out together while testing:
  * please rename the "DCU" section in BOTS to "Microsoft DCU". It's clearer
  * please add the required dependencies to setup.py and/or the User-Guide installation:
    * the azure package (version 0.11.0+) in python
    * cffi (apt-get install python-cffi) - needed for building the  azure python package
    * apt-get install libffi-dev - also needed for building the azure python package

Most importantly, please find out how to give the azure httpclient function the proxy (http_proxy and https_proxy env vars) support (compare with /opt/intelmq/etc/system.conf)

Please also do one single pull request (this one with #201 ) in one go. 

Thanks!

PR #191, opened 2015-05-29
Convert implicit python to explicit python2 calls

Phantasus:
With this fix it should be possible to have two
python installations running.

One python2 and a python3 version as this change is far less intrusive than a complete 
conversion to python3 I think it is more acceptable.

(I was initially unsure if that solution would be a good solution as normally under Debian the
time goes slower than for most other distributions, but as I had recently the pleasure to test
against a Debian installation I saw that even under Debian python2 is a working symlink)


SYNchroACK:
@Phantasus , the python 3 compatibility is not a priority and involves a lot of efforts to test everything at the moment. Sorry. If you want to 'force' in documentation the version 2, my suggestion is use the following command (debian example):

```
update-alternatives --install /usr/bin/python python /usr/bin/python2.7 1
```

Reference: http://linuxconfig.org/how-to-change-from-default-to-alternative-python-version-on-debian-linux

PR #192, opened 2015-05-30
RFC1918 and more plugin

Th4nat0s:
Hi, i did a small plug to filter out all internal lan (basicaly to feed my needs), but also multicast, documentation, locallink etc... in both ipv4 and 6...

You may select the properties (a list of fields) to check for sanitisation or if you want, you could drop the whole record if found.

  "rfc1918-expert": {
>->->-"fields": " reported_source_ip ,  ip",
>->->-"drop": "yes"
    },

Open to ideas,


sebix:
Does `intelmq/bots/utils.py` only include changed line endings? Without rebase on current upstream branch (see my other comment), it's hard to see the changes.

sebix:
@SYNchroACK do you want to have this bot? I can cherry-pick the commit and adapt it to current master.

SYNchroACK:
@sebix yes, make sense to include. So, yes, adapt it ;)

PR #194, opened 2015-06-01
TCP Collector

helderfernandes1279:
IntelMQ TCP collector bot added

### DIFF
--- a/intelmq/bots/BOTS
+++ b/intelmq/bots/BOTS
@@ -19,6 +19,15 @@
                 "rate_limit": "0"
             }
         },
+	"IntelMQ TCP collector": {
+            "module": "intelmq.bots.collectors.tcp.collector",
+            "description": "IntelMQ TCP collector is Client/Server bot, this bot was built to receive events from remote clients.",
+            "parameters": {
+                "rate_limit": "3600",
+                "host": "0.0.0.0",
+		"port": "8001"
+            }
+	},	
         "Arbor": {
             "module": "intelmq.bots.collectors.url.collector",
             "description": "Arbor Collector is the bot responsible to get the report from source of information.",
--- a/intelmq/bots/collectors/tcp/collector.py
+++ b/intelmq/bots/collectors/tcp/collector.py
@@ -1,0 +1,55 @@
+from intelmq.lib.bot import Bot, sys
+import socket,base64
+
+class IntelMQCollectorBot(Bot): 
+
+    tcpserver=None
+    
+    def process(self):  
+  
+        if self.tcpserver==None:
+	   self.tcpserver=TCPSock(self.parameters.host,self.parameters.port,self.send_message,self.logger)	   
+	   self.tcpserver.start()		
+
+    def killbot(self):
+	self.tcpserver.stop()	
+	self.logger.info("Disconnected") 	
+	
+	
+
+class TCPSock():
+            	    
+    def __init__(self,host,port,object_send_msg,object_logger):
+	self.collectors=object_send_msg	
+	self.collectorl=object_logger
+	self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+	server_address = (host,port)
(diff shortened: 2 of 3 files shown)


SYNchroACK:
```
from intelmq.lib.bot import Bot, sys
import socket
import base64


class TCPCollector(Bot):

    def init(self):
        self.logger.info("Starting up on %s port %s" % (self.parameters.host, self.parameters.port))
        self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        self.sock.bind((self.parameters.host, self.parameters.port))
        self.logger.info("Server Started")
        self.sock.listen()

    def process(self):

        while True:
            connection, client_address = self.sock.accept()

            try:
                self.logger.info("Connection from %s" % client_address[0])
                total_data = ''

                while True:
                    data = connection.recv(1024)
                    if not data:
                        break
                    total_data += data

                event = unicode(base64.b64decode(total_data))
                self.send_message(event)
                self.logger.info("Event received")

            finally:
                # Clean up the connection
                connection.close()

    def stop(self):
        self.sock.stop()
        self.logger.info("Disconnected")
        sys.exit(1) # FIXME: need to change bot.py to support the overwrite


if __name__ == "__main__":
    bot = TCPCollector(sys.argv[1])
    bot.start()
```

@helderfernandes1279 , what is the difference between your code and this code? In both code it seems that for each message received a socket is being open, isn't it?

helderfernandes1279:
No,in the code I submitted, the var tcpserver is checked, and only if the var tcpserver == None the socket is created, since this object (IntellMQCollectorBot) is created once at the time the bot starts, it's possible to check if the socket exists or not, by checking if the var tcpserver has a reference of a TCPsocket object.

SYNchroACK:
Yes, you're right.
Regarding the class TCPSock class, why is it necessary?
I'm asking all these questions to improve the code and keep the usual structure.

Thank you very much @helderfernandes1279

helderfernandes1279:
The only reason I wrote the code that way was to avoid calling the method process with a forever loop several times, since the method process is always called after the rate_limit time.
I'm Ok with the questions :) Some times these questions improve our way of thinking

helderfernandes1279:
About the TCPSock class, I wrote the code that way because I think it's easier to to add new features if necessary without changing the bot object, and it's easier to understand.

SYNchroACK:
New code:

```python
from intelmq.lib.bot import Bot, sys
import socket
import base64


class TCPCollector(Bot):

    def init(self):
        self.logger.info("Starting up on %s port %s" % (self.parameters.host, self.parameters.port))
        self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        self.sock.bind((self.parameters.host, self.parameters.port))
        self.logger.info("Server Started")
        self.sock.listen()
        self.connection, self.client_address = self.sock.accept()

    def process(self):

        try:
            self.logger.info("Connection from %s" % self.client_address[0])
            total_data = ''

            while True:
                data = self.connection.recv(1024)
                if not data:
                    break
                total_data += data

            event = unicode(base64.b64decode(total_data))
            self.send_message(event)
            self.logger.info("Event received")

        finally:
            # Clean up the connection
            self.connection.close()

    def stop(self):
        self.sock.stop()
        self.logger.info("Disconnected")
        sys.exit(1) # FIXME: need to change bot.py to support the overwrite


if __name__ == "__main__":
    bot = TCPCollector(sys.argv[1])
    bot.start()
```

helderfernandes1279:
Hi Tomás, I did my best with the changes, I hope its Ok now.
Regards

sebix:
@helderfernandes1279 @SYNchroACK  What are possible usecases of this bot?

helderfernandes1279:
Hi there,

This bot was initially developed to collect events from other instances of
intelmq, there's also an output bot to send events to this type of
collectors


2016-02-16 16:40 GMT+00:00 Sebastian <notifications@github.com>:

PR #201, opened 2015-06-02
Parser Bot - Microsoft DCU

Phantasus:
This adds code for the DCU parser, which parses output of the
dcu-collector and converts it into IntelMQ internal format.

Some fields to which no reasonable mapping was found or which
contained not yet useable information, were simply ignored.

It also breaks the ontology of having a lowercase "malware" field.
Why did I do that? Because the dcu format contains a threat code,
which is mixed case, to include that information I just assigned it.

So let me know if we should just convert that thing to lowercase
or throw it completly away.

### DIFF
--- a/intelmq/bots/BOTS
+++ b/intelmq/bots/BOTS
@@ -180,6 +180,12 @@
                 "rate_limit": "0"
             }
         },
+        "DCU": {
+            "module": "intelmq.bots.parsers.dcu.parser",
+            "description": "Parses dcu output created by dcu collector into IntelMQ format",
+            "parameters": {
+            }
+        },
         "VXVault": {
             "module": "intelmq.bots.parsers.vxvault.parser",
             "description": "VXVault Parser is the bot responsible to parse the report and sanitize the information.",
--- a/intelmq/bots/parsers/dcu/lib.py
+++ b/intelmq/bots/parsers/dcu/lib.py
@@ -1,0 +1,56 @@
+from datetime import datetime, timedelta
+from dateutil.tz import tzutc
+import types
+
+def convert_windows_timestamp(windows_ts):
+    """Convert a windows file utc timestamp to a datetime object"""
+    us = int(windows_ts)/10
+    dt = datetime(1601,1,1, tzinfo=tzutc()) + timedelta(microseconds=us)
+    return dt
+
+def filetimeutc_as_source_time(ts):
+    """Convert the given FileTimeUTC timestamp to a UTC datetime object """
+    parts = ("reported_source_time", str(convert_windows_timestamp(ts)))
+    return parts
+
+def convert_malware_field(malware_name):
+    return ("reported_malware", malware_name.lower())
+
+INTELMQ_TX = {"SourcedFrom": None,  # TBD
+              "FileTimeUtc": filetimeutc_as_source_time,
+              "Botnet": convert_malware_field, 
+              "SourceIp": "reported_source_ip", 
+              "SourcePort": "reported_source_port", 
+              "SourceIpAsnNr": "reported_source_asn", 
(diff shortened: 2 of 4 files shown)


SYNchroACK:
@Phantasus the code looks awesome! Well done!

Some suggestions:

* Please remove the tests files (unless you see future benefits to keep them, even though you should make a new pull request for test files)
intelmq/tests/__init__.py
intelmq/tests/bots/__init__.py
intelmq/tests/bots/parsers/__init__.py
intelmq/tests/bots/parsers/dcu/__init__.py
intelmq/tests/bots/parsers/dcu/lib.py

* From Data Harmonization document the 'malware' value means "A URL is the most common resource with reference to malware binary distribution."

* From Data Harmonization document the 'botnet drone' value means "This is a compromized machine, which has been observed to make a connection to a command and control server."

aaronkaplan:
On Jun 2, 2015, at 4:50 PM, Tomás Lima <notifications@github.com> wrote:

> @Phantasus the code looks awesome! Well done!

:) I agree!

> 
> Some suggestions:
> 
> 	• Please remove the tests files (unless you see future benefits to keep them, even though you should make a new pull request for test files)
> intelmq/tests/init.py
> intelmq/tests/bots/init.py
> intelmq/tests/bots/parsers/init.py
> intelmq/tests/bots/parsers/dcu/init.py
> intelmq/tests/bots/parsers/dcu/lib.py

Well, I thought in general we would like to have unit tests.
I guess this is a proposal on how to do them (?)

> 
> 	• From Data Harmonization document the 'malware' value means "A URL is the most common resource with reference to malware binary distribution."
> 
> 	• From Data Harmonization document the 'botnet drone' value means "This is a compromized machine, which has been observed to make a connection to a command and control server."
> 

Yes, this is in general all not very clear in the DCU feed.
So it is a bit of guesswork for DCU.

SYNchroACK:
Yes, you can have the tests but it should be commited in a different pull req.

Regarding the keys of DCU, what I meant is 'malware' in Data Harmonization means an URL not an infected machine, which is the case.

aaronkaplan:
On Jun 2, 2015, at 5:57 PM, Tomás Lima <notifications@github.com> wrote:

> Yes, you can have the tests but it should be commited in a different pull req.
> 
> Regarding the keys of DCU, what I meant is 'malware' in Data Harmonization means an URL not an infected machine, which is the case.
> 

Ah!! okay.
So if that was not even clear for me, we need to probably be more specific in that document.

SYNchroACK:
yap. make a proposal and I will follow you @aaronkaplan ;)

aaronkaplan:
Hi, we took a look at the data harmonisation document and it **clearly** says that the **field** malware means: "A malware family name in lower case."
What you referenced in the data harmonisation document was the eCSIRT taxonomy type.

So , we leave it for now. We can always change it again :) No worries.

aaronkaplan:
we tested and reported the testing results in #184 . Closing this for now, please send another PR.
Thanks, much appreciated.

SYNchroACK:
Well...the '__THREAT_CODES' dict has a match between the malware name and the classification.type.

Im trying to say that sometimes the match between malware name and classification.type doest make sense, example:
```
   "B106-Agent\tmalware\n"
```
classification.type="malware" means "A URL is the most common resource with reference to malware binary distribution.". In this case, I think the 'malware' value for classification.type key is not suitable. It should be 'botnet drone'.

Make sense?

SYNchroACK:
Why did you close? The reference that you put here goes to collector, not the parser... im confused.

aaronkaplan:
because that is what Philip and me just discussed (sitting next to each other): that he will make another pull request.

SYNchroACK:
well..ok.

Issue #64, opened 2014-07-22
Repository size


SYNchroACK:
http://dound.com/2009/04/git-forever-remove-files-or-folders-from-history/

SYNchroACK:
@sebix do you know any good solution to solve this problem?

sebix:
Have there been any big files in the past?

To keep the number of commits small, please don't push to master directly. When looking through master I often see small commits and following commits correcting some minor bug, typo. Please use 1) branches and optionally 2) push them to your fork and review the PR another day, and 3) use `git rebase` and `git commit --amend` to stash commits. Have a look here for some hints on rebasing: http://stackoverflow.com/questions/6934752/combining-multiple-commits-before-pushing-in-git

Commits represent *small independent* changes, this is good. but also don't exaggerate.

SYNchroACK:
yes, thats why I ask. Do you know how to remove that old file that was committed by mistake?

Ok, thank you for the explanation. Always learning heheh

sebix:
You can use the [https://rtyley.github.io/bfg-repo-cleaner/](BFG Repo cleaner), remove the file `src/bots/experts/geoip/GeoLite2-City.mmdb` or do `bfg --strip-blobs-bigger-than 10M  my-repo.git`. You need to force the push (`-f`) as remote has more commits than local and would be rejected otherwise.

This will reduce the size of the repo from 19MB to 2MB.

aaronkaplan:
On Aug 10, 2015, at 9:20 AM, Sebastian <notifications@github.com> wrote:

> You can use the https://rtyley.github.io/bfg-repo-cleaner/, remove the file src/bots/experts/geoip/GeoLite2-City.mmdb or do bfg --strip-blobs-bigger-than 10M my-repo.git. You need to force the push (-f) as remote has more commits than local and would be rejected otherwise.
> 
> This will reduce the size of the repo from 19MB to 2MB.

@SynchroAck: okay? Do you want to do this or should I?

sebix:
The file is still there:

https://github.com/certtools/intelmq/blob/ef24878a400d857c3dbf237f31df419b5ff8ef46/src/bots/experts/geoip/GeoLite2-City.mmdb

I'm just trying to get my fork right. Updating master and v0.9 branches are quite easy, but I'm struggling with the feature branch. Here's what I have yet to get the history fuck up right:

First, check if you have the upstream repository configured:

    git remote -v
If there's no upstream repository from certtools, add it with SSH or HTTPS:

[...]

PR #213, opened 2015-06-04
Generic Parser for beta version.

Th4nat0s:
Since we can't do named group with "." I added a mapping parameter, here a configuration file : 
here test for abuse.ch...  regex group is ip, then mapped to source.ip

```
  "generic-parser": {
          "error_procedure": "retry",
                    "error_retry_delay": 30,
                                    "error_log_message": true,
                                                    "error_dump_message": true,
    "regex": "^\\s*(?P<ip>(?:(?:\\d){1,3}\\.){3}\\d{1,3})",
                "mapping": "ip:source.ip",
                "feed_name": "AbuseCH",
                "feed_url": "https://zeustracker.abuse.ch/blocklist.php?download=badips",
                "classification_type": "malware"
  }
```


Th4nat0s:
Example, replacing the phishtank parser by the generic one

```
  "phishtank-collector": {
    "error_procedure": "retry",
    "error_retry_delay": 30,
    "error_log_message": true,
    "error_dump_message": true,
    "url": "http://data.phishtank.com/data/online-valid.csv",
    "rate_limit": 3600
  },
  "phishtank-parser": {
    "error_procedure": "retry",
    "error_retry_delay": 30,
    "error_log_message": true,
    "error_dump_message": true,
    "regex": "^\\d+,(?P<url>[\\S:]+),(?P<durl>http:\\/\\/www.phishtank.com\\/phish_detail.php\\?phish_id=\\d+),(?P<time>\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\+\\d{2}:\\d{2}),yes,\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\+\\d{2}:\\d{2},yes,(?P<target>.*)$",
    "mapping": "url:source.url,durl:description,time:time.source,target:comment",
    "feed_name": "phishtank",
    "classification_type": "phishing",
    "feed_url": "http://data.phishtank.com/data/online-valid.csv"
  },




Results
{"comment": "Other", "time.source": "2015-06-06T06:02:49+00:00", "description": "http://www.phishtank.com/phish_detail.php?phish_id=3246644", "source.url": "http://documentsapp-reviewallfeeb.rhcloud.com/viewer/open", "time.observation": "2015-06-06T08:16:12+00:00", "classification.type": "phishing", "feed.name": "phishtank", "feed.url": "http://data.phishtank.com/data/online-valid.csv"}
{"comment": "Other", "time.source": "2015-06-06T06:01:58+00:00", "description": "http://www.phishtank.com/phish_detail.php?phish_id=3246636", "source.url": "http://hot-wire.ru/chase.update/info.php", "time.observation": "2015-06-06T08:16:12+00:00", "classification.type": "phishing", "feed.name": "phishtank", "feed.url": "http://data.phishtank.com/data/online-valid.csv"}

```

Th4nat0s:
Waiting the Official "Royce Rolls, MISP Plugin", here how to parse data from MISP feed with beta version of IntelMQ. This example get URL

You need the header patch : https://github.com/certtools/intelmq/pull/216 , and this generic parser https://github.com/certtools/intelmq/pull/213 and a good misp broker :) with the following feature : https://github.com/MISP/MISP/commit/71635b00a175e2aeb74d42973d42c912acc9e218 in production

and the following config : 
```
  "mispuri-collector": {
    "error_procedure": "retry",
    "error_retry_delay": 30,
    "error_log_message": true,
    "error_dump_message": true,
    "header": "Authorization: mymipsauthapikey",
    "url": "https://mywonderfullmispbroker/events/csv/download/false/false/false/false/url/false/false/false/1d",
    "rate_limit": 3600
  },
  "mispuri-parser": {
    "error_procedure": "retry",
    "error_retry_delay": 30,
    "error_log_message": true,
    "error_dump_message": true,
    "regex": "^[\\d\\S\\-]{36},(?P<id>\\d+),[\\S\\ ]+,url,\"(?P<uri>.*)\",\\d,(\\d+)$",
    "mapping": "uri:source.url,id:comment",
    "feed_name": "mispuri",
    "classification_type": "malware",
    "feed_url": "https://mywonderfullmispbroker/"
  },

```

Timing interval in url of misp could be tuned to "2h"

otmarlendl:
One note here:

I also wrote a generic parser for various data-feeds for CERT.at's legacy event-database. This is pretty similar, but I made a slightly different design decision back then:

Most of these feeds are de-facto CSV files, and most of them (as the phishtank one) contain a header line defining column names. I rely on these for my generic parser, whose configuration contains the following information:

* csv-flavour definition (separator, handling of empty fields, ... in other words the parameters of new() within http://search.cpan.org/~makamaka/Text-CSV-1.33/lib/Text/CSV.pm#new_%28\%attr%29 )
* optional: a header line if the input format doesn't contain one
* a dictionary mapping the column-names from the CSV into my internal field names
* code callbacks for manual fixups (e.g. timezone adjustments)

My reasons for this approach were:

* a change in the imput format (e.g. a new column added somewhere) won't break the parser
* rely on a well-tested csv library which handles all the quoted strings and escaped quotes
* make the config clearer as it's plain what field on the input gets mapped to what flield on the output
* do the "are all fields in the right format" checking afterwards in the pipeline and independent of the parser

Th4nat0s:
Yes @otmarlendl  It make sense for some feeds I deeply agree. For parsing a CSV with clean header your solution seems nice and robust.

My little parser is able since it's purely regex based to grab anything in a page. I think this option make sense in some cases.

At the end I think we need to offer 3 "universal" parsers, a csv dedicated, a regex based and an object based (json/xml).  It may expand the parsing capabilities to any feed and could give essential bricks for anyone for adding any feed.

Could you push your code, i'm interested to look at the way you setup fixes for time fixup per example (got the same problem).

In term of flexibility it could be nice also to be able (for my parser too) to "expand" a data.. dummy example...   you got an event id in misp, it will be nice to expand to the url of the event in the record 

Does @aaronkaplan or @SYNchroACK  agree with my views also or am i alone :) ?

SYNchroACK:
I completely agree with @otmarlendl .
@Th4nat0s , yes, we need 3 "universal" parsers! You're right.

Regarding the following phrase, can you explain a little more:
```
In term of flexibility it could be nice also to be able (for my parser too) to "expand" a data.. dummy example... you got an event id in misp, it will be nice to use this id to generate the url of the event in the record.
```

otmarlendl:
@Th4nat0s 

Yes, the tree base types of parsers sound like a good idea.

Regarding my code: that's in perl; I can share the relevant snippets, but not here on github. Please ping my by mail (lendl@cert.at) so I can send some examples.

In my setup I used a class hierarchy to subtype a generic parser. A new subclass only had to parametrize the generic functions from the parent by either setting class variables or overloading functions, e.g.:

*  set REs for grabbing the right tickets/attachments from our RTIR
*  column mappings

Invoke generic function with the right CSV paramenters

    sub parse_csv {
        my ($self, $content) = @_;

        return($self->SUPER::parse_csv($content, {
                binary => 1,
                allow_whitespace => 1,
                blank_is_undef => 1,
                empty_is_undef => 0,
                }));
    }

And provide manual fixups to be called for each event:

    sub fixup_eventdb {
        my ($self, $row) = @_;
        $row->{ts} = $row->{ts} . " UTC";
        $row->{malware} = 'freak';
    }

aaronkaplan:
On Jun 8, 2015, at 1:10 PM, Tomás Lima <notifications@github.com> wrote:

> I completely agree with @otmarlendl .

Yes, I also agree.

> @Th4nat0s , yes, we need 3 "universal" parsers! You're right.
> 

But please make sure that they stay understandable.
How about adding a "comment": "this is a comment explaining the regexp"  key-value to the json config file?

Th4nat0s:
For flexibility I have two cases in mind.

Sometimes you got an ID of the event in the feed, and the feed website offer direct link like http://website/event?id.  It will be nice to have a solution to construct this kind of feed uri and put it in the record.

Then time... (I hate time)... I think there are as much feed that time format used !... but maybe it is the role of "normalise" of the object time to try to find the time format.

Is it possible to add an Hint to every parameter, for example :   
"regex": "^[\\d\\S\\-]{36},(?P<id>\\d+),[\\S\\ ]+,url,\"(?P<uri>.*)\",\\d,(\\d+)$",
"regex.hint": "Put a named group regex",
"mapping": "uri:source.url,id:comment",
"mapping.hint": "Used to link regex group to event",
The work should be done in intelmq-manager

SYNchroACK:
Sorry. Due branches changes, this pull req was automatically closed. @Th4nat0s "don't pull (fetch and merge) from remote, use rebase instead."

sebix:
@Th4nat0s: Please see https://github.com/certtools/intelmq/issues/64#issuecomment-130311349 and the following comment. Hope I can help you.

For Merging/Rebasing have a look here: https://www.atlassian.com/git/tutorials/merging-vs-rebasing

PR #224, opened 2015-06-09
Add docstrings to lib

Phantasus:
This PR does the following:

1. It adds docstrings to files in intelmq/lib/
2. It re-formats those files mostly according the output of the pep8 utility


aaronkaplan:
looks good from my side

SYNchroACK:
yes, cool. The only problem is the changes applied to master branch... I will merge but i think the idea is to change the beta-branch, otherwise its hard work to always merging... :S

PR #229, opened 2015-06-12
Proposal for testing a bot with a UnitTest

Phantasus:
Hi,

I wrote this proposal for testing a bot in a unittest. Maybe you find it useful, I think
with it we can check if a bot still behaves as we want it after a change. This is
at the moment propably only useful for experts or parsers as collectors have some
side effects, which need special care.

What does this rather long list of commits (most of them are actually small changes), 
which tried as much as possible to not change existing behaviour. 

## What did I basically do?

1. Add a `config=` argument to `Bot()`, which contains configuration which is then passed through the bots internal parsing. IMHO this is cleaner as we have only one code pathway. Not a "special test" pathway.

2. It refactors/reformats Bot() internals to be more readable and less WTF.

3. It renames `log()` to `create_file_logger` (with an alias `log()`) and adds `create_stream_logger` which does the same thing as log() but only on streams not on files. I also added a test for that behaviour.

4. It adds a `TestPipeline`, so a `Pipeline()` object which can be passed to a bot in a TestCase, which doesn't open a new redis/whatever connection in the background. 

5. It adds a crude TestCase to test the whole contraption with a naive bot parser, I called it TestBot in 
`intelmq/tests/lib/bot` and it can be run with `python -m intelmq.tests.lib.bot`. It's not beautiful yet,
but for a proposal "good enough".

6. It adds a .run_once method to Bot(), which isn't set by a config file. Maybe someone finds that a nice "feature" for a bot to run only once, besides testing. If yes then we could add it. 

7. It fixes a naming error in Pipeline(), probably a typo. Which caused that when setting a destination_queue with destination_queues() it overwrote itself with that value and as python is dynamic this didn't cause an error. From what I have seen of the beta-branch this was fixed there.

8. Adds `create_bot_test_configuration' a convenience function for setting up test cofiguration
for a bot.

## Keys in config= and what they mean

key | description
---- | -------------
"runtime" | a dict() containing the keys in `intelmq/conf/runtime.conf'
"system" | a dict() containing the keys in `intelmq/conf/system.con'
"pipeline" | a dict() containing the pipeline config (like in `intelmq/conf/pipeline.conf')
"logger" | a logger, which can log to a file or a stream
"source_pipeline" | a Pipeline() object for the bots source_pipeline
"destination_pipeline" | a Pipeline() object for the bots destination_pipeline

And if you don't want to do all the hassles of setting that dict() yourself up 
you can use the new convenience function `create_bot_test_configuration', which
expects the bot-id of the test bot (can be test-bot or anything you want), 
a log stream (e.g a file logger or a stream logger), a dict() for the inital state of
the bot pipelines and bot parameters dict() as it would normally be set in a startup.conf file. 

## Added UnitTests (and how to run them)

Summarized I added following UnitTests:

A test for stream_loggers and file_loggers: `python -m intelmq.tests.lib.utils -v`

A test for a TestPipeline, so that it behaves almost as a redis powered Pipeline() would do: `python -m intelmq.tests.lib.test -v`

And finally a test for all those components combined to test a simple "TestBot": `python -m intelmq.tests.lib.bot -v`

This tests also contains the logic of how future tests for bots could be written. I would only
suggest to re-factor the code contained in there in a special "BotTestCase" class which could
then be used for testing a real bot.


intelmq/tests/lib/utils.py
```diff
@@ -0,0 +1,64 @@
...
+        expected_line = log_format.format(name="test", type="ERROR", text="too bad")
+        match = re.match(expected_line, log_lines[2])
+        self.assertTrue(match, "Logline 2 was: '%s' expected '%s'" % (log_lines[2], expected_line))
+
+    def test_stream_logger_creation(self):
+        """Tests if a logger for a stream can be generated with log().
+           Why? This is useful for writing tests for loglines. 
+           For example, when a bot needs to be tested if it generates a specific logline"""
+
+        io = StringIO.StringIO()
```
sebix: `io` is a misleading name, as there is also the [`io` module](https://docs.python.org/2.7/library/io.html), introduced in 2.6

intelmq/tests/lib/utils.py
```diff
@@ -0,0 +1,64 @@
+from intelmq.lib import utils
+import unittest
+import tempfile
+import re
+import os.path
+import StringIO
```
sebix: Using `import io.StringIO` or `from io import StringIO` is Python3-proof. This is not a requirement currently, but will make it easier later.

intelmq/tests/lib/bot.py
```diff
@@ -0,0 +1,108 @@
+from intelmq.lib.bot import Bot
+from intelmq.lib.test import create_bot_test_configuration
+
+from intelmq.bots import utils
+from intelmq.lib.message import Event
+
+import unittest
+import StringIO
+
+import pdb
```
sebix: `pdb` is not used

intelmq/lib/bot.py
```diff
@@ -52,45 +58,44 @@ def start(self):
...
                 local_retry_delay = self.parameters.retry_delay
                 self.logger.info("Last Correct Message(event): %r" % self.last_message)
                 self.logger.info("Current Message(event): %r" % self.current_message)
                 self.logger.exception("Check the following exception:")
                 self.logger.error('Pipeline connection failed (%r)' % ex)
                 self.logger.info('Pipeline will reconnect in %s seconds' % local_retry_delay)
-                self.source_pipeline = None
-                self.destination_pipeline = None
-
             except KeyboardInterrupt as e:
```
sebix: Why have these lines been removed? `e` is not used.

intelmq/lib/bot.py
```diff
@@ -108,11 +113,40 @@ def check_bot_id(self, str):
             print "Invalid bot id."
             self.stop()
 
-    def load_system_configurations(self):
+    def load_config(self, thing):
```
sebix: `thing` is not a meaningful variable name.

intelmq/tests/lib/utils.py
```diff
@@ -0,0 +1,64 @@
...
+        logger = utils.log(dirname, "test")
+
+        logger.info("Hello")
+        logger.warning("Oh lovely")
+        logger.error("too bad")
+        
+        filename = os.path.join(dirname, "test.log")
+
+        self.assertTrue(os.path.isfile(filename))
+        log_lines = open(filename).readlines()
```
sebix: `open()` used without `with`-statement.

intelmq/lib/pipeline.py
```diff
@@ -5,28 +5,34 @@
...
             destination_queues = destination_queues.split()
         self.destination_queues = destination_queues
 
+    def connect(self):
+        """Connect or re-connects to a pipeline provider(e.g redis)"""
+        self.redis = redis.Redis(host=self.host,
+                                 port=self.port,
+                                 db=self.db,
+                                 socket_timeout=self.socket_timeout)
+        pass
```
sebix: `pass` is superfluous here.

sebix:
Would like to see this merged soon.

sebix:
> I wrote this proposal for testing a bot in a unittest.

Could you please also add a test for a bot?
Also, docs about testing (as described in your text above) in `docs/Testing.md` would be nice.

sebix:
Can be closed. Everything is included in #282.

SYNchroACK:
Thank you @Phantasus 
Thank you @sebix

aaronkaplan:
Tomas why was this closed?

The unit tests are not yet finalized. 
Closing this issue means I should open another one?

Please advise what you mean. Thanks :)


---
Mobile

SYNchroACK:
@aaronkaplan , check last comments from @sebix in this issue.

sebix:
On 08/15/2015 08:07 PM, AaronK wrote:
> The unit tests are not yet finalized.
> Closing this issue means I should open another one?
The Unittests are neither complete nor final, yes. But the ideas and
actual changes proposed in this PR are all incorporated in my fork and
PR, see #282.

If you want an issue to track the progress of unit tests, please open
one, this is a PR.

aaronkaplan:
Got it. Thanks for the clarification. 

---
Mobile

PR #237, opened 2015-06-19
Sets default loglevel to INFO in beta

Phantasus:
This PR sets the default loglevel to debug in beta,
it wasn't tested as it looked trivial.


SYNchroACK:
v1.0-beta is not in use. Use v1.0-final branch.

PR #245, opened 2015-07-01
ASN lookup discarding events

robcza:
An event without source.ip or destination.ip can't go through asn_lookup expert and is just discarded.


SYNchroACK:
Nice one! ;)

aaronkaplan:
Yes, thanks Robert!

PR #269, opened 2015-07-10
Fix for missing user_agent parameter

robcza:
simple check, whether user_agent parameter exists

### DIFF
--- a/intelmq/bots/collectors/http/collector_http.py
+++ b/intelmq/bots/collectors/http/collector_http.py
@@ -12,7 +12,7 @@ class URLCollectorBot(Bot):
                                 chunk_size = 16384,
                                 http_proxy=self.parameters.http_proxy,
                                 https_proxy=self.parameters.https_proxy,
-                                user_agent=self.parameters.user_agent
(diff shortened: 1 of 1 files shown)


SYNchroACK:
Sorry @robcza but We have been discussing this issue and the temporary solution is to put the parameter 'user_agent: null' in defaults.conf. The solution that you propose it not correct because for each parameter in the same situation we will have to add in all bots that if, which cannot be. So, the solution is to define the default value in some generic place like defaults.conf. The best solution will be created in some days, because the configurations need some improvements. I hope you dont mind. 

Check the solution:
3fe60694aadb7983959b2383123bcd701973263e
35f5aad4992ffd041d84c0d41226129ec0882aa5
d521093bef70afe9958179ed4fce36362e91ef96

robcza:
I don't, seems legit :)

SYNchroACK:
;)

SYNchroACK:
If you have time, test the current solution. ;) I dint have time to do it... :(

robcza:
I've just rerun the botnet with the changes of yours and it works

PR #270, opened 2015-07-10
Parser fixes

robcza:
minor fixes to Turris and CI Army parsers (didn't work)


SYNchroACK:
Awesome! Thanks for the bug fix

PR #274, opened 2015-07-29
ZeuS, Palevo and Feodo trackers

robcza:
collectors and parsers for abuse.ch projects


SYNchroACK:
@robcza , what do you think about collect the information from RSS sources instead of a simple list:

https://zeustracker.abuse.ch/feeds.php

Give me your opinion and lets decide together ;)

SYNchroACK:
Sorry. Due branches changes, this pull req was automatically closed. @robcza can you give me feedback about the collection procedure that I mentioned in the previous comment?

robcza:
The reason is to have additional information (ASN, CC, level, etc.) included or am I missing something?

SYNchroACK:
yes... it has more context information that is useful for analysis

robcza:
makes sense :) i'll rewrite those bots and create a new pull request

PR #275, opened 2015-08-03
Various minor changes/improvements in Intelmqctl

sebix:
The commit messages should be rather self-explaining:

* MAINT: PEP8 for intelmqctl
  using autopep8
* ENH: Use new string formating instead of old % method
* ENH: use open() only in with-statement
* ENH: catch wrong user input and generalize log_bot
  * If bot ID is not given or invalid, no exception will be thrown now
  * The log_bot_* functions now have the positional argument first, then the optional parameter (e.g. bot id)
  * Also, multiple (including named) arguments can now be used

### DIFF
--- a/intelmq/bin/intelmqctl
+++ b/intelmq/bin/intelmqctl
@@ -21,7 +21,7 @@ class Parameters(object):
     pass
 
 PIDDIR = "/opt/intelmq/var/run/"
-PIDFILE = "/opt/intelmq/var/run/%s.pid"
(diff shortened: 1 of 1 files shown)


SYNchroACK:
Awesome! ;) Thank you @sebix

sebix:
The warm-up :)

SYNchroACK:
WoW! :D I hope so... you have my support! ;)

Issue #257, opened 2015-07-07
intelmqctl improvements


SYNchroACK:
- load parameters from all configurations
- bot.py must has static methods to load configurations and then intelmqctl will be able to use them

use the following code: https://github.com/certtools/intelmq/blob/v1.0-final/tests/intelmqctl/intelmqctl.py

sebix:
The methods used by `Bot` class use `utils.load_configuration` to load it and then filter out the interesting section. Using the utility function is easier here and has less overhead.

> load parameters from all configurations

Currently, `RUNTIME_CONF_FILE` is not used, thus there's no functionality needing it.

sebix:
I think this should be fixed in my fork

SYNchroACK:
@sebix is everything mentioned in this issue fixed by your pull req?

sebix:
> * load parameters from all configurations
Default, System, pipeline, runtime and startup are read
> bot.py must has static methods to load configurations and then intelmqctl will be able to use them
I commented above that I don't see the sense here. The functions in utils are used now.

SYNchroACK:
ok cool.

Issue #218, opened 2015-06-07
Issue with the Bot file output


Th4nat0s:
I got a really strange issue with the last beta 1. I had linked the parser to the deduplicator then to file output

The bot don't start...
```
# ./intelmqctl --bot status --id file-output                                                                                                  
file-output is stopped.
# ./intelmqctl --bot start --id file-output                                                                                                   
Starting file-output...
file-output is already running.
# ./intelmqctl --bot status --id file-output                                                                                                  
file-output is stopped.
```

It didn't generate any logs.. BUT ...Manually... seems Ok.. no error, and no output
```
# python -m intelmq.bots.outputs.file.output file-output
```
The logs then start... but it do "nothing"...
```
2015-06-07 02:48:42,995 - file-output - INFO - Bot is starting
2015-06-07 02:48:42,995 - file-output - DEBUG - Defaults configuration from '/opt/intelmq/etc/defaults.conf' file
2015-06-07 02:48:42,995 - file-output - DEBUG - Defaults configuration: parameter 'destination_pipeline_host' loaded with value '127.0.0.1'
2015-06-07 02:48:42,995 - file-output - DEBUG - Defaults configuration: parameter 'error_max_retries' loaded with value '3'
2015-06-07 02:48:42,995 - file-output - DEBUG - Defaults configuration: parameter 'destination_pipeline_port' loaded with value '6379'
2015-06-07 02:48:42,996 - file-output - DEBUG - Defaults configuration: parameter 'destination_pipeline_db' loaded with value '2'
2015-06-07 02:48:42,996 - file-output - DEBUG - Defaults configuration: parameter 'rate_limit' loaded with value '0'
2015-06-07 02:48:42,996 - file-output - DEBUG - Defaults configuration: parameter 'error_log_exception' loaded with value 'True'
2015-06-07 02:48:42,996 - file-output - DEBUG - Defaults configuration: parameter 'broker' loaded with value 'redis'
2015-06-07 02:48:42,996 - file-output - DEBUG - Defaults configuration: parameter 'error_procedure' loaded with value 'retry'
2015-06-07 02:48:42,996 - file-output - DEBUG - Defaults configuration: parameter 'error_retry_delay' loaded with value '15'
2015-06-07 02:48:42,996 - file-output - DEBUG - Defaults configuration: parameter 'source_pipeline_host' loaded with value '127.0.0.1'
[...]

PR #277, opened 2015-08-06
Intelmqctl

### LINKED ISSUE #244: Clear a queue
Hello,

maybe I'm have overlooked something, but I've not found a convenient way, to clear a queue.

Possible usecases:
- unintentionally fetch garbage, which then keeps stuck somewhere in the pipe
- debugging, when the original collector format or the parser changes and old data is still in the queue
- aborting processing, when it would clearly take too long

Therefore I would suggest to have a clear command in intelmqctl and a "clear queue" button in the "monitor"-tab of the Manager.

sebix:
* ENH: intelmqctl uses print_function
* BUG: intelmqctl silently exits in undefined function
* ENH: intelmqctl command line argument shorthands
* ENH: intelmqctl log parsing/reading simplified

  The logs are read and parsed backwards, which is more efficient than reading and parsing the entire file from the beginning
The code to read the file is easier to read and understand now, however some code in utils reads the file backwards in chunks.
This change is currently not compatible with intelmq-manager as it wants to read the line number which is not read anymore as its not possible to get it by reading the file backwards. Yet a line number does not make sense in log files.

See certtools/intelmq-manager#53
* ENH: intelmqctl logs to file and add stream logger

  intelmqctl has a log file as requested in certtools/intelmq#218 (partial fix)
To achieve this, the logger got extended by a stream handler aside the existing file handler. This helps debugging bots as running them solely on the console, directly gives the log output in stderr (default of StreamHandler)

* ENH: `intelmqctl --type` alias `-t`
* ENH: read all configs using utils, fixes certtools/intelmq#257
* ENH: `intelmqctl --clear` queue added, fixes #244 
* BUG: intelmqctl Give all lines for -1

### DIFF
--- a/intelmq/bin/intelmqctl
+++ b/intelmq/bin/intelmqctl
@@ -161,24 +161,25 @@ class IntelMQContoller():
 
         parser.add_argument('-v', '--version',
                             action='version', version=VERSION)
-        parser.add_argument("--id", dest='bot_id', default=None, help='bot ID')
+        parser.add_argument('--id', '-i',
+                            dest='bot_id', default=None, help='bot ID')
         parser.add_argument('--type', choices=RETURN_TYPES,
                             default=RETURN_TYPES[0],
                             help='choose if it should return regular text or '
                                  'other forms of output')
 
-        group_list.add_argument('--log',
+        group_list.add_argument('--log', '-l',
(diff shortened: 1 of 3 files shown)


robcza:
this will actually help a lot, thank you @sebix 
ping @rattuscz

Issue #255, opened 2015-07-07
bot __init__ improvements


SYNchroACK:
Any exception in init() method should be fatal. It means that init
need to catch every exception and log a messsage like "im stopping" and
raise something....

Other situation: 
check_bot_id should raise an exception like mentioned before

sebix:
Fixed by #282 when merged.

SYNchroACK:
merged

Issue #248, opened 2015-07-07
Generic stop method


SYNchroACK:
* stop method should be generic to fit the kill method from @helderfernandes1279

sebix:
Currently the `stop()` method kills the bot with `exit(-1)`. I would like to split the functionality into `stop` and `terminate`. While `stop` does not exit the bot, `terminate` will exit, but also disconnects all pipelines, so it can be used in exception handling. Stop can then be extended, like it is done in #197

SYNchroACK:
Ok, make sense. Send a pull request to discuss minimal details with you. ;)
Thank you @sebix

sebix:
9a82ce8 splits the functionality in stop() and terminate(). Stop currently closes the pipelines and can be extended by derived classes (using super) to close other connections. If `exit_on_error` is true (default), terminate will be called which exits, this is disabled for tests.

SYNchroACK:
merged

sebix:
Keep this issue in mind for #345

PR #236, opened 2015-06-18
Better bot test interface proposal

Phantasus:
Hi there :-)

I have the following proposal for an improved interface for
testing one individual bot (e.g a parser), which I had in my mind
before some business related "inconveniences" hit me right after 
the last unit-test PR (#229). I hope I don't have to deal with 
people who cause such business related nonsense ever again.

Here is how this "improved" interface looks like at the moment:
```
class TestBotTestCase(BotTestCase):
  def setUp(self):
    self.bot_type = TestBot
    self.reset_bot()

  def test_bot_run():
    self.input_queue = ["1.1.2.3\t2.2.4.5\t2015-06-11 16:49:42.262784"]
    expected_event = {"feed": "test", 
                                 "feed_url": "hxx://127.0.0.1",
                                 "type": "brute-force", 
                                  "reported_source_ip": "1.1.2.3",
                                  "reported_destination_ip": "2.2.4.5",
                                  "source_time": "2015-06-11 16:49:42.262784" }
    self.run_bot()
    self.assertLogLengthEqual(20)
    self.assertLoglineEqual(0, "Bot is starting", "INFO")
    self.assertNotRegexpMatchesLog("ERROR")
    self.assertRegexpMatchesLog("WARNING - It is soo awesome, we put also a warning there")
    self.assertEventAlmostEqual(0, expected_event)
```

That code can be found in full in `intelmq/tests/lib/bot.py` if you want that I choose
a different unittest framework -- no problem. Or if you think something needs to be improved,
let me hear it :-)

Here is the list of new assertions which this PR includes for the "better bot testing interface":
1. `assertLoglineEqual`, asserts that the fields of a specific logline which was generated by a bot is equal.
2. `assertRegexpMatchesLog`, asserts that a regular expression matches against the whole log generated by a tested bot.
3. `assertNotRegexpMatchesLog`, asserts that a regular expression matches NOT against the whole log generated by a tested bot. 
4. `assertLogLengthEqual`, asserts that the log has the given lines. 
5. `assertEventAlmostEqual`, asserts that the specified event at a specific position in the output queue of a tested bot is almost equal to a given dict. 

I hope that explanation is helpful to you :-)

PR #282, opened 2015-08-12
Add unittests for lib components

### LINKED ISSUE #159: Extra - Automatic unit testing
intelmq so far completely lacks unit testing features. It would be very important to have unit testing in place - especially since the project is growing. 

There are a couple of mechanisms in Python to do unit testing - for example: https://docs.python.org/2/library/unittest.html 

For unit-testing individual  bots, I propose having a special "testing" mode where a well defined pipeline data comes into the bot and goes out of the bot. Then it is compared against expected output.

Comments welcome :)

sebix:
Adds tests for all core functionalities.

Obsoletes #229 and #236. Closes #159, #248 and #255.


SYNchroACK:
@sebix for me this pull req is a priority. If you have time, please update the pull req and I think we should split your core changes that are related to some bug fixes and additional improvements from the unittest feature.

Is it ok?

sebix:
I rebased my branch to fix merge conflicts.
My intention is to fix the code style of code I had to modify. Otherwise, I would have to manage two branches, switching between those and apply the changes to only one of them. This would increase the overhead enormously, including merge conflicts.

In my opinion the PR is finished now, basic unittests are implemented including tests for bots. You can then start from there to write more tests.

intelmq/lib/bot.py
```diff
@@ -1,66 +1,64 @@
-import re
-import sys
+# -*- coding: utf-8 -*-
+from __future__ import print_function, unicode_literals
```
SYNchroACK: @sebix i didnt understand this line...

intelmq/lib/bot.py
```diff
@@ -224,53 +235,52 @@ def dump_message(self, ex):
...
-        self.logger.debug("Defaults configuration"
-                          " from '%s' file" % DEFAULTS_CONF_FILE)
+        config = config or utils.load_configuration(DEFAULTS_CONF_FILE)
 
-        for option, value in config.iteritems():
+        for option, value in config.items():
             setattr(self.parameters, option, value)
-            self.logger.debug("Defaults configuration: parameter '%s' "
-                              "loaded with value '%s'" % (option, value))
+            # TODO: Log that really?
```
SYNchroACK: Didnt understand... i think we should log of course... in debug level...

intelmq/lib/bot.py
```diff
@@ -224,53 +235,52 @@ def dump_message(self, ex):
...
-        config = utils.load_configuration(RUNTIME_CONF_FILE)
+        config = config or utils.load_configuration(RUNTIME_CONF_FILE)
 
         self.logger.debug("Runtime configuration: loading '%s' section from"
                           " '%s' file" % (self.bot_id, RUNTIME_CONF_FILE))
-
-        if self.bot_id in config.keys():
-            for option, value in config[self.bot_id].iteritems():
+        self.logger.debug("{}".format(list(config.keys())))
+        if '__default__' in list(config.keys()):
```
SYNchroACK: from line 265 to 270 we can remove that....that feature is deprecated... not we have defaults.conf file.

intelmq/lib/exceptions.py
```diff
@@ -15,8 +15,16 @@ def __init__(self, message):
 
 class InvalidArgument(IntelMQException):
 
-    def __init__(self, argument):
-        message = "argument %s is invalid" % repr(argument)
+    def __init__(self, argument, got=None, expected=None, docs=None):
```
SYNchroACK: just curiosity, why you add this? :)

intelmq/lib/harmonization.py
```diff
@@ -1,19 +1,22 @@
-import sys
+# -*- coding: utf-8 -*-
+"""
+
+TODO: staticmethods and inheritance?
```
SYNchroACK: heheh :) @sebix what would be your suggestion to improve? :P

intelmq/lib/bot.py
sebix: The `__future__` package brings some changes that are mandatory in future python versions, see [docs](https://docs.python.org/2.7/library/__future__.html). The import of `print_function` deletes the print statement and imports the print function, the `unicode_literals` makes all strings like `'foobar'` to unicode strings by default `u'foobar'`. Testing for unicode in quite essential for the logger.

Both features are standard in python3, so this file is compatible to python3 in regard of unicode (there are also other changes, but these are important here).

intelmq/lib/bot.py
sebix: Good to know, will change the code accordingly.

intelmq/lib/message.py
```diff
@@ -1,66 +1,101 @@
...
+import json
 import hashlib
-import base64
 from intelmq import HARMONIZATION_CONF_FILE
 from intelmq.lib import utils
 import intelmq.lib.harmonization
 import intelmq.lib.exceptions as exceptions
 
 
+harm_config = utils.load_configuration(HARMONIZATION_CONF_FILE)
```
SYNchroACK: Why did you put this line here? We evaluate this line when we wrote it and we found that line there will be load multiple times...less perfomance...   dont remember well but I think that was one of the reasons

intelmq/lib/exceptions.py
sebix: It was very handy to have more informations about the actual wrong value (not just the variable name) when working on the pipeline tests.

intelmq/lib/message.py
sebix: Because it is used in Message.__init__ and MessageFactory.unserialize. Instead of loading it multiple lines, it's only loaded once.

SYNchroACK:
well, the code looks really good! We will discuss a little bit more some details but is almost here.

intelmq/lib/bot.py
sebix: As discussed the log messages are cached until the logger is setup. If something fails in this time, messages will be printed to stdout.

sebix:
@aaronkaplan proposed to rename `reset_bot` and probably `run_bot` to something more intuitive. As `run` is not possible (used by `unittest.TestCase`), I chose `prepare_bot` and `run_bot`.

Additionally the constants in Bot tests (id, class reference, ...) are now set in a function called `set_bot` so we got rid of the `super()` call. `set_bot` is now called in `setUp` by `BotTestCase` and not in `reset_bot` anymore.

intelmq/lib/harmonization.py
sebix: Proposal is here: https://github.com/sebix/intelmq/tree/harmonization

SYNchroACK: Do you will add in this pull req?

SYNchroACK:
@sebix , can you fix the conflicts?

intelmq/lib/bot.py
```diff
@@ -24,26 +28,34 @@
...
+            self.check_bot_id(bot_id)
+            self.bot_id = bot_id
+
+            self.load_defaults_configuration()
+            self.load_system_configuration()
+            self.logger = utils.log(self.parameters.logging_path, self.bot_id,
+                                    self.parameters.logging_level)
+        except:
+            self.print_log_buffer()
+            raise
```
SYNchroACK: why not call 'terminate' method?

intelmq/lib/bot.py
```diff
@@ -24,26 +28,34 @@
...
 
-        self.logger = utils.log(
-                                self.parameters.logging_path,
-                                self.bot_id,
-                                self.parameters.logging_level
-                               )
         self.logger.info('Bot is starting')
-
-        self.load_defaults_configuration()
         self.load_runtime_configuration()
```
SYNchroACK: why this lines are not inside of try:except?

sebix: Because the logger is already loaded and the try-except block is only to catch errors before and during log creation.

intelmq/lib/harmonization.py
sebix: Done.

intelmq/lib/bot.py
sebix: Called `self.stop` instead.

sebix:
Also rebased on upstream master, no more conflicts.

intelmq/lib/bot.py
SYNchroACK: What happens if for some reason the bot cannot load the runtime configuration?

sebix: The same as before... I just changed this, the exception is first logged and then passed through (to the caller).

SYNchroACK:
@sebix Thank you very much. Awesome work!

Issue #292, opened 2015-08-20
ZeuS, Palevo and Feodo trackers RSS


robcza:
This is a followup to this pull request https://github.com/certtools/intelmq/pull/274
@SYNchroACK I have investigated the RSS feeds (direct links below) and I can see a problem there as all of these feeds include only 20 recently identified hosts and IntelMQ will miss the whole picture.

I think the best way is to prepare both parsers - for the complete lists and for the rss feeds as well. What do you say?

https://zeustracker.abuse.ch/rss.php
https://palevotracker.abuse.ch/?rssfeed
https://feodotracker.abuse.ch/feodotracker.rss

aaronkaplan:
I'd say let's go for the complete feed and make the RSS collector as soon a it's needed (for some other feed maybe?).  

My 2 cents. 

---
Mobile

robcza:
I have the complete feed ready. I will just retest and could go for pull request.

aaronkaplan:
Thanks Robcza!

SYNchroACK:
@robcza the following link has all the RSS feed for Zeus:
https://zeustracker.abuse.ch/feeds.php

If the problem is the RSS feed jjust show the last 20 events, well, we just need to configure the collector to get the report every 30 min..... make sense @robcza ?

aaronkaplan:
any news on this? I'd be willing to test the RSS bots.

SYNchroACK:
@aaronkaplan @robcza any news or can we close this one since we are handling the Abuse.ch data without RSS feeds.

robcza:
We can close, I din't find any use case for the RSS bot. Maybe @aaronkaplan which originally requested that one.

wagner-certat:
Closing this. Once we find out that one is needed, please reopen.

Issue #185, opened 2015-05-28
DataHarmonization Document improvements


Phantasus:
Hi,

the DataHarmonization document needs some improvements:

1. When looking at it in a browser the table is too big to be actually displayed in a browser window.
2. The document itself isn't so clearly understandable. e.g When to prefix a key with reported? When not? How is someone supposed to add a new field? Like http method? Should that then be reported_http_method?

aaronkaplan:
Thx nice catch. What you saw was an intermediate version/commit on github. Will update it. 

---
Mobile

SYNchroACK:
True. Regarding reported keys, IMHO, its something to remove because in the new version (https://github.com/certtools/intelmq-beta) we have raw field where we will specify the raw event.

I think its easy to agree with this one.

Feedback?

Phantasus:
@SYNchroACK, still not read. Trying to focus on that one :(

sebix:
I need your help @aaronkaplan and @SYNchroACK: When working with the postgres (see [my branch](https://github.com/sebix/intelmq/tree/postgres)) I saw that the Data-Harmonization document uses underscore to subclass values. But the code uses dots. Replacing all underscores with docs does not make sense in all senses, e.g. with `source.reverse_domain_name`.

Additionally, some fields have other names in code and documentation, as with `source.time` and `time.source`. This is kind of blocking any work that involves harmonization.

aaronkaplan:
---
Mobile


> On 20.08.2015, at 10:49, Sebastian <notifications@github.com> wrote:
> 
> I need your help @aaronkaplan and @SYNchroACK: When working with the postgres (see my branch) I saw that the Data-Harmonization document uses underscore to subclass values. But the code uses dots. Replacing all underscores with docs does not make sense in all senses, e.g. with source.reverse_domain_name.
> 

Got it. Well the underscores come from abusehelper compatibility.  I don't think we still need that anymore since we'll need a mapping anyway by now. So I would also be fine with camelCase or some other renaming. BUT! Let's be very careful about this! It means we will have to refactor everything and every bot 

@synchroack: what do you say?
> Additionally, some fields have other names in code and documentation, as with source.time and time.source. This is kind of blocking any work that involves harmonization.
> 

That seems like a bug then.

sebix:
On 08/20/2015 12:44 PM, AaronK wrote:
> > On 20.08.2015, at 10:49, Sebastian <notifications@github.com> wrote:
> >
[...]

PR #293, opened 2015-08-20
abuse.ch collectors and parsers

robcza:
Full list collectors and parsers for Zeus, Palevo and Feodo trackers. Does not include RSS feeds.


robcza:
RSS feeds could be included separately later as discussed here https://github.com/certtools/intelmq/issues/292

intelmq/bots/parsers/abusech/parser_domain.py
```diff
@@ -0,0 +1,40 @@
+import re
```
sebix: Unused import re

sebix:
You are missing the `bots/parsers/abusech/__init__.py`. Otherwise the submodule cannot be imported -> Parser fails to start.

sebix:
The documentation on Harmonization does not mention a field called `source.fqdn`, only `source_domain_name` (or `source.domain_name`), see also #185 

How to handle this @aaronkaplan and @SYNchroACK?

robcza:
thank you @sebix forgot to add ```bots/parsers/abusech/__init__.py```

intelmq/bots/parsers/abusech/parser_domain.py
robcza: import removed, thank you @sebix

aaronkaplan:
On Aug 21, 2015, at 10:34 AM, Sebastian <notifications@github.com> wrote:

> The documentation on Harmonization does not mention a field called source.fqdn, only source_domain_name (or source.domain_name), see also #185
> 

@sebix and me looked at it now. We'll go for {source,destination}.fqdn.
Why?
Because it is the most commonly used.

However, it is not documented in the Harmonisation document. I'd say we should also document this there :) Thank you!

Best,
a.

SYNchroACK:
#294 - issue with the enhancements mentioned by @sebix

I will create a new document with the new harmonization.

Thank you @robcza for the code.

PR #296, opened 2015-08-24
Code Style, PEP8 and repo cleanup

sebix:
-  Applies PEP8 to all source files
- cleanup of imports (order etc)
- import `unicode_literals` everywhere
- Removes `tests/` and `docs/IHAP-ideas-2015.mm`

Special thanks to autopep8, autoflake, isort and sed :)

Tests are passing. Local setup with some collectors, parsers and experts works.


SYNchroACK:
Awesome @sebix

PR #309, opened 2015-09-01
New bots, Encoding fixes in core, other bot fixes

sebix:
new bots:
 - cert.at contact database expert
 - fraunhofer DGA collector+parser
 - Spamhaus cert collector+parser

more docs and more tests


SYNchroACK:
nice!

PR #312, opened 2015-09-03
some bots fixes, Changelog updated

sebix:
Addresses some of the bug reports recently brought up.

And all tested components are Python 3 compatible, except for the parsers using unicodeccsv. Try running `nosetests3` :)
The changes needed to make it compatible brought some encoding issues up, and it actually solves some issues with the built-in csv parser. So I guess it will run better on 3 than on 2 ...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ dist
 *.old
 .vagrant/
 *~
+.coverage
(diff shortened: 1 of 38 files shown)


SYNchroACK:
thank you @sebix

PR #319, opened 2015-09-11
Modify bot corrections

sebix:
As merged too early, here are the fixes.

Config is not in `/etc/modify`, but in `var/lib/bots/modify` by default now, but configurable.


SYNchroACK:
ty @sebix

PR #330, opened 2015-09-18
Fixes: Py 3 installation, bots, CRLF, harmonization simplified

sebix:
Docs: recommend Py 3 by default
encoding fix in file output
fixes blocklist.de parser
harmonization: conf shortened, removed ascii
applied fromdos on some files for CRLF -> LF
setup.py has minimal requirements


SYNchroACK:
thanks @sebix 

_o_

PR #340, opened 2015-09-21
Documentation Improvements

SYNchroACK:
Installation support for Ubuntu 14.04, Debian 8 and CentOS 7
Installation support for python 3.4 and python 2.7


SYNchroACK:
@sebix whats your feedback?

sebix:
+1 for using a branch in your fork
side note: The prefix for the doc improvements should be DOC not ENH

Why do you require pycurl, pyzmq, pyasn1, pyOpenssl from pypi in REQUIREMENTS2, and additionally require python-zmq python-pycurl python-openssl python-pyasn1 from apt? The users end up with 2 different versions of these libraries in `/usr/lib/python2.7/dist-packages/` and `/usr/local/lib/python2.7/dist-packages/`.
Some of the package versions in REQUIREMENTS are current, others are not. I don't understand the policy -.-

You use the commands `python2.7` and `python3.4`. If we stick to this convention, we also have to change some shebangs and the start command in `intelmqctl`.

SYNchroACK:
> +1 for using a branch in your fork
> side note: The prefix for the doc improvements should be DOC not ENH

Well pointed.

> Why do you require pycurl, pyzmq, pyasn1, pyOpenssl from pypi in REQUIREMENTS2, and additionally require python-zmq python-pycurl python-openssl python-pyasn1 from apt? The users end up with 2 different versions of these libraries in > `/usr/lib/python2.7/dist-packages/` and `/usr/local/lib/python2.7/dist-packages/`.

From what i remember, is because the python-<packages> come with libraries(dependencies) that allows new versions be installed by pip. 

> Some of the package versions in REQUIREMENTS are current, others are not. I don't understand the policy -.-
> 

The versions there are the versions which installed without any problems. If you see any version that can be upgraded, change it ;)

> You use the commands `python2.7` and `python3.4`. If we stick to this convention, we also have to change some shebangs and the start command in `intelmqctl`.

sebix:
> From what i remember, is because the python- come with libraries(dependencies) that allows new versions be installed by pip. 
Holding back new versions does not solve compatibility issues in the long term. If you do encounter bugs in our software with specific library versions, please open an issue with a problem description.

SYNchroACK:
No. Thats not the case. I agree with you when you say 'Holding back new versions does not solve compatibility issues in the long term. If you do encounter bugs in our software with specific library versions, please open an issue with a problem description.'. Other thing is when a new version of a library is broken, not our solution. So, again, if you see any possibility to increase the current versions in REQ and setup.py propose. ;) ty

sebix:
> Other thing is when a new version of a library is broken, not our solution.

It's not always clear, who is right in these cases. Maybe a bug was fixed and intelmq relied on the buggy behavior? Maybe an incompatible API change, so we should adapt our API calls? Maybe the bug is not know yet by the maintainers?

In all cases, the bug should be reported, either here or at the modules bug tracker.

SYNchroACK:
>> Other thing is when a new version of a library is broken, not our solution.
>
>It's not always clear, who is right in these cases. Maybe a bug was fixed and intelmq relied on the buggy behavior? Maybe an incompatible API change, so we should adapt our API calls? Maybe the bug is not know yet by the maintainers?
>
>In all cases, the bug should be reported, either here or at the modules bug tracker.

Thats right and thats the reason to keep with specific version. With this approach we always sure that someone in internet can install without any problems even if we have a bug or a dependency has a bug. In dev branch, we can always use >= to test the next versions...

Conclusion: merged

Issue #338, opened 2015-09-19
CleanMX Phishing Parser Bot - wrong keys


SYNchroACK:
```
2015-09-19 01:19:12,339 - cleanmx-phishing-parser - ERROR - Bot has found a problem.
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/intelmq-1.0.0-py3.4.egg/intelmq/lib/bot.py", line 117, in start
    self.process()
  File "/usr/local/lib/python3.4/dist-packages/intelmq-1.0.0-py3.4.egg/intelmq/bots/parsers/cleanmx/parser_phishing.py", line 60, in process
    key = COLUMNS[key]
KeyError: None
2015-09-19 01:19:12,564 - cleanmx-phishing-parser - INFO - Last Correct Message(event): None.
2015-09-19 01:19:12,653 - cleanmx-phishing-parser - INFO - Current Message(event): {'time.observation': '2015-09-19T00:19:11+00:00', 'feed.name': 'CleanMX', 'raw': 'ImxpbmUiLCJpZCIsImZpcnN0dGltZSIsImxhc3R0aW1lIiwicGhpc2h0YW5rIiwidmlydXNuYW1lIiwidXJsIiwicmVjZW50IiwicmVzcG9uc2UiLCJpcCIsInJldmlldyIsImRvbWFpbiIsImNvdW50cnkiLCJzb3VyY2UiLCJlbWFpbCIsImluZXRudW0iLCJuZXRuYW1lIiwiZGRlc2NyIiwibnMxIiwibnMyIiwibnMzIiwibnM0IiwibnM1IgoiMSIsIjY1MDUyODciLCIyMDE1LTA5LTE5IDAxOjM3OjI4IiwiMTk3MC0wMS0wMSAwMTowMDowMCIsIiIsIiIsImh0dHA6Ly92ZXJpdHlpbnRlbC5jb20vY3NzL3BheXBhbC9iZj
```

sebix:
What was the fix of this issue? Why did it even occur?

SYNchroACK:
I think the problem was mentioned in #311 ...but ok, now this is the issue with the format problem.

sebix:
This error occurs, when a value appears in a column without a header, e.g.:

    id,foo,bar
    1,"some","value","error"

Will provide a fix, which gives a warning in these cases, so we can have a more detailed look. The current file from cleanmx does not have these values.

Issue #326, opened 2015-09-15
ENH: Error Handling


SYNchroACK:
```
2015-09-15 12:26:31,922 - hphosts-parser - INFO - Pipeline ready.
2015-09-15 12:26:31,922 - hphosts-parser - INFO - Start processing.
2015-09-15 12:26:31,922 - hphosts-parser - DEBUG - Receiving Message.
2015-09-15 12:26:50,997 - hphosts-parser - DEBUG - Receive message u'{"raw": "IyBocEhvc3RzIGxhc3QgdXBkYXRlZCBvbjoJCTIxLzA4LzIwMTUNCiMgaHBIb3N0cyBsYXN0IHZlcmlmaWVkIGJ5IFN0ZXZlbiBCdXJuOgkyMS8wOC8yMDE1DQojDQojIElNUE9SVEFOVDoJUmVuYW1lIHRoaXMgZmlsZSB0byAiSE9TVFMiIChubyAudHh0IGV4dGVuc2lvbikNCiMNCiMgU3VwcG9ydDoJaHR0cDovL215c3RlcnlmY20uY28udWsvP21vZGU9Y29udGFjdA0KIwkJaHR0cDovL2ZvcnVtLmhvc3RzLWZpbGUubmV0DQojDQojIERvd25sb2FkOglodHRwOi8vaG9zdHMtZmlsZS5uZXQvP3M9RG93bmxvYWQNCiMgTWlycm9yczoJaHR0cDovL2hvc3RzLWZpbGUubmV0Lz9zPUhlbHAjZGxtaXJyb3JzDQojDQojIEludGVybWl0dGVudCB1cGRhdGV'...
2015-09-15 12:26:58,873 - hphosts-parser - ERROR - Bot has found a problem.
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/intelmq-1.0.0-py2.7.egg/intelmq/lib/bot.py", line 117, in start
    self.process()
  File "/usr/lib/python2.7/site-packages/intelmq-1.0.0-py2.7.egg/intelmq/bots/parsers/hphosts/parser.py", line 45, in process
    event.add("source.fqdn", values[1], sanitize=True)
  File "/usr/lib/python2.7/site-packages/intelmq-1.0.0-py2.7.egg/intelmq/lib/message.py", line 108, in add
    raise exceptions.InvalidValue(key, value, reason=valid_value[1])
[...]

PR #341, opened 2015-09-22
Fixes from last days

sebix:
* dragon research parsers: protocols added to events
* fix for #326, bot does not dump message
* remove zmq as required dependency
* fix for #338, additional values in csv
* DOC: meaning of REQUIREMENTS clarified
* DOC: intelmqdump in User-Guide
* utils/logger can skip stream logger


SYNchroACK:
nice!

Issue #320, opened 2015-09-12
Timefilter bot


robcza:
Lots of feeds provide historical IoCs and there could be several reasons not to process them. You may want to exclude such obsolete events from the output and it could actually improve the performance a little if it would be possible to exclude those events before passing them to the pipeline of experts, hence the ideal positioning of the bot would be right after the parsers, even before deduplicator.

My proposal is to prepare an expert bot with two specific runtime parameters: ```not_before``` and ```not_after``` This should allow even configurations to gather IoCs from specific timeframe.
The timestamp for the comparison would be just and only ```time.source```

It could be completely independent new bot or it could be an extension to the existing filter bot. What do you think?

SYNchroACK:
@robcza I think it should be an extension for the existing filter bot. :)

aaronkaplan:
On Sep 13, 2015, at 4:44 AM, Tomás Lima <notifications@github.com> wrote:

> @robcza I think it should be an extension for the existing filter bot. :)
> 

IMHO: either way - if it is an extension or its own bot, I think it's an excellent idea.

robcza:
Thank you for the comments, will focus on extending the recent filter bot.

sebix:
I would implement it with relative times, i.e. timedeltas (module [datetime.timedelta](https://docs.python.org/3/library/datetime.html#datetime.timedelta)). So e.g.:

no before: 1 month
not after: None (deactivating check, same as empty)

aaronkaplan:
On Sep 13, 2015, at 9:25 PM, Sebastian <notifications@github.com> wrote:

> I would implement it with relative times, i.e. timedeltas (module datetime.timedelta). So e.g.:
> 
> no before: 1 month
> not after: None (deactivating check, same as empty)
> 

makes sense to me, good idea!

SYNchroACK:
+1

robcza:
I like it @sebix will do

sebix:
Implemented by #349

PR #344, opened 2015-09-23
Documentation improvements and parameters harmonization

SYNchroACK:

PR #349, opened 2015-10-01
time-based filtering

robcza:
filter expert modification and test, as discussed previously in here: https://github.com/certtools/intelmq/issues/320

Parameters for time based filtering:
* not_before - events before this time will be dropped
* not_after - events after this time will be dropped

Both parameters accept string values describing absolute or relative time:
* absolute
 * basically anything parsable by datetime parser, eg. "2015-09-012T06:22:11+00:00"
 * time.source taken from the event will be compared to this value to decide the filter behavior
* relative
 * accepted string formatted like this "<integer> <epoch>", where epoch could be any of following strings (could optionally end with trailing 's'): hour, day, week, month, year
 * time.source taken from the event will be compared to the value (now - relative) to decide the filter behavior

Examples of time filter definition:
* ```"not_before" : "2015-09-012T06:22:11+00:00"``` events older than the specified time will be dropped
* ```"not_after" : "6 months"``` just events older than 6 months will be passed through the pipeline


SYNchroACK:
@robcza Awesome! Thank you very much.
Please check this link https://travis-ci.org/certtools/intelmq/jobs/83119441 to fix the pull req.

robcza:
will do :)

intelmq/bots/experts/filter/expert.py
```diff
@@ -3,20 +3,75 @@
 import sys
 
 from intelmq.lib.bot import Bot
+from intelmq.lib.harmonization import DateTime
+from datetime import datetime, timedelta
+from dateutil import parser
+import re
```
sebix: Is a built-in packages, should go in line before sys-import. You may use `isort -rc filename` (python package from pip)

intelmq/bots/experts/filter/expert.py
```diff
@@ -3,20 +3,75 @@
...
 
+class FilterExpertBot(Bot):
```
sebix: PEP8: 2 newlines before definitions

intelmq/tests/bots/experts/filter/test_expert.py
```diff
@@ -0,0 +1,44 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+import json
+import unittest
+
+import intelmq.lib.test as test
+from intelmq.bots.experts.filter.expert import FilterExpertBot
+
```
sebix: second newline missing here.

robcza:
@SYNchroACK passed! :)
@sebix thank you for the comments, using pyCharm now to help me with the pep8 issues

sebix:
With #344 the bots docs have been migrated from separate READMEs to the central docs/BOTS.md file.

I will test the code on Tuesday.

mauroasilva:
Sorry for the late reply, but I think the time key used for filtering should be configurable. Something like time_filter_key, which could be any of the timestamps present on the message. It makes the filtering more flexible and it's pretty easy to replace the current static field for a parameter.

aaronkaplan:
how's the status on this @robcza ? Do you think you are ready for the PR? Can we test?

intelmq/bots/experts/filter/expert.py
```diff
@@ -1,22 +1,79 @@
...
+from dateutil import parser
 from intelmq.lib.bot import Bot
+from intelmq.lib.harmonization import DateTime
+
 
+class FilterExpertBot(Bot):
 
-class FilterBot(Bot):
+    # number of minutes in time units
+    timespans = ({'hour': 60, 'day': 24*60, 'week': 7*24*60, 'month': 30*24*60, 'year': 365*24*60})
```
sebix: Split into multiple lines and remove the redundant round brackets.

intelmq/bots/experts/filter/expert.py
```diff
@@ -1,22 +1,79 @@
...
+        if len(result) == 1 and len(result[0]) == 2 and result[0][1] in FilterExpertBot.timespans:
+            return int(result[0][0]) * FilterExpertBot.timespans[result[0][1]]
+        else:
+            raise AttributeError("Could not process result of regex for attribute " + repr(relative_time))
+
+    # decide format of timefilter value and parse it
+    def parse_timeattr(self, time_attr):
+        try:
+            absolute = parser.parse(time_attr)
+        except:
```
sebix: Only catch ValueError here.

intelmq/bots/experts/filter/expert.py
```diff
@@ -1,22 +1,79 @@
...
+    # number of minutes in time units
+    timespans = ({'hour': 60, 'day': 24*60, 'week': 7*24*60, 'month': 30*24*60, 'year': 365*24*60})
+
+    # parse relative time attributes
+    @staticmethod
+    def parse_relative(relative_time):
+        try:
+            result = re.findall(r'(\d+)\s+(\w+)?s', relative_time, re.UNICODE)
+        except Exception as e:
+            raise AttributeError("Could not apply regex to attribute \"%s\" with exception %s",
```
sebix: This should be ValueError, also on line 29 and 38.

intelmq/bots/experts/filter/expert.py
```diff
@@ -1,22 +1,79 @@
...
+    def parse_timeattr(self, time_attr):
+        try:
+            absolute = parser.parse(time_attr)
+        except:
+            try:
+                relative = timedelta(minutes=self.parse_relative(time_attr))
+            except Exception as e:
+                self.logger.error("Could not parse \"%s\" as neither absolute or relative time, "
+                                  "exception raised: %s", time_attr, repr(e.args))
+                self.stop()
```
sebix: Error handling is done by the bots class, just raise the exception here.

intelmq/bots/experts/filter/expert.py
```diff
@@ -26,26 +83,57 @@ def process(self):
             self.acknowledge_message()
             return
 
-        if self.parameters.filter_action == "drop":
+        # time based filtering
+        if event.contains('time.source'):
+            try:
+                event_time = parser.parse(event.value('time.source')).replace(tzinfo=pytz.timezone('UTC'))
```
sebix: pytz is never imported

intelmq/tests/bots/experts/filter/test_expert.py
```diff
@@ -0,0 +1,44 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+import json
```
sebix: json is not used here

intelmq/tests/bots/experts/filter/test_expert.py
```diff
@@ -0,0 +1,44 @@
...
+
+class TestFilterExpertBot(test.BotTestCase, unittest.TestCase):
+    """
+    A TestCase for FilterExpertBot.
+    """
+
+    @classmethod
+    def set_bot(self):
+        self.bot_reference = FilterExpertBot
+        self.sysconfig = {'not_before': '2014-01-01T00:00:00+00:00',
```
sebix: trailing whitespace

sebix:
The tests are not complete. You only test one possibility. There are relative and absolute times and the result can also filter or keep. Results in 4 tests minimum. Then there's before and after, should both be at least tested once.

robcza:
Thank you @sebix I am aware of the incomplete tests, however, I had problems initializing the bot for the test for four different configurations. Could you please advice how to achieve that?

I'm going to fix the inline comments as well

sebix:
You can set `self.sysconfig = {'foo': 'bar'}` before running the bot using `self.run_bot()` in each test.

robcza:
Bah! :( locally the tests go through n1. Could there be an issue with the same class defined in multiple test scripts? Seemed to me as a nice way to describe different cases.

SYNchroACK:
@robcza Im not a unittest expert but It seems to me that you forge to add 'EXAMPLE_OUTPUT' in multiple files.

robcza:
The problem does not seem to be in the input output events, but in the bot sysconfig:
```
test-bot: DEBUG: System configuration: parameter 'not_before' loaded  with value '2014-01-01T00:00:00+00:00'.
test-bot: DEBUG: System configuration: parameter 'not_after' loaded  with value '1 month'.
```

This is the same bot configured with two different values, however none of the tests assigns two different attributes to a test. @sebix can I ask you for a help. I'm worried the way I store individual tests is not right.

intelmq/bots/experts/filter/expert.py
```diff
@@ -1,22 +1,78 @@
...
 from __future__ import unicode_literals
+
+import re
 import sys
+import pytz
+from datetime import datetime, timedelta
 
+from dateutil import parser
 from intelmq.lib.bot import Bot
+from intelmq.lib.harmonization import DateTime
```
sebix: imported, but unused

intelmq/bots/experts/filter/expert.py
```diff
@@ -1,22 +1,78 @@
...
+
+    # decide format of timefilter value and parse it
+    def parse_timeattr(self, time_attr):
+        try:
+            absolute = parser.parse(time_attr)
+        except:
+            try:
+                relative = timedelta(minutes=self.parse_relative(time_attr))
+            except Exception as e:
+                raise e
```
sebix: Why do you catch the error and raise it immediatly?

intelmq/bots/experts/filter/expert.py
```diff
@@ -1,22 +1,78 @@
...
 
-class FilterBot(Bot):
+    # number of minutes in time units
+    timespans = ({'hour': 60, 'day': 24*60, 'week': 7*24*60, 'month': 30*24*60, 'year': 365*24*60})
+
+    # parse relative time attributes
+    @staticmethod
+    def parse_relative(relative_time):
+        try:
+            result = re.findall(r'(\d+)\s+(\w+[^s])s?', relative_time, re.UNICODE)
```
sebix: `'-2 months'` matches and becomes `[('2', 'month')]`. Add markers for begin and end of string: `r'^(\d+)\s+(\w+[^s])s?$'`.

intelmq/bots/experts/filter/expert.py
```diff
@@ -1,22 +1,78 @@
...
+        except Exception as e:
```
sebix: Which exceptions do you expect here?

intelmq/bots/experts/filter/expert.py
```diff
@@ -1,22 +1,78 @@
...
+            raise AttributeError("Could not apply regex to attribute \"%s\" with exception %s",
```
sebix: Definitely not an Attribute error. If the input is invalid, it's a ValueError. See also comment above, which error can occur here?

intelmq/bots/experts/filter/expert.py
```diff
@@ -1,22 +1,78 @@
...
+                                 repr(relative_time), repr(e.args))
+        if len(result) == 1 and len(result[0]) == 2 and result[0][1] in FilterExpertBot.timespans:
+            return int(result[0][0]) * FilterExpertBot.timespans[result[0][1]]
+        else:
+            raise AttributeError("Could not process result of regex for attribute " + repr(relative_time))
```
sebix: As above, this is a ValueError (invalid value in input):
https://docs.python.org/3/library/exceptions.html#AttributeError
> exception AttributeError

> Raised when an attribute reference (see Attribute references) or assignment fails. (When an object does not support attribute references or attribute assignments at all, TypeError is raised.)

intelmq/tests/bots/experts/filter/test_expert_relative_after.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @classmethod
+    def set_bot(self):
+        self.bot_reference = FilterExpertBot
+        self.input_message = EXAMPLE_INPUT
+        self.sysconfig = {'not_after': '1 month'}
+
+    def test_Relative_After(self):
+        self.run_bot()
+        self.assertOutputQueueLen(1)
```
sebix: Please use `self.assertMessageEqual(0, EXAMPLE_INPUT)`, so also the content of the message is checked. (Input = output here)

robcza:
@sebix Thank you for the comments, appreciate it

intelmq/bots/experts/filter/expert.py
```diff
@@ -1,22 +1,74 @@
...
+    timespans = {'hour': 60, 'day': 24*60, 'week': 7*24*60,
+                 'month': 30*24*60, 'year': 365*24*60}
+
+    # parse relative time attributes
+    @staticmethod
+    def parse_relative(relative_time):
+        try:
+            result = re.findall(r'^(\d+)\s+(\w+[^s])s?$', relative_time, re.UNICODE)
+        except Exception as e:
+            raise AttributeError("Could not apply regex to attribute \"%s\" with exception %s",
```
sebix: ValueError please here too.

robcza:
Hallelujah, thank you for the help with test @sebix 
Is it ok, to keep the adjusted test.py lib in this pull request?

intelmq/bots/experts/filter/expert.py
sebix: As already said, we don't want to catch all Exceptions here (no IOError, MemoryError, OverflowError, SystemError, SystemExit etc.)

sebix:
The code itself is fine, but many lines are still too long. There's also still one too broad except-clause.

robcza:
@sebix Ok, I understand the exception handling process now and fixed the broad except-clauses.

One thing I don't really understand is the line length culprit. My pyCharm studio says line length is pep8 compliant, you say it is not - who is right?

aaronkaplan:
> On 09 Dec 2015, at 13:19, Robert Šefr <notifications@github.com> wrote:
> 
> @sebix Ok, I understand the exception handling process now and fixed the broad except-clauses.
> 
> One thing I don't really understand is the line length culprit. My pyCharm studio says line length is pep8 compliant, you say it is not - who is right?
> 
Does your pyCharm read a .pep8 file which I added?

I have to admit, I don't care so much if its' 80 chars / line or not. for me this is a good exception to full PEP8 compliancy.
Some of the code will look extremely ugly if you *have to* break a line at 80 chars. See for example the n6stomp parser.

robcza:
I have 120 columns right now and seems good for readability on github (github has 126 I believe), but is not fully compliant with pep8 as stated in developers guide in documentation, however there is this .pep8 ```max-line-length = 132```

I vote for 120, however I'll be happy with any other reasonable final statement to have this sorted out.

aaronkaplan:
> On 09 Dec 2015, at 14:51, Robert Šefr <notifications@github.com> wrote:
> 
> I have 120 columns right now and seems good for readability on github (github has 126 I believe), but is not fully compliant with pep8 as stated in developers guide in documentation, however there is this .pep8 max-line-length = 132
> 
> I vote for 120, however I'll be happy with any other reasonable final statement to have this sorted out.
> 

good enough for me.

PR #370, opened 2015-10-14
Error reporting

e3rd:
When developing a new module, it took me hours to locate where the error is. Maybe it would be useful if this kind of problem would be written in logs too. (Up to the present, error reporting is devnulled.)


sebix:
Piping the stdout and stderr to a file would result in a log file for all bots. Not only the startup messages are logged, but also all other further messages.

Maybe a debug switch for intelmqctl could help in these cases, which does not deattach the process and shows all log output directly.

How to run a bot individually for debugging purpose is documented: https://github.com/certtools/intelmq/blob/master/docs/Developers-Guide.md#bot-developer-guide Perhaps the usage scenario can be emphasized better.

e3rd:
You're right, the debug switch would be far better.
I'm closing this.

PR #381, opened 2015-10-28
ENH: Accuracy feature for all feeds

mauroasilva:
This pull request adds an accuracy feature that is meant to represent the amount of false positives the feed has. Say you have a feed with 10 false positives for every 100 events, than you would give it a 90 in the accuracy parameter (as in 90%).

The reason I put a quantitative value (0 to 100), instead of a qualitative value (low, medium, high), was because I was thinking that, in the future, a feedback loop could be done in which the operators give feedback to the system regarding whether a given event is correctly identified or not, and that could be used to automatically calculate the feed's accuracy.

Although it is not hard to change it into a qualitative value if you guys prefer. What do you think is the best option?

### DIFF
--- a/intelmq/conf/harmonization.conf
+++ b/intelmq/conf/harmonization.conf
@@ -127,6 +127,10 @@
             "description": "All anecdotal information, which cannot be parsed into the data harmonization elements. E.g. os.name, os.version, user_agent. TODOs: Must be JSON encoded for machine readability.",
             "type": "String"
         },
+        "feed.accuracy": {
+            "description": "A float between 0 and 100 that represents how accurate the data in the feed is",
+            "type": "Accuracy"
+        },
         "feed.code": {
             "description": "Code name for the feed, e.g. DFGS, HSDAG etc.",
             "length": 100,
--- a/intelmq/conf/harmonization.conf
+++ b/intelmq/conf/harmonization.conf
@@ -311,6 +311,10 @@
             "description": "The URL of a given abuse feed, where applicable",
             "type": "URL"
         },
+        "feed.accuracy": {
+            "description": "A float between 0 and 100 that represents how accurate the data in the feed is",
+            "type": "Accuracy"
+        },
         "raw": {
             "description": "The original raw and unparsed data encoded in base64.",
             "length": 100000000,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ v1.0 (in development, master branch)
 - STY: code obeys PEP8, new code should always be properly formatted
 - ENH: More code is Python 3 compatible
 - DOC: Updated user and dev guide
+- ENH: New parameter and field named accuracy to represent the accuracy of each feed
 
 ### Harmonization
 - ENH: Additional data types: integer, float and Boolean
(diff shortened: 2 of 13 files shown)


intelmq/conf/harmonization.conf
```diff
@@ -311,6 +311,10 @@
             "description": "The URL of a given abuse feed, where applicable",
             "type": "URL"
         },
+        "feed.accuracy": {
```
sebix: You also need it for events, not just reports.

SYNchroACK: @sebix in your opinion we should use the same procedure like the following?
https://github.com/certtools/intelmq/blob/master/intelmq/lib/message.py#L221

mauroasilva: @sebix you are right... If you check the message.py it already has that field being passed on to the events, but I forgot to put it into the harmonization document.

SYNchroACK: @sebix should we change the unittests to be compliant?

sebix:
- tests for the Accuracy type are missing
- Is the accuracy value mandatory for all collectors? Your modifications in the collectors depend on the existance of `self.parameters.accuracy`. If it does not exist, the bot will fail.

        AttributeError: 'Parameters' object has no attribute 'accuracy'
  - If yes, the accuracy can't be defined when spawning new bots with the manager, as accuracy is not defined in `intelmq/bots/BOTS`
  - If no, should there be a default value if nothing is given? Or does the value not exist in the reports and events?
- Please generate a new `docs/Harmonization-fields.md` using `intelmq/bin/intelmq_gen_harm_docs.py`
- document the addition in the changelog

intelmq/conf/harmonization.conf
sebix: The unittests are missing a check for the Event(report) feature, yes.

sebix:
Ok, fine for me. As it's assigned to you, feel free to merge now.

Please also note that you can always easily correct the last commit by using the `--amend` flag: `git commit --amend`.

mauroasilva:
Ok thanks for the heads up.

aaronkaplan:
Hi folks, especially @mauroasilva : did you see this ? https://github.com/MISP/misp-taxonomies . Especially the Admiralty Scale seems reasonable for me for "accuracy".
https://github.com/MISP/misp-taxonomies/tree/master/admiralty-scale

What do you think? Should we quantify the information into A,B,C,D,E,F? It would make sense to build ontop of existing schemas.

robcza:
Other scale being used is TLP https://www.us-cert.gov/tlp
For compatibility reasons, there should be straightforward guide to convert form 0-100 to TLP or Admiralty Scale, yet IntelMQ could use 0-100 in the background as it easier to handle with different score factors influencing the final value.

mauroasilva:
@aaronkaplan that looks nice indeed and it is good that it is already a "standard". I begun this PR exactly by asking opinions about wether we should use qualitative values (like the one you described) or quantitative. The reason why I chose quantitative as I explained previously was to be able to have a kind of false positive percentage let's call it. Which could be calculated automatically based on feedback gathered by operators. I can change it if you guys feel it is the best approach, but that was my mindset when I implemented this feature.

@robcza I think TLP will eventually have to be added to IntelMQ but it is not really an accuracy measure, but a disclosure "range".

wagner-certat:
@robcza How is the traffic light protocol related to accuracy?

robcza:
@wagner-certat that was a misunderstanding as @mauroasilva pointed out

PR #386, opened 2015-11-03
Expert timefilter (debugging only)

sebix:
Just for debugging test failures on travis in #349

Don't merge!

### DIFF
--- a/intelmq/bots/experts/filter/README.md
+++ b/intelmq/bots/experts/filter/README.md
@@ -1,8 +1,25 @@
 ### Filter Bot
 
-A simple filter.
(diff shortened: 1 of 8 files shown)


robcza:
@sebix any progress on this debug?

sebix:
@robcza See this logline: https://travis-ci.org/certtools/intelmq/jobs/92923882#L1761 The `not_before` was set from a previous run, it has been saved in the bot config (this logic here: https://github.com/certtools/intelmq/blob/master/intelmq/lib/test.py#L50). So add the `.copy()` method and it will work.

Was partially (?) fixed in master.

PR #387, opened 2015-11-03
added malware.name and test for abusech parsers

robcza:
- added malware.name to both parsers
- created test for botha parsers
- fixed behaviour for empty messages (thank you tests)


aaronkaplan:
thank you @robcza

Issue #374, opened 2015-10-22
BUG: parser dismissing whole report on exception


robcza:
I have a trouble with hphosts-parser for several days now. First record in the list is ```127.0.0.1	-ads.avast.dwnldfr.com``` which is obviously malformed fqdn and parser raises an exception. So far that is an expected behavior.
The think is, I would expect parser to continue with following lines and not to purge the rest of the report. I would suggest parser to catch the sanitization exception and log it instead of passing them further and continue with the parsing cycle.

```
2015-10-22 12:23:32,131 - hphosts-parser - DEBUG - Receiving Message.
2015-10-22 12:23:37,586 - hphosts-parser - DEBUG - Receive message u'{"raw": "IyBocEhvc3RzIGxhc3QgdXBkYXRlZCBvbjoJCTExLzEwLzIwMTUNCiMgaHBIb3N0cyBsYXN0IHZlcmlmaWVkIGJ5IFN0ZXZlbiBCdXJuOgkxMS8xMC8yMDE1DQojDQo
jIElNUE9SVEFOVDoJUmVuYW1lIHRoaXMgZmlsZSB0byAiSE9TVFMiIChubyAudHh0IGV4dGVuc2lvbikNCiMNCiMgU3VwcG9ydDoJaHR0cDovL215c3RlcnlmY20uY28udWsvP21vZGU9Y29udGFjdA0KIwkJaHR0cDovL2ZvcnVtLmhvc3RzLWZpbGUubmV0DQojDQojIERv
d25sb2FkOglodHRwOi8vaG9zdHMtZmlsZS5uZXQvP3M9RG93bmxvYWQNCiMgTWlycm9yczoJaHR0cDovL2hvc3RzLWZpbGUubmV0Lz9zPUhlbHAjZGxtaXJyb3JzDQojDQojIEludGVybWl0dGVudCB1cGRhdGV'...
2015-10-22 12:23:38,021 - hphosts-parser - ERROR - Bot has found a problem.
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/intelmq-1.0.0-py2.7.egg/intelmq/lib/bot.py", line 95, in start
    self.process()
  File "/usr/local/lib/python2.7/dist-packages/intelmq-1.0.0-py2.7.egg/intelmq/bots/parsers/hphosts/parser.py", line 45, in process
    event.add("source.fqdn", values[1], sanitize=True)
[...]

PR #405, opened 2015-11-18
Modified sanitation exception handling

robcza:
Proposed this change based on this issue: https://github.com/certtools/intelmq/issues/374
Actually running this branch for quite a while and seems stable. I can see one possible flaw, that just the last exception is recorded this way.

Should I have list of exceptions there and add them as they appear? That would mean to log all of them in the end.

### DIFF
--- a/intelmq/lib/bot.py
+++ b/intelmq/lib/bot.py
@@ -221,6 +221,10 @@ class Bot(object):
             self.logger.warning("Sending Message: Empty message found.")
             return False
 
+        if message.exception:
+            self.logger.warning(message.exception)
+            return False
+
         self.logger.debug("Sending message.")
         self.message_counter += 1
         if self.message_counter % 500 == 0:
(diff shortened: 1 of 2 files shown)


robcza:
Ok, I get it, it breaks the tests actually, as it is not raising the exception right away. What do you think? Is this the right way to go and tests have to be changed or is this a wrong direction?

sebix:
I'm not sure if this actually solves the issue. I would have just put a try-except in the parsers' loops and called a modified dump routine for single lines that are malformed/problematic. You just saved the error message some where. It will be overridden and forgotten. The malformed data will never be reported, data will be silently thrown away.

mauroasilva:
I have to agree with @sebix ... I think the best approach would be to have something similar to the already implemented message dump that dumps problematic lines. That would actually also help with the #406 issue.

robcza:
@sebix try-except in the parsers is another complexity for the coder, that is why I tried to move that on the lib level. Malformed data are logged in my code, not actually forgotten, however not really prepared to reprocess.

Could we dump the message without changing anything on the parser level?

PR #414, opened 2015-12-13
BOT: url2fqdn expert

robcza:
expert bot to parse fqdn from the url and add it to the event

### DIFF
--- a/intelmq/bots/experts/url2fqdn/expert.py
+++ b/intelmq/bots/experts/url2fqdn/expert.py
@@ -2,12 +2,12 @@
 
 from __future__ import unicode_literals
 import sys
-import re
 try:
     from urllib.parse import urlparse
 except ImportError:
     from urlparse import urlparse
 from intelmq.lib.bot import Bot
+import intelmq.lib.harmonization
 
 
 class Url2fqdnExpertBot(Bot):
@@ -27,9 +27,8 @@ class Url2fqdnExpertBot(Bot):
                 continue
 
             hostname = urlparse(event.value(key_url)).hostname
-            if not re.compile("^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$").match(hostname):
-                event.add(key_fqdn, hostname, sanitize=True,
-                          force=True)
+            if intelmq.lib.harmonization.FQDN.is_valid(hostname, sanitize=True):
+                event.add(key_fqdn, hostname, sanitize=True, force=True)
 
         self.send_message(event)
         self.acknowledge_message()
--- a/intelmq/bots/experts/url2fqdn/expert.py
+++ b/intelmq/bots/experts/url2fqdn/expert.py
@@ -2,7 +2,11 @@
 
 from __future__ import unicode_literals
 import sys
-from urlparse import urlparse
+import re
+try:
+    from urllib.parse import urlparse
+except ImportError:
+    from urlparse import urlparse
 from intelmq.lib.bot import Bot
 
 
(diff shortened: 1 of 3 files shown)


intelmq/bots/experts/url2fqdn/expert.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+import sys
+from urlparse import urlparse
```
sebix: Use these lines for urlparse import: https://github.com/certtools/intelmq/blob/master/intelmq/lib/harmonization.py#L34
This will work in all versions.

intelmq/tests/bots/experts/url2fqdn/test_expert.py
```diff
@@ -0,0 +1,43 @@
...
+from __future__ import unicode_literals
+
+import os
+import unittest
+
+import intelmq.lib.test as test
+from intelmq.bots.experts.url2fqdn.expert import Url2fqdnExpertBot
+
+EXAMPLE_INPUT = {"__type": "Event",
+                 "source.url": "http://sample.url.com/something/index.php",
```
sebix: Please don't use private domains, example.com/org/net are reserved: https://tools.ietf.org/html/rfc2606

sebix:
Think of this one:

    urlparse.urlparse('http://10.0.0.1/').hostname
    '10.0.0.1'

Is this a valid FQDN?

robcza:
good point with the IP @sebix, will fix the other comments as well. thank you

robcza:
This PR should be ready

intelmq/bots/experts/url2fqdn/expert.py
```diff
@@ -0,0 +1,39 @@
...
+        for key in ["source.", "destination."]:
+
+            key_url = key + "url"
+            key_fqdn = key + "fqdn"
+            if not event.contains(key_url):
+                continue
+
+            hostname = urlparse(event.value(key_url)).hostname
+            if intelmq.lib.harmonization.FQDN.is_valid(hostname, sanitize=True):
+                event.add(key_fqdn, hostname, sanitize=True, force=True)
```
sebix: Could the `force` parameter be a configuration option? Or, at least, be mentioned in the bot description in `BOTS` file.

aaronkaplan: > On 19 Jan 2016, at 09:49, Sebastian <notifications@github.com> wrote:
> 
> In intelmq/bots/experts/url2fqdn/expert.py:
> 
> > +        event = self.receive_message()
> > +
> > +        if event is None:
> > +            self.acknowledge_message()
> > +            return
> > +
> > +        for key in ["source.", "destination."]:
> > +
> > +            key_url = key + "url"
> > +            key_fqdn = key + "fqdn"
> > +            if not event.contains(key_url):
> > +                continue
> > +
> > +            hostname = urlparse(event.value(key_url)).hostname
> > +            if intelmq.lib.harmonization.FQDN.is_valid(hostname, sanitize=True):
> > +                event.add(key_fqdn, hostname, sanitize=True, force=True)
> 
> Could the force parameter be a configuration option? Or, at least, be mentioned in the bot description in BOTS file.

How about force=True as a a) global system option (then it applies to all bots' event.add() ) or b) as a per bot option?

adulau:
Maybe you should have a look at https://github.com/stricaud/faup by @stricaud it's usually faster and especially more accurate for parsing strange URLS and extracting each element.

robcza:
@adulau I believe faup won't be needed for the simple task of url2fqdn. Should we extend the the functionalities of the bot, I have to admit, faup looks really nice (e.g. I consider TLD extraction an uneasy task)

stricaud:
More than happy to spend some time with you on the problem you want to solve here. And please use the git version of faup, lots of improvements, the only thing that prevents me from the release are the tests I have not finished. Happy yo improve faup to make it more useful.

sebix:
What's the conclusion about the `force=True` issue?

Issue #306, opened 2015-08-31
IPv6 handling


sebix:
Currently, many bots do not support IPv6 addresses and are either failing or silently ignoring IPv6 values.

What is the projects policy for IPv6?

If it's not supported yet, there are two possible approaches:
- IPv6 can be ignored if not supported by bots.
- Bots may fail with IPv6 data.

SYNchroACK:
Approach proposed:
* identify which bots will crash with IPv6
* evaluate the complexity of the changes
* if we choose to go through IPv6, apply the changes to support IPv6

sebix:
Not all sources contacted by experts do support IPv6. What to do in these cases?

Ignore and raise a warning? (warning = unexpected, but handled behavior, see https://github.com/certtools/intelmq/blob/master/docs/Developers-Guide.md#log-levels)

aaronkaplan:
---
Mobile


> On 31.08.2015, at 11:42, Sebastian <notifications@github.com> wrote:
> 
> Not all sources contacted by experts do support IPv6. What to do in these cases?
> 


I believe a reasonable approach is:
* Check Which experts are essential (ip2asn for example) and make them ipv6 aware 
* all other bots: ignore v6 for the moment but raise a warning. 
* later fix this after the release

SYNchroACK:
@aaronkaplan agreed!

aaronkaplan:
okay, forgot that we already discussed this in August. Yes #401 can be closed.
Anyway, needs to be done.

sebix:
Still to do:

- [ ] tor nodes
- [ ] reverse dns lookuo
- [ ] abusix

PR #427, opened 2016-01-03
Universal csv parser

robcza:
Proposal for a first universal parser, this one for csv inputs. There is a huge space for improvement:
- opportunistic parsing implemented just for time and IP
  - in the time/IP columns try to find the value whatever it takes and use it
  - similar behavior could be implemented for other attributes
  - it could actually be implemented on the lib level (e.g. with ```opportunistic=True``` while adding a value to the event
- other universal parsers could be implemented (json, xml, freetext, etc.)


robcza:
@sebix you've referred to this type of parser as "generic" instead of "universal" and I liked it. Should I change it to generic? @aaronkaplan

aaronkaplan:
> On 09 Jan 2016, at 19:06, Robert Šefr <notifications@github.com> wrote:
> 
> @sebix you've referred to this type of parser as "generic" instead of "universal" and I liked it. Should I change it to generic? @aaronkaplan
> 
+1

sebix:
* generic is the better term here (use `git mv` to rename the files)
* add an entry in the BOTS file.
* your parser is not compatible to IPv6, but we do want to support it. See also #306 
* Document the features like ignoring comments (`#`) inline and in the BOTS

sebix:
Tests are fine, failures are caused by some changes in DNS/whois.

sebix:
I adapted the code a bit to work with current master, after all my changes :) : https://github.com/sebix/intelmq/commit/75e376679f6a6c8fc3709ada0de6a6a08f69f44e

robcza:
thank you @sebix 
I hope to introduce improvements to this bot step by step, however any proposals are welcome

PR #428, opened 2016-01-10
collector-http autounzip

robcza:
Should there be a zip archive downloaded by the collector, unzip all files and create a report for each of them and log the unzipped file names.

### DIFF
--- a/intelmq/bots/collectors/http/collector_http.py
+++ b/intelmq/bots/collectors/http/collector_http.py
@@ -15,6 +15,8 @@ http_proxy, http_ssl_proxy: string
 from __future__ import unicode_literals
 import requests
 import sys
+import io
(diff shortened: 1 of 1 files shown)


robcza:
@sebix is the current solution with imports proper and valid for both python2 and python3? Because if yes, I would leave it as it is.

sebix:
There's no `StringIO` or `cStringIO` module in current Python :) They only exist in old legacy Python.

robcza:
seems a valid reason :) will elaborate further, do not merge

robcza:
Seems ok to me now. Works smoothly, tested against both - zip and plaintext files. 
Mergeable from my point of view.

sebix:
Thanks, also tested successfully.

PR #431, opened 2016-01-10
Added initial contrib cron job scripts.

### LINKED ISSUE #408: geoip / asn lookup - keep the source data up-to-date
This is low priority issue, however I'm sure everyone will have to deal with this sooner or later. There is no mechanism in IntelMQ to keep local databases used for different sort of lookups up-to-date. E.g. you download maxmind geoip database, you run the IntelMQ botnets and several days later you run expert bot with outdated data.
- Should this be handled by IntelMQ itself? - I think yes, but it introduces another complexity
- If not, do you handle it byt a set of external script updating the databases and restarting the bots?

aaronkaplan:
Fixes #408

### DIFF
--- a/contrib/cron-jobs/pyasn
+++ b/contrib/cron-jobs/pyasn
@@ -31,4 +31,5 @@ rm -f $latest
 cp ipasn.dat  /opt/intelmq/var/lib/bots/asn_lookup/
 chown intelmq:intelmq /opt/intelmq/var/lib/bots/asn_lookup/ipasn.dat
 
+rm -f /tmp/pyasn
 
--- a/contrib/cron-jobs/pyasn
+++ b/contrib/cron-jobs/pyasn
@@ -1,0 +1,34 @@
+#!/bin/sh
+
+
+# How often should this run? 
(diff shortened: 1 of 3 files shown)


contrib/cron-jobs/pyasn
```diff
@@ -0,0 +1,34 @@
...
+  exit 255
+fi
+
+cd /tmp
+mkdir -p pyasn
+cd pyasn
+
+pyasn_util_download.py --latest
+
+latest=$(ls -tr | tail -n 1)
```
sebix: I'm feeling very uncomfortable.. We could filter for the filename ending at least.

contrib/cron-jobs/pyasn
```diff
@@ -0,0 +1,34 @@
...
+
+pyasn_util_convert.py --single $latest ipasn.dat
+rm -f $latest
+
+cp ipasn.dat  /opt/intelmq/var/lib/bots/asn_lookup/
```
sebix: This leaves a copy at `/tmp`, intended?

PR #441, opened 2016-01-26
Add coverage & codecov.io

Rafiot:


Rafiot:
I love gamification of testcase :)

PR #445, opened 2016-02-04
initial version of the STOMP output bot

aaronkaplan:
by @mauroasilva and aaron


aaronkaplan:
@sebix what test cases would make sense here?

sebix:
- [ ] I don't think we can do any tests at all here, except for importing the bot, so we can at least detect `ImportError`s and `SyntaxError`s, like here: https://github.com/certtools/intelmq/blob/master/intelmq/tests/bots/outputs/mongodb/test_output.py
- [ ] Please add a requirements file for the stomp package, like done here: https://github.com/sebix/intelmq/blob/bugfixes/intelmq/bots/outputs/mongodb/REQUIREMENTS.txt
- [ ] The content of the local `.gitignore` can be moved to the gobal one. `*.pyc` is already there, adding the keys and certs is a good idea

aaronkaplan:
Agreed to all points.

---
Mobile

PR #467, opened 2016-04-04
intelmq_psql_initdb.py JSON instead of String

dmth:
Hey folks,

as already discussed in one of our phone calls, we suggest to move change from String to JSON for the data in the "Extra" field.

To do so, we modified the file in order to create database columns of type
json instead of string.

This should enhance the processing speed of data which is stored in the
extras column.

ATTENTION: This is a schema-alteration of the EventDB.
If data already exists in the EventDB it might be required, that data is
converted from string to json.
Refer to: http://www.postgresql.org/docs/9.2/static/datatype-json.html

This change requires a PostgreSQL 9.2 or higher.

### DIFF
--- a/intelmq/bin/intelmq_psql_initdb.py
+++ b/intelmq/bin/intelmq_psql_initdb.py
@@ -30,7 +30,7 @@ def main():
     for field in DATA.keys():
         value = DATA[field]
 
-        if value['type'] in ('String', 'Base64', 'URL', 'FQDN', 'JSON',
+        if value['type'] in ('String', 'Base64', 'URL', 'FQDN',
                              'MalwareName', 'ClassificationType'):
             dbtype = 'varchar({})'.format(value.get('length', 2000))
         elif value['type'] in ('IPAddress', 'IPNetwork'):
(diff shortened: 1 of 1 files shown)


aaronkaplan:
A Travis check failed: https://travis-ci.org/certtools/intelmq/jobs/120601871#L506

dmth:
I know, that's because of broken (?) testdata for the Cymru AS-Lookup:
Cymru Whois returns a different network than expected by the test.
```
u'destination.network': u'186.226.224.0/20', << Expected
u'destination.network': u'186.226.224.0/21', << Returned by Cymru
```

aaronkaplan:
On Mon, Apr 04, 2016 at 06:05:44AM -0700, Dustin Demuth wrote:
> I know, that's because of broken (?) testdata for the Cymru AS-Lookup:
> Cymru Whois returns a different network than expected by the test.
> ```
> u'destination.network': u'186.226.224.0/20', << Expected
> u'destination.network': u'186.226.224.0/21', << Returned by Cymru
> ```

okay. So how about fixing this in the PR as well? Then the PR will sail
through the automatic testing .
Or do you prefer a separate PR for this?

a.

sebix:
Don't mind the check. Will fix that.

Issue #473, opened 2016-04-13
Maxmind-geoip-expert won't start


chappi6:
Hi , 

I have a problem with the Maxmind-Geoip-expert . 
It wont start.... 
Actually there are no logs which describe the problem. 
The only entry i can find is that intelmqctl says maxmind geoip-expert is running 
The other experts and parsers are running well. 
I have taken the Geoip db file from the script which is under intelmq/contrib/cron-jobs/maxmind
Any ideas how to get logs out to find the problem?

sebix:
Do you have `geoip2`package installed?
Try running this as intelmq user: `python3 -m intelmq.bots.experts.maxmind_geoip.expert bot-id-here` and replace the bot with what you have configured.

chappi6:
jup geoip2 package was missing. 
Error from the command python3 -m intelmq.bots.experts.maxmind_geoip.expert
was 
Import Error: No module named 'geoip2'
I have no version geoip2 2.2.0 install 
if somebody needs it command for installing is: 
pip install geoip2 

is it possible to add the libary into the REQUIREMENTS file? So it would be possible to avoid this error ;)

Issue #475, opened 2016-04-13
Bots with dependencies: Don't fail on startup


sebix:
Some bots do have dependencies to external libraries, such as `geoip2`. Starting of these bots fails without error messages. The import should be done on initialization, after logging has been initialized.

PR #476, opened 2016-04-13
Update REQUIREMENTS

chappi6:
would avoid the problem 
https://github.com/certtools/intelmq/issues/473


sebix:
Please see
https://github.com/certtools/intelmq/blob/master/intelmq/bots/experts/maxmind_geoip/REQUIREMENTS.txt

We not have all packages that you might never use in this requirements file.

With #475 this will be easier to handle and detect (the issue has been raised earlier already).

chappi6:
Thx 
Seems I missed that file ;)

PR #483, opened 2016-04-16
Bitsight collector and parser

pedromreis:

PR #484, opened 2016-04-17
Bitsight bots

pedromreis:
Bitsight Alert Stream bots.
Tested with my feed, working.


intelmq/bots/BOTS
```diff
@@ -832,6 +840,11 @@
             "module": "intelmq.bots.parsers.autoshun.parser",
             "parameters": {}
         },
+        "Bitsight Ciberfeed Stream":{
+            "description":"Bitsight Ciberfeed parser is the bot reponsible to parse and sanitize the information",
+            "module":"intelmq.bots.parsers.bitsight.parser",
+            "parameters":{}
```
sebix: Here's a space missing

sebix:
As I pointed out in certtools/intelmq#483:

* Write tests for the parser, see for examples: https://github.com/certtools/intelmq/blob/master/intelmq/tests/bots/parsers/malwaredomainlist/test_parser.py
* Write a dummy test for the collector, see e.g. https://github.com/certtools/intelmq/blob/master/intelmq/tests/bots/collectors/http/test_collector.py 

When you are done, squelch all your commits into one or two by doing an interactive rebase: `git rebase -i master`

pedromreis:
Maybe need a more elegant solution on the " Bitsight Collector would not stop, not sure if found best " issue.

intelmq/bots/parsers/bitsight/parser.py
```diff
@@ -0,0 +1,70 @@
...
+                    event.add('source.port', value["remote_port"])
+                if "server_addr" in value:
+                    event.add('destination.ip', value["server_addr"])
+                if "server_port" in value:
+                    event.add('destination.port', value["server_port"])
+                if "server_name" in value:
+                    try:
+                        event.add('destination.fqdn', value["server_name"])
+                    except InvalidValue:
+                        event.add('destination.fqdn', "")
```
sebix: Just don't add the field. Replace the line with `pass`.

PR #495, opened 2016-04-21
Redis bot

pedromreis:
Redis bot impementation, outputs to a defined redis server, db  and queue

### DIFF
--- a/intelmq/tests/bots/outputs/redis/test_output.py
+++ b/intelmq/tests/bots/outputs/redis/test_output.py
@@ -32,28 +32,6 @@ EXAMPLE_EVENT = {"classification.type": "malware",
                  "W50cnlfbmFtZSI6IkRvbWluaWNhbiBSZXB1YmxpYyJ9fQ=="
                  }
 
-EXAMPLE_OUTPUT = {"classification.type": "malware",
-                  "__type": "Event",
-                  "feed.url": "http://alerts.bitsighttech.com:8080/stream?",
-                  "feed.name": "BitSight",
-                  "feed.accuracy": 100.0,
-                  "source.port": 65118,
-                  "source.ip": "152.166.119.2",
-                  "source.geolocation.country": "Dominican Republic",
-                  "malware.name": "salityp2p",
-                  "raw": "eyJ0cm9qYW5mYW1pbHkiOiJTYWxpdHlwMnAiLCJlbnYiOnsicmVtb3"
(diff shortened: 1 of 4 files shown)


sebix:
As I pointed out in one of your numerous previous pull requests, please add tests for the output bot.

pedromreis:
@sebix not forgotten, but since there is no test example for output bots, I'm trying to understand how can I build one. If I understand correctly, the test will look at the output queue to compare? Loaded from pipeline right?
I need to see this with more time. But I'm it.

sebix:
> @sebix <https://github.com/sebix> not forgotten, but since there is no
> test example for output bots, I'm trying to understand how can I build
> one. If I understand correctly, the test will look at the output queue
> to compare? Loaded from pipeline right?
> I need to see this with more time. But I'm it.
>
The bots needs to write to db 10 (used for all tests), and then you
check get the data with the redis-libs and compare the data for equality.

intelmq/tests/bots/outputs/redis/test_output.py
```diff
@@ -0,0 +1,88 @@
...
+                 "extra.non_ascii": "ççãããã\x80\ua000 \164 \x80\x80 abcd \165\166",
+                 "raw": "eyJ0cm9qYW5mYW1pbHkiOiJTYWxpdHlwMnAiLCJlbnYiOnsic"
+                 "mVtb3RlX2FkZHIiOiIxNTIuMTY2LjExOS4yIiwicmVtb3RlX3"
+                 "BvcnQiOiI2NTExOCIsInNlcnZlcl9hZGRyIjoiNTIuMTguMTk"
+                 "2LjE2OSIsInNlcnZlcl9wb3J0IjoiOTc5NiJ9LCJfdHMiOjE0"
+                 "NjExMDc3NjgsIl9nZW9fZW52X3JlbW90ZV9hZGRyIjp7ImNvd"
+                 "W50cnlfbmFtZSI6IkRvbWluaWNhbiBSZXB1YmxpYyJ9fQ=="
+                 }
+
+EXAMPLE_OUTPUT = {"classification.type": "malware",
```
sebix: This is exactly the same as EXAMPLE_EVENT. Why define it two times?

pedromreis:
hummm, no need now really, useful when testing. I'll remove that,

sebix:
Merged via 268bc88a153d462682703060d1c1f4e01a99c08e

PR #500, opened 2016-04-28
Overhaul of the XMPP-Collector and Output bot

dmth:
* The Collector-bot is now capable of joining Rooms
 * The Output-bot can send Messages to users (not rooms)

What we tested (and what worked)
 * Connect with the bot to a server
 * Join one room
 * Receive a message from a room
 * Receive a message from a user

What we didn't test:
 * Password protected rooms
 * "Real Data"
 * XMPP-Servers with funky SSL-settings

We used sleekxmpp 1.0.0-beta5, which is available from ubuntu 14.04 repositories for our test-setup.


intelmq/bots/collectors/xmpp/collector.py
```diff
@@ -1,10 +1,29 @@
-# -*- coding: utf-8 -*-
+"""
+XMPP Collector Bot
+Connects to a XMPP Server and a Room and reads data from the room.
+If no room is provided it collects events which were sent to the xmpp user
```
sebix: "no room" is equivalent to an empty string? I'm missing this information in the docs.

intelmq/bots/outputs/xmpp/output.py
```diff
@@ -9,45 +29,61 @@
...
+            # the message was sent successfully.
+            self.logger.debug("XMPP Trying to send Event: %r , to %r",
+                            jevent,
+                            receiver)
+            self.xmpp.send_message(mto=receiver, mbody=jevent)
+        except XMPPError as err:
+            self.logger.error('There was an error when sending the event')
+            self.logger.error(err.iq['error']['condition'])
 
         self.acknowledge_message()
```
sebix: I think this should go into the `else`-clause.

intelmq/bots/outputs/xmpp/output.py
```diff
@@ -1,4 +1,24 @@
...
+Parameters:
+xmpp_user: string
+xmpp_server: string
+xmpp_password: boolean
+# TODO xmpp_room: string
+# TODO xmpp_room_password: string
+# TODO xmpp_room_nick: string
+"""
+
 import json
```
sebix: not used

intelmq/bots/outputs/xmpp/output.py
```diff
@@ -9,45 +29,61 @@
...
-        self.xmpp.send_message(mto=self.parameters.xmpp_to_user + '@' + self.parameters.xmpp_to_server,
-                               mbody=event.to_json())
+        try:
+            # TODO: proper error handling. Right now it cannot be detected if
+            # the message was sent successfully.
+            self.logger.debug("XMPP Trying to send Event: %r , to %r",
+                            jevent,
+                            receiver)
+            self.xmpp.send_message(mto=receiver, mbody=jevent)
+        except XMPPError as err:
```
sebix: XMPPError is not defined

intelmq/bots/outputs/xmpp/output.py
```diff
@@ -1,4 +1,24 @@
-# -*- coding: utf-8 -*-
+"""
+XMPP Output Bot
+Connects to a XMPP Server and sends data to a user.
+
+TODO: Introduce Multi User Chat like in XMPP Collector
+
+Requires Python >= 3.4
+Requires sleekxmpp >= 1.0.0-beta5
```
sebix: `1.0.0-beta5` does not exist on pypi.
`1.0beta6.1` exists, but does not have IqError and IqTimeout
Also note, that `1.0~beta5-2` from trusty is not sufficient too.

intelmq/bots/outputs/xmpp/output.py
```diff
@@ -1,4 +1,24 @@
-# -*- coding: utf-8 -*-
+"""
+XMPP Output Bot
+Connects to a XMPP Server and sends data to a user.
+
+TODO: Introduce Multi User Chat like in XMPP Collector
+
+Requires Python >= 3.4
```
sebix: Why is it `>=3.4` only? What about 3.3?

intelmq/tests/bots/outputs/xmpp/test_output.py
```diff
@@ -0,0 +1,8 @@
+"""
+Test for the XMPP Output Bot
+"""
+
+import intelmq.bots.outputs.xmpp.output
+
+# This file is a stub.
+
```
sebix: One newline too much.

sebix:
One additional note on the style: The indentation of continuation lines is not comprehensible

intelmq/bots/collectors/xmpp/collector.py
dmth: Yes, that is equivalent to an empty string. Will add this information.

intelmq/bots/outputs/xmpp/output.py
dmth: > 1.0.0-beta5 does not exist on pypi.
But on Ubuntu 14.04 Repositories.

>  1.0beta6.1 exists, but does not have IqError and IqTimeout
That's funny because https://github.com/fritzy/SleekXMPP/blob/sleek-1.0-Beta6.1/sleekxmpp/exceptions.py has these classes!There seems to be a problem with pip?

intelmq/bots/outputs/xmpp/output.py
dmth: That's because I tested with 3.4 only. Do you have a target Platform for 3.3?

dmth:
Thanks for Reviewing.

Some things broke whilst merging with your changes.
I.e. those from intelmq/bots/outputs/xmpp/output.py 

I'm going to correct them asap.


> One additional note on the style: The indentation of continuation lines is not comprehensible
Do you have an example?

intelmq/bots/outputs/xmpp/output.py
dmth: Yes, makes sense.

intelmq/bots/collectors/xmpp/collector.py
```diff
@@ -21,32 +40,78 @@ def init(self):
...
+                room,
+                room_nick,
+                room_password,
+                logger):
         sleekxmpp.ClientXMPP.__init__(self, jid, password)
-        self.add_event_handler("session_start", self.session_start)
-        self.add_event_handler("message", self.message_logging)
+
         self.logger = logger
         self.logger.info("XMPP connected")
```
sebix: In case of connection issues, this line will be printed anyway. Could be replaced with `initiated`.

intelmq/bots/collectors/xmpp/collector.py
```diff
@@ -21,32 +40,77 @@ def init(self):
...
+            self.xmpp.register_plugin('xep_0045')  # Multi-User Chat
+            self.xmpp.add_event_handler("message", self.log_message)
+            self.xmpp.add_event_handler("groupchat_message", self.log_message)
+
     def killbot(self):
         self.xmpp.disconnect(wait=True)
         self.logger.info("Disconnected")
 
+    def log_message(self, msg):
+        self.logger.debug("XMPP Received Event: %r , from %r", msg['body'],
```
sebix: As the bot can also receive reports, the log may be confusing. We could call it just "stanza"?

intelmq/bots/outputs/xmpp/output.py
```diff
@@ -1,53 +1,87 @@
...
+            jevent = event.to_json()
+
+            try:
+                # TODO: proper error handling. Right now it cannot be
+                # detected if the message was sent successfully.
+                self.logger.debug("XMPP Trying to send Event: %r , to %r",
+                                  jevent,
+                                  receiver)
+                self.xmpp.send_message(mto=receiver, mbody=jevent)
+            except sleekxmpp.exceptions.XMPPError as err:
```
sebix: `sleekxmpp.exceptions` is not imported

intelmq/bots/collectors/xmpp/collector.py
```diff
@@ -21,32 +40,77 @@ def init(self):
...
+        # Try to decode the message as json, it might be an event.
+        # if this is not possible treat it as raw.
+        # TODO: There might be additional work to decode the message
+        # as base64
+        try:
+            event = json.loads(msg['body'])
+        except:
+            self.logger.info("XMPP Could not interpret the message as json. "
+                             "Treating as Raw")
+            self.logger.debug("XMPP I'm not able to tell if it is"
```
sebix: Please remove the `XMPP` at the beginning of all log outputs. The bot name will be inserted by another layer. Currently this is the outcome (without date etc):
> xmpp-collector: XMPP I'm not able to tell if it isbase64 or not!

And as you can see, there's a whitespace missing

intelmq/bots/collectors/xmpp/collector.py
```diff
@@ -21,32 +40,77 @@ def init(self):
...
+        raw_msg = None
 
-    def __init__(self, jid, password, send_message, logger):
+        # Try to decode the message as json, it might be an event.
+        # if this is not possible treat it as raw.
+        # TODO: There might be additional work to decode the message
+        # as base64
+        try:
+            event = json.loads(msg['body'])
+        except:
```
sebix: You want to except ValueErrors only here

sebix:
Replace `sleekxmpp.exception` with `sleekxmpp.exceptions`.

sebix:
A single SIGINT is not enough to stop the bot:

```python
<SIGINT>
xmpp-collector: Received KeyboardInterrupt.
xmpp-collector: Disconnecting from source pipeline.
xmpp-collector: Disconnecting from destination pipeline.
xmpp-collector: Bot stopped.
xmpp-collector: Exiting.
Exiting
<SIGINT>
Exception ignored in: <module 'threading' from '/usr/lib/python3.4/threading.py'>
  File "/usr/lib/python3.4/threading.py", line 1294, in _shutdown
    t.join()
  File "/usr/lib/python3.4/threading.py", line 1060, in join
    self._wait_for_tstate_lock()
  File "/usr/lib/python3.4/threading.py", line 1076, in _wait_for_tstate_lock
    elif lock.acquire(block, timeout):
KeyboardInterrupt
```

You wrote the function `killbot`, but it is never used. Please extend the `stop` function of the bot class to disconnect properly.

intelmq/bots/outputs/xmpp/output.py
```diff
@@ -1,53 +1,87 @@
...
+Requires sleekxmpp >= 1.0.0-beta5
+
+Copyright (C) 2016 by Bundesamt für Sicherheit in der Informationstechnik
+Software engineering by Intevation GmbH
+
+Parameters:
+xmpp_user: string
+xmpp_server: string
+xmpp_password: boolean
+# TODO xmpp_room: string
```
sebix: `xmpp_to_user` and `xmpp_to_server` are missing here.

intelmq/bots/collectors/xmpp/collector.py
dmth: > In case of connection issues, this line will be printed anyway.

True. Will be fixed.

intelmq/bots/collectors/xmpp/collector.py
dmth: Stanza is a valid suggestion, as this term is widely used with xmpp. Will fix that.

intelmq/bots/outputs/xmpp/output.py
dmth: From my POV it's not necessary as sleekxmpp is imported.

Are you experiencing problems with this LOC?

sebix:
Please test the bots again, including some pipeline inbetween (e.g. connect collector with output). There are several issues. E.g. You are sending dictionaries directly, which is not valid. Have a look at [`MessageFactory.unserialize`](https://github.com/certtools/intelmq/blob/master/intelmq/lib/message.py#L46) and other functions like this one.

intelmq/bots/outputs/xmpp/output.py
sebix: Not anymore, not sure why I had problems. Probably an older version.

intelmq/bots/outputs/xmpp/output.py
sebix: > > 1.0.0-beta5 does not exist on pypi.
> But on Ubuntu 14.04 Repositories.
Using `1.0~beta5-2` from trusty:
```python
>>> sleekxmpp.exceptions.IqError
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'IqError'
>>> sleekxmpp.__version__
'1.0beta5'
```
using `1.0beta6.1` via pip:
```python
>>> sleekxmpp.exceptions.IqError
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'IqError'
>>> sleekxmpp.__version__
'1.0beta6.1'
```
`1.0rc1` is the first version that works for me.

dmth:
@sebix Sebix, can you have a look at this PR again?
From my point of view, everything the XMPP-Collector reads should be written into the Raw Field.

A parser should now be responsible to convert this data into events. I attached a  naive (imagine the double pointed i here...) json-parser to do this.

aaronkaplan:
in addition to @sebix's comments, the travis-ci test fails. 
Please also supply a parser bot. Otherwise this is useless.
Also: I know that AH setups usually use an SSL enabled XMPP server. So, SSL support is needed.

intelmq/bots/collectors/xmpp/collector.py
```diff
@@ -21,48 +39,84 @@ def init(self):
...
-        self.logger.info("Disconnected")
-
+            # Add Handlers and register Plugins
+            self.xmpp.register_plugin('xep_0030')  # Service Discovery
+            self.xmpp.register_plugin('xep_0045')  # Multi-User Chat
+            self.xmpp.add_event_handler("message", self.log_message)
+            self.xmpp.add_event_handler("groupchat_message", self.log_message)
+
+    def stop(self):
+        self.xmpp.disconnect()
```
sebix: Just noticed that `stop()` could also be called when `self.xmpp` is not created yet. E.g.:
```bash
2016-05-17 11:39:00,845 - xmpp-collector - INFO - XMPPCollectorBot initialized with id xmpp-collector and version 3.4.3 (default, Oct 14 2015, 20:28:29).
2016-05-17 11:39:00,846 - xmpp-collector - INFO - Bot is starting.
2016-05-17 11:39:00,846 - xmpp-collector - ERROR - Pipeline configuration: no key 'xmpp-collector'.
2016-05-17 11:39:00,846 - xmpp-collector - ERROR - Bot initialization failed.
  File "/home/r/dev/intelmq/intelmq/lib/bot.py", line 64, in __init__
    self.load_pipeline_configuration()
  File "/home/r/dev/intelmq/intelmq/lib/bot.py", line 363, in load_pipeline_configuration
    self.stop()
  File "/home/r/dev/intelmq/intelmq/bots/collectors/xmpp/collector.py", line 60, in stop
    self.xmpp.disconnect()
AttributeError: 'NoneType' object has no attribute 'disconnect'

```

dmth: Nice catch!

intelmq/bots/collectors/xmpp/collector.py
```diff
@@ -21,48 +39,84 @@ def init(self):
...
+
+        # Read msg-body and add as raw to a new report.
+        # now it's up to a parser to do the interpretation of the message.
+        if raw_msg:
+            report = Report()
+            report.add("raw", raw_msg)
+            self.send_message(report)
+
+
+class XMPPClientBot(sleekxmpp.ClientXMPP):
```
sebix: Could you rename this to `XMPPClient`, as this class is not a bot? (also for output)

Also see https://github.com/certtools/intelmq/blob/master/intelmq/bin/intelmqctl.py#L298 (yes, it's not nice)

intelmq/bots/outputs/xmpp/output.py
```diff
@@ -1,53 +1,94 @@
...
         self.xmpp.connect(reattempt=True)
         self.xmpp.process()
 
     def process(self):
         event = self.receive_message()
 
+        receiver = self.parameters.xmpp_to_user + '@' \
+                   + self.parameters.xmpp_to_server
+
         if event is None:
```
sebix: This is obsolete now

intelmq/bots/parsers/json/parser.py
```diff
@@ -0,0 +1,54 @@
...
+        raw_decoded = utils.base64_decode(raw)
+        raw_json = None
+
+
+        self.logger.debug("Raw Report: " + raw)
+        self.logger.debug("Raw Decoded: " + raw_decoded)
+        try:
+            raw_json = json.loads(raw_decoded)
+            self.logger.debug("Raw Json: " + str(raw_json))
+        except:
```
sebix: What do you want to catch here?

intelmq/bots/parsers/json/parser.py
```diff
@@ -0,0 +1,54 @@
...
+            self.logger.error("Could not convert report to json")
```
sebix: `logger.exception` would also give the traceback and thus a more accurate error message

intelmq/bots/parsers/json/parser.py
```diff
@@ -0,0 +1,54 @@
...
+            raise
+
+        event = Event(raw_json)
+
+        #for key, value in raw_json:
```
sebix: Is this some remnant?

intelmq/bots/parsers/json/parser.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class JSONParserBot(Bot):
+
+    def process(self):
+        report = self.receive_message()
+
+        raw = report.get("raw")
+        raw_decoded = utils.base64_decode(raw)
+        raw_json = None
```
sebix: Why the initialization? AFAI understand the code, its superfluous.

intelmq/bots/parsers/json/parser.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+        self.logger.debug("Raw Report: " + raw)
```
sebix: Done by the bot-class too (truncated).

sebix:
I'm missing tests for the JSON-parser

intelmq/bots/parsers/json/parser.py
dmth: raw_decoded might not be valid json, that's why the try-catch block exists.

sebix: This gives a ValueError. We want to catch only this one. (not a KeyboardInterrupt for example)

sebix:
E.g. we have the following incoming message: `{"soure.fqnd": "example.com"}` (note the invalid key)
Then the Event is exactly this string. No observation time, invalid data.

Is this really the intention?

dmth:
@sebix what to do with this "old" PR? From my Point of view the code for the XMPP Collectors proposed here is "less broken" than the code which is currently upstream. Is there the possibility to apply this PR, except the Json-Parser stuff?

dmth:
This PR is deprecated.
Most recent and successfully tested code is in https://github.com/certtools/intelmq/pull/669

PR #503, opened 2016-04-29
Output Bot to UDP servers

pedromreis:
Simple bot to forward events to UDP servers, with basic configuration. Instructions and examples in README.md file.
Tested with rsyslog deamon.


intelmq/bots/outputs/udp/output.py
```diff
@@ -0,0 +1,56 @@
...
+from intelmq.lib.bot import Bot
+
+
+class UDPBot(Bot):
+
+    def init(self):
+        self.delimiter = self.parameters.field_delimiter
+        self.header = self.parameters.header
+        self.udp_host = self.parameters.ip
+        self.udp_port = self.parameters.port
```
sebix: Better use `int(self.parameters.port)` here, in case someone's saving it as port (intelmq-managers default type...)

intelmq/bots/outputs/udp/output.py
```diff
@@ -0,0 +1,56 @@
...
+            log_line += self.delimiter + key + ':' + str(value)
+
+        return log_line
+
+    def send(self, rawdata):
+        data = utils.encode(rawdata)
+        try:
+            self.udp.sendto(data, self.upd_address)
+        except:
+            self.logger.exception('UDPBot: failled to sent message to {}:{} !'
```
sebix: Remove the `UDPBot: ` here, the botname will be inserted by other components.

intelmq/bots/outputs/udp/output.py
```diff
@@ -0,0 +1,56 @@
...
+
+        if self.format.upper() == 'JSON':
+            self.send(self.header + ' ' + event.to_json())
+        else:
+            self.send(self.delimited(event))
+
+    def delimited(self, event):
+        log_line = self.header
+        for key, value in event.items():
+            log_line += self.delimiter + key + ':' + str(value)
```
sebix: If the strings contain `|` or `:` it's not parsable anymore

intelmq/bots/outputs/udp/output.py
```diff
@@ -0,0 +1,56 @@
...
+        self.udp = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        self.raw_field = self.parameters.raw_field
+        self.format = self.parameters.format
+
+    def process(self):
+        event = self.receive_message()
+
+        if event is None:
+            self.acknowledge_message()
+            return
```
sebix: When I cleaned up the code in certtools/intelmq#494, this block has been removed.

sebix:
Thanks for this bot, just successfully tested it (with netcat as server)!

Could you also squelch the 9 commits for the readme into one? (Use interactive rebasing with `rebase -i master`). And probably also 45fee27 and 6830636.

intelmq/bots/outputs/udp/output.py
pedromreis: What do you propose?

pedromreis:
I think I didn't get the rebase thing right ... sorry :)

intelmq/bots/outputs/udp/output.py
sebix: And what about newlines in strings? We need to escape all these characters. (And/Or quote them depending on the char)

pedromreis: When and how can that happen? Do you have an example to work on?
Should we keep the chars or simply strip them out? How relevant can the chars be to a log/syslog/alert/etc?

intelmq/bots/BOTS
```diff
@@ -789,6 +789,18 @@
...
+        "UDP": {
+            "description": "UDP is a simple UDP bot responsible to send events to a udp port (e.g.: syslog daemon).",
+            "module": "intelmq.bots.outputs.udp.output",
+            "parameters": {
+                "field_delimiter": "|",
+                "format": "<json/delimited> read README.md",
+                "header": "<header text>",
+                "ip": "<ip>",
+                "port": "<port>",
+                "raw_field": "<keep/drop>"
```
aaronkaplan: how about changing this field to "keep_raw_field": true/false?
This way we can use the built in true and false boolean values of JSON.

intelmq/bots/outputs/udp/output.py
aaronkaplan: Well, you could escape the characters maybe.

intelmq/bots/BOTS
sebix: Yep, I think this would be more intuitive.

sebix:
Okay, until we have a better solution, we could just raise a warning if the delimiter is in the value of some field.

For the implementation of this: Event/Report inherit from dict. You can just iterate over the values with `Event.values()` (not tested, but should work :D )

pedromreis:
Regarding the delimiter field and the case of some event field has the delimiter char(s), thinking ahead, if we want (and I do) output to CEF format, the delimiter field on CEF will always be the pipe '|'. So, maybe I can define alternate chars to substitute in the event field if it is present? What do you ppl think? Example: define delimiter field = ':', alternate chars '::'. If one event field value has the string, for example, "username:password" it would be changed to "username::password" in the output. It has the obvious drawback of changing the original event...

aaronkaplan:
On Tue, May 17, 2016 at 03:47:00PM -0700, Pedro Miguel Reis wrote:
> Regarding the delimiter field and the case of some event field has the delimiter char(s), thinking ahead, if we want (and I do) output to CEF format, the delimiter field on CEF will always be the pipe '|'. So, maybe I can define alternate chars to substitute in the event field if it is present? What do you ppl think? Example: define delimiter field = ':', alternate chars '::'. If one event field value has the string, for example, "username:password" it would be changed to "username::password" in the output. It has the obvious drawback of changing the original event... 
> 


how about escape characters?


> ---
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/certtools/intelmq/pull/503#issuecomment-219876938

--

pedromreis:
> how about escape characters?
Already striped control chars, what other escape characters are you thinking about? Better yet, what chars should we allow?

aaronkaplan:
On Tue, May 17, 2016 at 03:54:37PM -0700, Pedro Miguel Reis wrote:
> > how about escape characters?
> Already striped control chars, what other escape characters are you thinking about? Better yet, what chars should we allow?
> 

well you could define an escape character to escape the separator....

Best,
a.

aaronkaplan:
On Tue, May 17, 2016 at 03:54:37PM -0700, Pedro Miguel Reis wrote:
> > how about escape characters?
> Already striped control chars, what other escape characters are you thinking about? Better yet, what chars should we allow?


... or maybe I am already to tired to think ;-)

pedromreis:
> well you could define an escape character to escape the separator....

And who would escape that? :)
Lets think about this.. how normally other application do logging to syslog?
Maybe someone that has more experience then me can give us some insights.

sebix:
I think we can merge this and eventually solve the problem with the delimiter later. Our TCP Output Bot can only send JSON, so this one can even do more. I think this is fine.
ping @aaronkaplan 

I'd do some small fixes after merging it.

intelmq/bots/BOTS
```diff
@@ -826,6 +826,18 @@
                 "port": "<port>"
             }
         },
+        "UDP": {
+            "description": "UDP is a simple UDP bot responsible to send events to a udp port (e.g.: syslog daemon).",
+            "module": "intelmq.bots.outputs.udp.output",
+            "parameters": {
+                "field_delimiter": "|",
```
aaronkaplan: here you could add a key: "escape_character": "\"

intelmq/bots/BOTS
```diff
@@ -826,6 +826,18 @@
...
+                "format": "<json/delimited> read README.md",
```
aaronkaplan: s/read README.md/for more explanations about this field, checkout out the README.md"

intelmq/bots/BOTS
```diff
@@ -826,6 +826,18 @@
...
+                "header": "<header text>",
+                "ip": "<ip>",
```
aaronkaplan: not implicitly clear what this is. Is this the ip address of the UDP desitnation server?
If so, I suggest nameing it ``destination_server`` or ``udp_server`` or similar. Does it need to be an IP address? Can't it be a hostname?

I am a big fan of naming this so that they are very clear what they imply. even without reading the README.md

intelmq/bots/BOTS
```diff
@@ -826,6 +826,18 @@
...
+                "keep_raw_field": "<true/false>",
+                "port": "<port>"
```
aaronkaplan: put next to ``ip``

intelmq/bots/outputs/udp/README.md
```diff
@@ -0,0 +1,46 @@
...
+* format            : json
+* Header            : header example
+* keep_raw_field    : true
+* ip                : 127.0.0.1
+* port              : 514
+
+**Line in syslog:**
+
+`
+Apr 29 11:01:29 header example {"raw": "MjAxNi8wNC8yNV8xMTozOSxzY2hpenppbm8ub21hcmF0aG9uLmNvbS9na0NDSnVUSE0vRFBlQ1pFay9XdFZOSERLbC1tWFllRk5Iai8sODUuMjUuMTYwLjExNCxzdGF0aWMtaXAtODUtMjUtMTYwLTExNC5pbmFkZHIuaXAtcG9vbC5jb20uLEFuZ2xlciBFSywtLDg5NzI=", "source": {"asn": 8972, "ip": "85.25.160.114", "url": "http://schizzino.omarathon.com/gkCCJuTHM/DPeCZEk/WtVNHDKl-mXYeFNHj/", "reverse_dns": "static-ip-85-25-160-114.inaddr.ip-pool.com"}, "classification": {"type": "malware"}, "event_description": {"text": "Angler EK"}, "feed": {"url": "http://www.malwaredomainlist.com/updatescsv.php", "name": "Malware Domain List", "accuracy": 100.0}, "time": {"observation": "2016-04-29T10:59:34+00:00", "source": "2016-04-25T11:39:00+00:00"}}
```
aaronkaplan: Is there a limit of the length of a line on the receiving server?

intelmq/bots/outputs/udp/output.py
```diff
@@ -0,0 +1,56 @@
...
+import intelmq.lib.utils as utils
+from intelmq.lib.bot import Bot
+
+
+class UDPBot(Bot):
+
+    def init(self):
+        self.delimiter = self.parameters.field_delimiter
+        self.header = self.parameters.header
+        self.udp_host = self.parameters.ip
```
aaronkaplan: here it is called ``udp_host``. Sounds good to also use that as a parameter in the config
Might want to do a gethostbyname() call on the name, if a name is given and not an IP address.

intelmq/bots/outputs/udp/output.py
aaronkaplan: another idea is to use a CSV output module for producing correct CSV output which does escape things the right way. Actually, when thinking about it, I prefer that approach.

sebix: I'd implement csv formatting for udp, tcp and possibly other outputs later. But good idea.

aaronkaplan: So... if we are in agreement that this comes later (let's keep this issue open under the release 1.1 tag?) - we can merge (from my side).

sebix: CSV could solve this problem, but it's complex to configure. A few question that popped up in my head: How so sort? How to give the headers? Is a list of header fields given by the configuration? etc.

sebix:
Rebased, fixed and merged.

sebix:
See d6f8c2ed1d18e56f3c5987c5275ee92e24c39cce

PR #508, opened 2016-05-10
Bot class: renamings, reloading on sighup, shutdown method

sebix:
* Bots are re-initialized on SIGHUP now :)
* new method `shutdown` is available for bots for a graceful shutdown
* internal names are "private" now (leading __)
* updated docs

### DIFF
--- a/docs/Developers-Guide.md
+++ b/docs/Developers-Guide.md
@@ -227,11 +227,12 @@ In the `intelmq/lib/` directory you can find some libraries:
 
 There's a dummy bot including tests at `intelmq/tests/bots/test_dummy_bot.py`.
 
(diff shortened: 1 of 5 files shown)


sebix:
Oh, an the sighup-reloading also initializes the logger again, so this can be used when the logfile is rotated!

aaronkaplan:
On Wed, May 11, 2016 at 01:40:56AM -0700, Sebastian wrote:
> Oh, an the sighup-reloading also initializes the logger again, so this can be used when the logfile is rotated!
> 

Excellent, thank you!

> ---
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/certtools/intelmq/pull/508#issuecomment-218397508

--

aaronkaplan:
looks good. I did not check if every renaming of a function/ private class member to ``__oldname`` was really made of if something was forgotten. How can we test that everything was done?

PR #509, opened 2016-05-10
Filecollector

dmth:
This PR contains code which adds a collector that is capable of reading data from the filesystem.

This can be used for custom data, or for testing of IntelMQ

**Attention**: This code also alters the Generic-CSV-Parser expert a little bit.
This alteration was necessary to enable the configuration of the Parser from IntelMQ-Manager.


aaronkaplan:
why did the checks fail? @dmth: do you actually test things in a live system before you make the pull request? The travis-ci check says that the BOTS file has an incorrect syntax. I did not verify this but basically  a unit test failed.

sebix:
@aaronkaplan The unittest also checks for sorting (beside parsability and indentation).

dmth:
aaronkaplan: Yes we test things, No we don't test things in a productive system.

The test fails because two parameters in BOTS json are not in alphabetic order.

intelmq/bots/collectors/file/collector_file.py
```diff
@@ -0,0 +1,75 @@
...
+class FileCollectorBot(Bot):
+
+    def init(self):
+        # Test if path can be accessed
+        try:
+            if not os.path.isdir(self.parameters.path):
+                self.logger.error("The path does not lead to " \
+                    "a directory")
+
+        except:
```
sebix: Two issues here: first, we don't want to catch all Exceptions, only `PermissionError`
Second, `os.path.isdir` does not raise for permission errors here (for me at least). `os.stat` does.

intelmq/bots/collectors/file/collector_file.py
```diff
@@ -0,0 +1,75 @@
...
+                    report = Report()
+                    report.add("raw", f.read())
+                    report.add("feed.name", self.parameters.feed)
+                    report.add("feed.url", "file:/%s" %filename)
+                    report.add("feed.accuracy", self.parameters.accuracy)
+                    self.send_message(report)
+
+                    f.close()
+
+                    if self.parameters.delete_file == True:
```
sebix: Why `== True`? Just use `if self.parameters.delete_file:`

(`==` calls `__eq__` on the object, while e.g. `is` compares object identities)

dmth: `== True` was necessary as  `if self.parameters.delete_file`` always evaluated to `true` in my testsetup

intelmq/bots/collectors/file/collector_file.py
```diff
@@ -0,0 +1,75 @@
...
+            p = os.path.abspath(self.parameters.path)
+
+            #iterate over all files in dir
+            for f in os.listdir(p):
+                if fnmatch.fnmatch(f, '*'+self.parameters.extension):
+                    filename = p + '/' + f
+
+                    self.logger.debug("Found file to process: %s" %filename)
+
+                    f = open(filename, 'r')
```
sebix: Please always use context managers when opening files

dmth:
Thank you for reviewing.Unfortunately I will not find the time to integrate your suggestion in this week or the next. So this PR will stay open for a bit longer, unless someone else has a look at these things.

intelmq/bots/collectors/file/collector_file.py
sebix: If false is given as as string, this evaluates to true, as it's a non-empty string. But using `true` itself, this works expected

```python
>>> bool(json.loads("[false]")[0])
False
>>> bool(json.loads("[\"false\"]")[0])
Out[8]: True
```

intelmq/bots/collectors/file/collector_file.py
```diff
@@ -0,0 +1,83 @@
...
+        self.logger.debug("Started looking for Files")
+
+        if os.path.isdir(self.parameters.path):
+            p = os.path.abspath(self.parameters.path)
+
+            #iterate over all files in dir
+            for f in os.listdir(p):
+                filename = os.path.join(p, f)
+                if os.path.isfile(filename):
+                    if fnmatch.fnmatch(f, '*'+self.parameters.extension):
```
sebix: Either you add a dot here or you add it in the parameter. Otherwise it's a postfix, not an extension.

dmth: Best way would be the parameter. Adding a dot in this line would make matching on all files in this directory impossible

aaronkaplan:
what about this PR ? Was this reviewed enough?

sebix:
Filename matching is still open.

So yes, I can merge this manually with a minor fix for filename matching.

sebix:
The fixes were not so minor. Final diff: https://github.com/certtools/intelmq/commit/2bd6a1e3457727a3c19be9c5d4f7c386cca5626f

PR #519, opened 2016-05-12
Update mail collector bots (URL and attachment)

gsiv:
These patches contain the following notable changes to the two mail
collector bots:

  - URL collector bot: retrieve from URLs with HTTP authentication
  - only mark emails as read if they match the bots' configured
    regexs
  - update parameters to match existing bot profiles

Please note that we need to use Imbox 0.8.  Previous versions run
into errors in case of multi-part messages in the mailbox.


intelmq/bots/BOTS
```diff
@@ -307,6 +307,41 @@
                 "rate_limit": 10800
             }
         },
+        "Generic Mail Attachment Fetcher": {
+            "description": "Monitor IMAP mailboxes and retrieve mail attachments",
+            "module": "intelmq.bots.collectors.mail.collector_mail_attach",
+            "parameters": {
+                "accuracy": "",
```
sebix: While this is a string, it's default is defined in `defaults.conf` and can be overwritten on per-bot basis. You are overriding this default with `None`. Just remove this line for both bots.

sebix:
Rest of the code looks fine. Thanks for your work!

PR #523, opened 2016-05-17
Make URL mail collector bot more robust

gsiv:
These changes attempt to make the mail URL collector bot more robust
and user friendly.

### DIFF
--- a/intelmq/bots/collectors/mail/collector_mail_url.py
+++ b/intelmq/bots/collectors/mail/collector_mail_url.py
@@ -41,6 +41,9 @@ class MailURLCollectorBot(Bot):
                     match = re.search(self.parameters.url_regex, str(body))
                     if match:
                         url = match.group()
+                        url = url.strip()     # strip leading and trailing spaces
+                        url = url.strip('\n') # strip newlines
+                        url = url.strip('\r') # strip carriage returns
 
                         # Build request
                         self.http_header = getattr(self.parameters,
--- a/intelmq/bots/collectors/mail/collector_mail_url.py
+++ b/intelmq/bots/collectors/mail/collector_mail_url.py
@@ -38,7 +38,7 @@ class MailURLCollectorBot(Bot):
                 self.logger.info("Reading email report")
 
                 for body in message.body['plain']:
-                    match = re.search(self.parameters.url_regex, body)
+                    match = re.search(self.parameters.url_regex, str(body))
                     if match:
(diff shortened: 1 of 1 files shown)


intelmq/bots/collectors/mail/collector_mail_url.py
```diff
@@ -38,9 +38,12 @@ def process(self):
                 self.logger.info("Reading email report")
 
                 for body in message.body['plain']:
-                    match = re.search(self.parameters.url_regex, body)
+                    match = re.search(self.parameters.url_regex, str(body))
                     if match:
                         url = match.group()
+                        url = url.strip()     # strip leading and trailing spaces
+                        url = url.strip('\n') # strip newlines
```
sebix: unnecessary:

```python
>>> "test  \n\r   ".strip()
'test'
```

intelmq/bots/collectors/mail/collector_mail_url.py
```diff
@@ -38,9 +38,12 @@ def process(self):
                 self.logger.info("Reading email report")
 
                 for body in message.body['plain']:
-                    match = re.search(self.parameters.url_regex, body)
+                    match = re.search(self.parameters.url_regex, str(body))
```
sebix: Why should this not be a string? Could this be a bytes in a special case? Could you describe/show such a case?

gsiv: On Tue, May 24, 2016 at 02:24:19AM -0700, Sebastian wrote:
> Why should this not be a string? Could this be a bytes in a special case? Could you describe/show such a case?

All I can say right now is that I observed this behavior while
testing the bot with a real-world message.  This is the traceback:

  2016-05-27 12:09:13,119 - cymru - ERROR - Bot has found a problem.
    File "/usr/local/lib/python3.4/dist-packages/intelmq-1.0.0.dev4-py3.4.egg/intelmq/lib/bot.py", line 102, in start
      self.process()
    File "/usr/local/lib/python3.4/dist-packages/intelmq-1.0.0.dev4-py3.4.egg/intelmq/bots/collectors/mail/collector_mail_url.py", line 41, in process
      match = re.search(self.parameters.url_regex, body)
    File "/usr/lib/python3.4/re.py", line 170, in search
      return _compile(pattern, flags).search(string)
  TypeError: can't use a string pattern on a bytes-like object

Issue #513, opened 2016-05-11
Send multiple messages at once


ondj:
Some collectors Report contain multiple Event, so parser in one `process` call receive one and send multiple messages. But this is vulnerable, because for example in this simple implementation is source IP address invalid or some other exception is thrown and `error_procedure` is set to `pass` (which is a default behavior), events processed before the invalid event are duplicated because bot tries to parse Report multiple times.

```python
def process(self):
    report = self.receive_message()
    for parsed in parse(report):
    	event = Event(report)
    	event.add("source.ip", parsed.source_ip)
    	self.send_message(event)
    self.acknowledge_message()
```

It can by solved by:
* adding ability to send_message to send multiple at once. Then, the usage look like:
```python
def process(self):
    report = self.receive_message()
    events = []
    for parsed in parse(report):
    	event = Event(report)
    	event.add("source.ip", parsed.source_ip)
    	events.push(event)
    self.send_message(events)
    self.acknowledge_message()
```
* or `send_message` will only store events in memory and are send at once after `acknowledge_message` call (but this is not suitable for collectors).
* or (my preffered variant) add new `send_message_and_acknowledge` which can take one message or multiple messages.

Little advantage is, that sending multiple messages is faster because `lpush` Redis method supports multiple arguments.

sebix:
Yes, this is a simple fix. But tough I would prefer this:

```python
self.send_message(*events)
```
As we do not need to check for the argument's type in the function.

But I'm working on a new architecture for parsers, which should solve this issue too.

PR #529, opened 2016-05-24
New parser's architecture, dumping of individual messages

sebix:
An additional possibility for parsers.

The parsing of the text itself (splitlines in the simplest case) has been separated from the parsing of individual lines. Now, single lines can be dumped.
As small drawback, we need now mechanisms to reconstruct the original messages, but this is usually a one-liner (see examples).

This also lowers the complexity.

All changes are backwards-compatible, parsers using `process()` work fine, nothing changed for them.

This is a proof-of-concept, ready for reviews.

Still missing: docs for devs, tests, more conversions

refers to #374, #513 and #405
ping at @robcza @SYNchroACK @aaronkaplan @ondj @mauroasilva (everybody how participated in linked discussions)

### DIFF
--- a/intelmq/bots/parsers/malwaredomainlist/parser.py
+++ b/intelmq/bots/parsers/malwaredomainlist/parser.py
@@ -35,7 +35,7 @@ class MalwareDomainListParserBot(ParserBot):
         event.add("source.asn", int(row[6]))
 
         event.add('classification.type', 'malware')
-        event.add("raw", ",".join(row))
+        event.add("raw", self.recoverline(row))
         yield event
 
 if __name__ == "__main__":
--- a/intelmq/bots/parsers/malwaredomainlist/parser.py
+++ b/intelmq/bots/parsers/malwaredomainlist/parser.py
@@ -16,9 +16,9 @@ class MalwareDomainListParserBot(ParserBot):
             return 'http://' + url
 
     parse = ParserBot.parse_csv
-    recoverline = ParserBot.recoverline_csv
+    recover_line = ParserBot.recover_line_csv
 
-    def parseline(self, row, report):
+    def parse_line(self, row, report):
         event = Event(report)
 
         event.add("time.source", row[0].replace('_', ' ') + " UTC")
@@ -35,7 +35,7 @@ class MalwareDomainListParserBot(ParserBot):
         event.add("source.asn", int(row[6]))
 
         event.add('classification.type', 'malware')
-        event.add("raw", self.recoverline(row))
+        event.add("raw", self.recover_line(row))
         yield event
 
 if __name__ == "__main__":
--- a/intelmq/bots/parsers/malwaredomainlist/parser.py
+++ b/intelmq/bots/parsers/malwaredomainlist/parser.py
@@ -1,16 +1,13 @@
 # -*- coding: utf-8 -*-
-import csv
-import io
 import sys
 from urllib.parse import urlparse
 
-from intelmq.lib import utils
-from intelmq.lib.bot import Bot
+from intelmq.lib.bot import ParserBot
 from intelmq.lib.exceptions import InvalidValue
 from intelmq.lib.message import Event
 
 
-class MalwareDomainListParserBot(Bot):
+class MalwareDomainListParserBot(ParserBot):
     def add_http(self, url):
         """
         We can assume here, that everything is http by default if not given.
@@ -18,33 +15,28 @@ class MalwareDomainListParserBot(Bot):
         if '://' not in url:
             return 'http://' + url
 
-    def process(self):
-        report = self.receive_message()
+    parse = ParserBot.parse_csv
+    recoverline = ParserBot.recoverline_csv
 
-        raw_report = utils.base64_decode(report.get("raw"))
-        for row in csv.reader(io.StringIO(raw_report)):
-            event = Event(report)
-            self.logger.debug(repr(row))
+    def parseline(self, row, report):
+        event = Event(report)
 
-            event.add("time.source", row[0].replace('_', ' ') + " UTC")
-            if row[1] != '-':
-                event.add("source.url", self.add_http(row[1]))
-            try:
-                event.add("source.ip", row[2])
-            except InvalidValue:
-                event.add("source.url", self.add_http(row[2]))
-                event.add('source.ip', urlparse(row[2]).netloc)
-            event.add("source.reverse_dns", row[3])
-            event.add("event_description.text", row[4])
-            # TODO: ignore abuse contact for now
-            event.add("source.asn", int(row[6]))
+        event.add("time.source", row[0].replace('_', ' ') + " UTC")
+        if row[1] != '-':
+            event.add("source.url", self.add_http(row[1]))
+        try:
+            event.add("source.ip", row[2])
+        except InvalidValue:
+            event.add("source.url", self.add_http(row[2]))
+            event.add('source.ip', urlparse(row[2]).netloc)
+        event.add("source.reverse_dns", row[3])
+        event.add("event_description.text", row[4])
+        # TODO: ignore abuse contact for now
+        event.add("source.asn", int(row[6]))
 
-            event.add('classification.type', 'malware')
-            event.add("raw", ",".join(row))
-
-            self.send_message(event)
-        self.acknowledge_message()
-
+        event.add('classification.type', 'malware')
+        event.add("raw", ",".join(row))
+        yield event
 
 if __name__ == "__main__":
     bot = MalwareDomainListParserBot(sys.argv[1])
--- a/intelmq/lib/bot.py
+++ b/intelmq/lib/bot.py
@@ -90,6 +90,37 @@ class Bot(object):
     def shutdown(self):
         pass
 
+    def parse(self, report):
+        """
+        Default parser yielding stripped lines.
+        """
+        for line in utils.base64_decode(report.get("raw")).splitlines():
+            yield line.strip()
+
+    def process(self):
+        self.tempdata = []
+        self.__failed = set()
+        report = self.receive_message()
+
+        for line in self.parse(report):
+            if not line:
+                continue
+            try:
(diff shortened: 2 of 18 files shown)


ondj:
I like your change, but my only proposal is to move `Bot.parse` and `Bot.process` methods to new `ParserBot(Bot)` class, because these methods are only useful for parser and not for collectors, experts and outputs.

aaronkaplan:
okay, will test

sebix:
I worked on further improvments and capsulated the new functionalities in a new ParserBot class as proposed by @ondj 

Feedback and testing appreciated!

Ping @dmth

ondj:
@sebix I think, it is perfect :+1:

dmth:
Will "old" parsers inheriting from bot still work?

sebix:
@dmth Yes, as you can see in the diff, there are only minor,
backwards-compatible changes to the Bot-class.

intelmq/lib/bot.py
```diff
@@ -256,18 +258,19 @@ def __disconnect_pipelines(self):
             self.__destination_pipeline = None
             self.logger.info("Disconnecting from destination pipeline.")
 
-    def send_message(self, message):
-        if not message:
-            self.logger.warning("Sending Message: Empty message found.")
-            return False
+    def send_message(self, *messages):
+        for message in messages:
+            if not message:
```
aaronkaplan: Is that necessary? the for message in messages shall not produce None IMHO.

intelmq/lib/bot.py
```diff
@@ -286,8 +289,10 @@ def acknowledge_message(self):
         self.__last_message = self.__current_message
         self.__source_pipeline.acknowledge()
 
-    def __dump_message(self, error_traceback):
-        if self.__current_message is None:
+    def _dump_message(self, error_traceback, message=None):
+        if message is None:
```
aaronkaplan: if message is None twice?

intelmq/lib/bot.py
```diff
@@ -394,5 +399,56 @@ def __load_harmonization_configuration(self):
...
 
 
+class ParserBot(Bot):
+
+    def parse_csv(self, report):
+        """
+        CSV parser.
+        """
+        raw_report = utils.base64_decode(report.get("raw"))
+        for line in csv.reader(io.StringIO(raw_report)):
```
aaronkaplan: what about headers?

intelmq/lib/bot.py
```diff
@@ -394,5 +399,56 @@ def __load_harmonization_configuration(self):
...
+                    self.send_message(*events)
+
+        if self.__failed:
+            for exc, line in self.__failed:
+                self._dump_message(exc, self.recoverline(line))
+
+        self.acknowledge_message()
+
+    def recoverline(self, line):
+        return '\n'.join(self.tempdata + [line])
```
aaronkaplan: are we always \n separated?

intelmq/lib/bot.py
```diff
@@ -394,5 +399,56 @@ def __load_harmonization_configuration(self):
                             "Key %s is not valid." % _key)
 
 
+class ParserBot(Bot):
+
+    def parse_csv(self, report):
```
aaronkaplan: where does this get used? I like it but ... should we not have a subclass of ParseBot called CSVParseBot for that?

sebix:
Concerning the provided functions `parse`, `parse_csv`, `recoverline`, `recoverline_csv`:
My approach was to provide functions which are simple and work for most cases out-of-the box, like any (non-csv) (delimiter separate) file, on per-line basis. So most bots do not need to adapt these pieces of code.
Other bots do need other parsing, like csv. The bot can now use one of these generic ones. But they can be overridden by custom ones too. So the generic class can provide a set of useful and generic functions, and the Bot can "choose" which one to use, if possible.

The other approach, you mentioned, is to provide different classes for each "type" of parser. I don't know yet which one of the approaches is feasible (complexity, usability for dev, ...). Possibly it's easier to say when more/all of bots are converted.

intelmq/lib/bot.py
sebix: I didn't change this part, but I do think it's useful: It's a situation which SHOULD NOT happen, but can be handled -> warning.

intelmq/lib/bot.py
sebix: For the ParserBot, `self.__current_message` lost it's meaning. Thus I needed to give it directly. But it would be cleaner anyway to remove the fallback to current message at all and use the parameter only. And raise a warning, if None is encountered.

intelmq/lib/bot.py
sebix: Not implemented and used yet. No converted parsers had a csv format with headers.

sebix:
Any open questions @aaronkaplan?

intelmq/bots/parsers/malwaredomainlist/parser.py
```diff
@@ -1,50 +1,42 @@
...
+        except InvalidValue:
+            event.add("source.url", self.add_http(row[2]))
+            event.add('source.ip', urlparse(row[2]).netloc)
+        event.add("source.reverse_dns", row[3])
+        event.add("event_description.text", row[4])
+        # TODO: ignore abuse contact for now
+        event.add("source.asn", int(row[6]))
+
+        event.add('classification.type', 'malware')
+        event.add("raw", ",".join(row))
```
sebix: Could be done by recoverline (especially useful for non-trivial formats, as csv)

aaronkaplan:
On Tue, Jun 14, 2016 at 02:12:56AM -0700, Sebastian wrote:
> Any open questions @aaronkaplan?
> 

nope!
I think we can pull this in.
Best,
a.

SYNchroACK:
@sebix , I would like to drop here some notes:

1. Since we are creating a new sub-class 'ParserBot', I **strongly recommend** to create more 3 sub-classes to each type of bot and make the changes in all bots accordingly to the type.
2. Regarding the methods '*line', I would like to see '*_line'. E.g. 'recoverline' -> 'recover_line'
3. Please, before pull in, make sure that we have documentation for these new features and also update the current documentation.

sebix:
> 1. Since we are creating a new sub-class 'ParserBot', I **strongly recommend** to create more 3 sub-classes to each type of bot and make the changes in all bots accordingly to the type.

This was the most important Bot-type. I do not plan to add other new Bot classes for 1.0.
> 2. Regarding the methods '*line', I would like to see '*_line'. E.g. 'recoverline' -> 'recover_line'

yes, should do this
> 3. Please, before pull in, make sure that we have documentation for these new features and also update the current documentation.

Of course.

intelmq/lib/bot.py
mauroasilva: I agree... if current_message lost it's meaning for the ParserBot, I would just drop it everywhere you can. The code becomes cleaner.

intelmq/lib/bot.py
mauroasilva: We need to be careful in having too many nested subclasses... It could lead to low code readability... but I agree this may not be the best place for this... maybe some kind of util package where we could have a few parse functions that several bots use?

intelmq/lib/bot.py
```diff
@@ -394,5 +399,56 @@ def __load_harmonization_configuration(self):
...
+        self.tempdata = []
+        self.__failed = []
+        report = self.receive_message()
+
+        for line in self.parse(report):
+            if not line:
+                continue
+            try:
+                # filter out None
+                events = list(filter(bool, self.parseline(line, report)))
```
mauroasilva: Shouldn't we have a default parseline? Even if it is just a `return line`, or even one just with `pass` and a comment saying you should override it on your parsers. That way people reading the code don't have to read the documentation to actually know they should override it.

Also... I'm not 100% sure of the readability of this line. But to be honest I don't have a better idea except for creating a loop which would probably become much slower, and definitely bulkier.

intelmq/lib/bot.py
```diff
@@ -394,5 +399,56 @@ def __load_harmonization_configuration(self):
...
+            except Exception as exc:
+                self.logger.exception('Failed to parse line.')
+                self.__failed.append((exc, line))
+            else:
+                if events:
```
mauroasilva: Do we really need this? The `send_message` already has the if, so we're basically just duplicating it.

intelmq/lib/bot.py
```diff
@@ -394,5 +399,56 @@ def __load_harmonization_configuration(self):
...
+
+    def parse(self, report):
+        """
+        Default parser yielding stripped lines.
+        """
+        for line in utils.base64_decode(report.get("raw")).splitlines():
+            yield line.strip()
+
+    def process(self):
+        self.tempdata = []
```
mauroasilva: Maybe this needs a little comment at the front explaining the meaning and use of it?

intelmq/lib/bot.py
sebix: I dropped these lines and now the parameter is used.

intelmq/lib/bot.py
sebix: I created a dummy function raising `NotImplementedError`.

An alternative would be: `[line for line in self.parse_line(line, report) if line]`

intelmq/lib/bot.py
sebix: Thanks, removed. (your rational is not correct tough, `send_message` would just iterate over an empty list).

sebix:
I implemented your suggestions @SYNchroACK and @mauroasilva

intelmq/lib/bot.py
mauroasilva: True... the old send_message was the one doing the if. But iterating over an empty list is still a valid thing to do so the argument still holds.

intelmq/lib/bot.py
mauroasilva: Great. I like the NotImplementedError. It adheres to the convention for methods that should be implemented on subclasses.

Issue #538, opened 2016-06-03
Pipeline failed - AttributeError: 'NoneType' object has no attribute 'decode'


MAHDTech:
Can anyone else confirm this issue:

To re-produce, create a simple bot configuration
1. Malc0de IP Collector
2. Malc0de IP Parser
3. File Output Bot
4. Start BotNet

```
Pipeline failed.
Traceback (most recent call last):
  File "/usr/lib/python3.4/site-packages/intelmq-1.0.0.dev4-py3.4.egg/intelmq/lib/pipeline.py", line 113, in receive
    return utils.decode(self.pipe.lindex(self.internal_queue, -1))
  File "/usr/lib/python3.4/site-packages/intelmq-1.0.0.dev4-py3.4.egg/intelmq/lib/utils.py", line 70, in decode
    return str(text.decode(encoding))
AttributeError: 'NoneType' object has no attribute 'decode'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.4/site-packages/intelmq-1.0.0.dev4-py3.4.egg/intelmq/lib/bot.py", line 102, in start
    self.process()
  File "/usr/lib/python3.4/site-packages/intelmq-1.0.0.dev4-py3.4.egg/intelmq/bots/outputs/file/output.py", line 16, in process
    event = self.receive_message()
  File "/usr/lib/python3.4/site-packages/intelmq-1.0.0.dev4-py3.4.egg/intelmq/lib/bot.py", line 258, in receive_message
    message = self.source_pipeline.receive()
  File "/usr/lib/python3.4/site-packages/intelmq-1.0.0.dev4-py3.4.egg/intelmq/lib/pipeline.py", line 117, in receive
    raise exceptions.PipelineError(exc)
intelmq.lib.exceptions.PipelineError: pipeline failed - AttributeError("'NoneType' object has no attribute 'decode'",)
```

MAHDTech:
This is looking for like a redis configuration issue, can any confirm whats a quick test to ensure I have it working correctly?

MAHDTech:
never mind, this one was a missing SELinux configuration.

issue cloed.

MAHDTech:
I thought it was SELinux, but even on a fresh VM with SELinux set in permissive mode I can get the same error occasionally, and when it occurs the output file has very few entries, it just bombs out and doesn't pass as expected if issue.

`2016-06-03 06:11:48,215 - Output-File - DEBUG - Receiving Message.
[...]

Issue #514, opened 2016-05-11
intelmqctl: error message for list bots


aaronkaplan:
Hi please take a look. Something broke.

```
$ intelmqctl list bots
(...)
Description: Turris Greylist Parser is the bot responsible to parse the report and sanitize the information.
Traceback (most recent call last):
  File "/usr/local/bin/intelmqctl", line 9, in <module>
    load_entry_point('intelmq==1.0.0.dev4', 'console_scripts', 'intelmqctl')()
  File "/usr/local/lib/python3.4/dist-packages/intelmq/bin/intelmqctl.py", line 526, in main
    x.run()
  File "/usr/local/lib/python3.4/dist-packages/intelmq/bin/intelmqctl.py", line 269, in run
    results = call_method()
  File "/usr/local/lib/python3.4/dist-packages/intelmq/bin/intelmqctl.py", line 399, in list_bots
    "".format(bot_id, self.startup[bot_id]['description']))
KeyError: 'description'

sebix:
One of your configured bots does not have a description (see `startup.conf`). But the tool should not fail.

Issue #501, opened 2016-04-28
Modify bot only accepts string fields as parameter?


pedromreis:
Or I'm I making something wrong...
If I have a `modify.conf` like:

```python
{
"default": {
    "set_target_tag": [{
            "source.asn": "65500"
        }, {
            "event_description.target": "My company"
        }]
    }
}
```

I get:

```python
Traceback (most recent call last):
  File "/usr/lib/python3.4/site-packages/intelmq-1.0.0.dev4-py3.4.egg/intelmq/lib/bot.py", line 102, in start
    self.process()
  File "/usr/lib/python3.4/site-packages/intelmq-1.0.0.dev4-py3.4.egg/intelmq/bots/experts/modify/expert.py", line 55, in process
    if matches(event, default_cond, rule_cond):
  File "/usr/lib/python3.4/site-packages/intelmq-1.0.0.dev4-py3.4.egg/intelmq/bots/experts/modify/expert.py", line 26, in matches
    if not re.search(rule, event[name]):
  File "/usr/lib64/python3.4/re.py", line 170, in search
    return _compile(pattern, flags).search(string)
TypeError: expected string or buffer
```

sebix:
Thanks for the report. Pushed a fix to master. Make sure, the types in the event and the rules do match (in your example, asn is a string).

pedromreis:
Hi.
So, how can I get a regex on a int field?
If I want to match all ^64[0-9][0-9][0-9] how should I do?
Using the above does not work if I put between "" or out.

sebix:
Okay, so I propose:
* If the types do match, and the types are not string, do a `==` comparison.
* If the types do not match
  * if rule is string and data is int or float, convert data to string and do a `re.search`
  * warn otherwise

pedromreis:
So we can match, for example:

"source.port": 443 
or
"source.port": "80|443"

Is that it? Like it :)

sebix:
You can now use this:
```python
"Standard Protocols": {
    "http": [{
            "source.port": "^(80|443)$"
        }, {
            "protocol.application": "http"
        }]
    }
```

PR #542, opened 2016-06-14
few small bugfixes

sebix:
* bugfix for #514 
* bugfix for #501
* bugfix for #538
* version internally and externally available
* new ctl param `--quiet`

Will be merged in a few days if there are no annotations or objections


sebix:
* Invalidate IP address `0.0.0.0`, as it appears in shadowserver data as invalid dummy data, see Intevation/intelmq@1922abbb6a
* bugfix for #543

PR #546, opened 2016-06-15
Update Bots.md: Modify description.

bernhardreiter:
From the description the link to the python doc should be towards the beginning of the string format syntax, not the mini-language section.
(I was searching the '{' description for a while there until I've noticed I had to scroll up to find it)


sebix:
Thanks!

PR #548, opened 2016-06-22
New Bot for Shadowserver feeds

dmth:
This PR introduces a new strategy to parse Shadowserver feeds.

**Known ToDos:**
 * Add Conversion Functions for some Data, as @sebix already suggested
  * IPMI: `*_auth`, `usernames`, `nulluser` and `anon_login` replace  yes/no with true/false
  * IPMI: `default_kg` replace the - or default  with None
  * Hostnames can contain IP's which has to be handled.
 * Add this Bot to the documentation or update the documentation.

PR #549, opened 2016-06-28
Universal shadowserver parser

dmth:
This PR introduces a new strategy to parse Shadowserver feeds.

It's a tidied version of https://github.com/certtools/intelmq/pull/548

**Known ToDos:**
 * Add Conversion Functions for some Data, as @sebix already suggested
  * IPMI: `*_auth`, `usernames`, `nulluser` and `anon_login` replace  yes/no with true/false
  * IPMI: `default_kg` replace the - or default  with None
  * Hostnames can contain IP's which has to be handled.
 * Add this Bot to the documentation or update the documentation.


sebix:
Current behavior for extra:

    'extra': '{"sic": 654321, "response_size": 116, "naics": 123456}',
new one:

    'extra': '{"shadowserver": {"naics": "123456", "sic": "654321", "size": ' '"116", "tag": "chargen"}}',

If we switch to this new behavior, we also should change the documentation about this field.

dmth:
> If we switch to this new behavior, we also should change the documentation about this field.

Correct. Which documentation do you refer to?

intelmq/bots/parsers/shadowserver/parser.py
```diff
@@ -0,0 +1,174 @@
...
+            # fields that coul not be added to the standard intelmq fields
+            # the parser is going to write this information into an object
+            # one level below the "extra root"
+            # e.g.: extra {'shadowserver': {'cc_dns': '127.0.0.1'}
+
+            # set classification.type
+            # TODO this might get dynamic in some feeds.
+            # How to handle that?
+
+            if not conf.get('classification_type'):
```
sebix: Why is this not in init?

intelmq/bots/parsers/shadowserver/parser.py
```diff
@@ -0,0 +1,174 @@
...
+        raw_report = utils.base64_decode(report["raw"])
+        csvr = csv.DictReader(io.StringIO(raw_report))
+
+        # create an array of fieldnames,
+        # those were automagically created by the dictreader
+        allfields = csvr.fieldnames
+
+        # Set a switch if the parser shall reset the feed.name,
+        # code and feedurl for this event
+        self.override = False
```
sebix: Why is this not in init?

dmth: Putting this to init should not have side effects.

intelmq/bots/parsers/shadowserver/parser.py
```diff
@@ -0,0 +1,174 @@
...
+            if not conf.get('classification_type'):
+                self.logger.warn("The classification type for "
+                    + self.parameters.feedname
+                    + " could not be determined. Check ParserConfig!")
+            else:
+                event.add('classification.type', conf.get('classification_type'))
+
+            # set feed.name and url, honor the override parameter
+
+            if 'feed.url' in event and self.override:
```
dmth: I doubt that the if-clause is correct in this place. It should be:
If "field is present and override is set **else if** field is not present"

this should also be corrected for the next two parameters.

sebix: Already catched that one, otherwise a KeyExists would be thrown.

aaronkaplan:
Can we close this PR? I believe the discussion moved to PR #555 ...

sebix:
I'm not keen to close it before the discussion ended. #549 will be closed automatically if #555 is merged.

aaronkaplan:
All right. We'll keep it open then. 

---
Mobile

PR #550, opened 2016-06-29
ENH: omit headers while csv parsing

robcza:
new parameter for generic csv parser to skip the first row with csv headers

### DIFF
--- a/intelmq/bots/BOTS
+++ b/intelmq/bots/BOTS
@@ -1036,6 +1036,7 @@
                 ],
                 "default_url_protocol": "http://",
                 "delimiter": ",",
+                "skip_header": true,
                 "type": "c&c"
             }
         },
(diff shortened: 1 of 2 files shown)


sebix:
Do you want to add the parameter to the template in `BOTS`?

intelmq/bots/BOTS
```diff
@@ -1034,6 +1034,7 @@
                     "",
                     "source.fqdn"
                 ],
+                "headers": true,
```
aaronkaplan: I'd prefer if the syntax was very clear what this does.
How about renameing to "skip_header": true  ?

Otherwise, I'll merge that in of course.
Nice addition, thanks!

robcza: Agree, will do and will merge the ugly commits.

aaronkaplan:
thanks!!

PR #555, opened 2016-07-06
Shadowserver Parsers

sebix:
Based on #549, including bugfixes, more documentation, more features and tests.

I verified them only with 5 feeds, see tests.


sebix:
For the changes relative to #549 see https://github.com/sebix/intelmq/compare/shadowserver-intevation...sebix:shadowserver?expand=1

intelmq/bots/parsers/shadowserver/config.py
```diff
@@ -85,10 +110,10 @@ def validate_ip(value):
         ('source.geolocation.cc', 'geo'),
         ('source.geolocation.region', 'region'),
         ('source.geolocation.city', 'city'),
-        # Other known fields  which will go into "extra"
+        # Other known fields which will go into "extra"
+        ('naics', int),
```
dmth: If NAICS is an empty string this will create trouble.

```
>>> int("")
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: ''
```

intelmq/bots/parsers/shadowserver/README.md
```diff
@@ -61,4 +61,11 @@ Every bot-type is defined by a dictionary with three values:
...
 - `additional_fields`: A dictionary with a static mapping of field name to
-  data, e.g. to set classifications or protocols.
\ No newline at end of file
+  data, e.g. to set classifications or protocols.
+
+The tuples can be of following format:
+
+- `('intelmqkey', 'shadowkey')`
+- `('intelmqkey', 'shadowkey', conversion_function)`
+- `('shadowkey', conversion_function)`, data will be added to extra in this case
```
dmth: I have the feeling, that this behaviour is not clear enough.
What about an `extra_fields` dict?

intelmq/bots/parsers/shadowserver/config.py
```diff
@@ -592,8 +671,9 @@ def validate_ip(value):
         ('source.geolocation.region', 'region'),
         ('source.geolocation.city', 'city'),
     ],
-    'classification_type': 'vulnerable service',
-    'feed_url': 'https://www.shadowserver.org/wiki/pmwiki.php/Services/Open-Memcached'
```
dmth: Why was this removed from configuration?

intelmq/bots/parsers/shadowserver/config.py
sebix: True. Is a NAICS or SIC of `0` valid? @aaronkaplan

intelmq/bots/parsers/shadowserver/config.py
sebix: * Because `feed.url` defines the source of the information, if downloaded. Not the documentation/specification.
* This field is added by the collector, not by a parser.

intelmq/bots/parsers/shadowserver/config.py
aaronkaplan: > On 06 Jul 2016, at 17:44, Sebastian <notifications@github.com> wrote:
> 
> In intelmq/bots/parsers/shadowserver/config.py:
> 
> > @@ -85,10 +110,10 @@ def validate_ip(value):
> >          ('source.geolocation.cc', 'geo'),
> >          ('source.geolocation.region', 'region'),
> >          ('source.geolocation.city', 'city'),
> > -        # Other known fields  which will go into "extra"
> > +        # Other known fields which will go into "extra"
> > +        ('naics', int),
> 
> True. Is a NAICS or SIC of 0 valid? @aaronkaplan

Nope, I would not assume so.

http://www.census.gov/cgi-bin/sssd/naics/naicsrch?chart=2012
has the NAICS codes as XLS files and no, the value 0 is not defined there.

So, actually the converter must be able to either convert to an int or to the empty value (or ignore the field if it is empty in the first place).

a.

intelmq/bots/parsers/shadowserver/README.md
sebix: An `extra_fields` dict may be confusing as all fields are added to extra, not only those mentioned there.

What about this:

    `('extra.', 'shadowkey', conversion_function)`

With the trailing dot it is clearly not a valid key, and thus more intuitive, that the field is added to extra using the conversion function.

intelmq/bots/BOTS
```diff
@@ -1112,30 +1112,14 @@
...
-        },
-        "ShadowServer SNMP": {
-            "description": "ShadowServer SNMP Parser is the bot responsible to parse the report and sanitize the information.",
-            "module": "intelmq.bots.parsers.shadowserver.snmp_parser",
-            "parameters": {}
+        "ShadowServer": {
+            "description": "ShadowServer Parser is a bot capable of parsing all shadowserver feeds, depending on configuration files. Parameter 'feedname' is used as identifier to chose the correct mapping.",
+            "module": "intelmq.bots.parsers.shadowserver.parser",
+            "parameters": {
+                "feedcode": "",
```
sebix: Why do we have this parameter at all @dmth ?

dmth: IMHO feedcode is used to set an alternative name for a feed, for instance in order to obfuscate the real feed name.

From my understanding this makes most sense in a **Collector**, but the shadowserver parse has the ability to override this seeting.

The parameter is not important for the parser

aaronkaplan: ---
Mobile

> On 07.07.2016, at 08:44, Dustin Demuth <notifications@github.com> wrote:
> 
> In intelmq/bots/BOTS:
> 
> > -            "parameters": {}
> > -        },
> > -        "ShadowServer QOTD": {
> > -            "description": "ShadowServer QOTD Parser is the bot responsible to parse the report and sanitize the information.",
> > -            "module": "intelmq.bots.parsers.shadowserver.qotd_parser",
> > -            "parameters": {}
> > -        },
> > -        "ShadowServer SNMP": {
> > -            "description": "ShadowServer SNMP Parser is the bot responsible to parse the report and sanitize the information.",
> > -            "module": "intelmq.bots.parsers.shadowserver.snmp_parser",
> > -            "parameters": {}
> > +        "ShadowServer": {
> > +            "description": "ShadowServer Parser is a bot capable of parsing all shadowserver feeds, depending on configuration files. Parameter 'feedname' is used as identifier to chose the correct mapping.",
> > +            "module": "intelmq.bots.parsers.shadowserver.parser",
> > +            "parameters": {
> > +                "feedcode": "",
> IMHO feedcode is used to set an alternative name for a feed, for instance in order to obfuscate the real feed name.
> 
> From my understanding this makes most sense in a Collector, but the shadowserver parse has the ability to override this seeting.
> 
> The parameter is not important for the parser
> 

Agreed. Should we take it out from the parser?

dmth: I've no objections.

sebix: Removed. If you need to override feed.code (set by collector anyway)
afterwards, use the modify-expert.

aaronkaplan:
okay, I will test this in production now finally and if it's ok merge...

intelmq/bots/parsers/shadowserver/README.md
```diff
@@ -0,0 +1,69 @@
...
+- `optional_fields`: Same format as above, but does not raise errors of field
+  does not exist. If there's no mapping to an intelmq field, you can give a
+  tuple of showserver key and conversion function or not mention it at all.
+  In both cases, the data will be added to extra.
+- `additional_fields`: A dictionary with a static mapping of field name to
+  data, e.g. to set classifications or protocols.
+
+The tuples can be of following format:
+
+- `('intelmqkey', 'shadowkey')`
```
aaronkaplan: unclear description. What is "shadowkey" ? is that literal verbatim text? Or is it an meta-description and the user needs to type in something in place of "shadowkey"? Same for "intelmqkey" (although that looks like it needs to be replaced).

An example: 
  section would be very valuable here.

intelmq/bots/parsers/shadowserver/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Add new Feedformats:
+Add a new feedformat and conversions if required to the file
+`config.py`. Don't forget to update the `feed_idx` dict.
+It is required to look up the correct configuration.
+
+### Configuration
+
+Every bot-type is defined by a dictionary with three values:
+- `required_fields`: A list of tuples containing intelmq's field name, field
```
aaronkaplan: unclear language

intelmq/bots/parsers/shadowserver/README.md
```diff
@@ -0,0 +1,69 @@
...
+  name from data and an optional conversion function. Errors are raisen, when
+  field does not exists in data.
+- `optional_fields`: Same format as above, but does not raise errors of field
```
aaronkaplan: unclear language

intelmq/bots/parsers/shadowserver/README.md
```diff
@@ -0,0 +1,69 @@
...
+  does not exist. If there's no mapping to an intelmq field, you can give a
+  tuple of showserver key and conversion function or not mention it at all.
+  In both cases, the data will be added to extra.
+- `additional_fields`: A dictionary with a static mapping of field name to
```
aaronkaplan: unclear language

bernhard-herzog:
We would like to switch from our own shadowserver parser
(https://github.com/certtools/intelmq/pull/549) to this one, so I've
looked at the differences between this pull request and ours,
concentrating on the four feeds we currently use in intelmq-mailgen
(Botnet-Drone-Hadoop, Sinkhole-HTTP-Drone, Open-Portmapper, Open-SNMP).
I've found two relevant differences:

### No intermediate shadowserver dictionary in extra

The parser of this PR puts the fields that do not correspond to IntelMQ
fields directly into the extra dictionary whereas our version has them
all in an sub-dictionary which is contained in extra under the key
"shadowserver". E.g. with this PR the value of extra might be

```json
{"naics": "0", "sic": "0", "sysdesc": "foo"}
```

with our code this would be

```json
{"shadowserver": {"naics": "0", "sic": "0", "sysdesc": "foo"}}
```

The sub-dictionary reduces the chance of collisions with other
information stored in extra, such as a the notification information the
certbund_contact bot adds to events.

### Botnet-Drone-Hadoop: different IntelMQ fields for 'cc_dns'

The parser for the Botnet-Drone-Hadoop in this PR stores the value of
'cc_dns' in 'destination.reverse_dns', whereas ours puts it into
'destination.fqdn'. The description of 'cc_dns' in the shadowserver
documentation says "Fully qualified domain name of the C&C". This sounds
more like 'destination.fqdn' than 'destination.reverse_dns' to me.

sebix:
> Botnet-Drone-Hadoop: different IntelMQ fields for 'cc_dns'

Yes, thanks. Data looks more like hostnames than reverse names. Also, I need to remove IPs, as they can appear in the data here.

sebix:
> No intermediate shadowserver dictionary in extra

After a short discussion with @dmth we remove this layer. data in extra is additional data which does not fit in the currently configured harmonization. This does not mean, that we don't know the meaning, we haven't validated it etc. The source is also clear: the feed itself (feed.name).

> The sub-dictionary reduces the chance of collisions with other
information stored in extra, such as a the notification information the
certbund_contact bot adds to events.

Why don't you create an extra field in your harmonization? You'll never have collisions and it's much easier to work with. We do the same for our setup.

dmth:
> After a short discussion with @dmth we remove this layer. data in extra is additional data which does not fit in the currently configured harmonization. This does not mean, that we don't know the meaning, we haven't validated it etc. The source is also clear: the feed itself (feed.name).

True, we talked about it. Still I'm not sure if it's a good idea to remove it. The removal requires to solve the problem for fields which have a high chance for collisions (for instance `size`). I think it's not good to rename these fields (for instance `size -> request_size`) as this renaming operation has to be made very transparent in order to be aware how this data was created and where did it come from. 

> Why don't you create an extra field in your harmonization? You'll never have collisions and it's much easier to work with. We do the same for our setup.

I feel that this creates a new Issue: If everyone has his own "harmonisation" data is not harmonised anymore. This will create fragmentation!
From my understanding the harmonisation is used to have a common, shared view on the events. It should not be required to add extra fields to the harmonisation just because some bots need it.

I'm not convinced that these are sufficient, long lasting solutions.

aaronkaplan:
> On 18 Jul 2016, at 10:07, Dustin Demuth <notifications@github.com> wrote:
> 
> After a short discussion with @dmth we remove this layer. data in extra is additional data which does not fit in the currently configured harmonization. This does not mean, that we don't know the meaning, we haven't validated it etc. The source is also clear: the feed itself (feed.name).
> 
> True, we talked about it. Still I'm not sure if it's a good idea to remove it. The removal requires to solve the problem for fields which have a high chance for collisions (for instance size). I think it's not good to rename these fields (for instance size -> request_size) as this renaming operation has to be made very transparent in order to be aware how this data was created and where did it come from.
> 
> Why don't you create an extra field in your harmonization? You'll never have collisions and it's much easier to work with. We do the same for our setup.
> 
> I feel that this creates a new Issue: If everyone has his own "harmonisation" data is not harmonised anymore. This will create fragmentation!
Yes, I agree.
> From my understanding the harmonisation is used to have a common, shared view on the events. It should not be required to add extra fields to the harmonisation just because some bots need it.
> 
> I'm not convinced that these are sufficient, long lasting solutions.

okay, understood your point and I also agree.

So... one solution here is to  allow some custom field set then ("x-local").
(and have a registry mechanism how to register new fields officially).


But in any case, I feel the "extra" field is not the place to put it to. We always used "extra" to put "fields by the source which we won't process within the DHO but we still want to pass along to a client".
So, it's really just a JSON dict (flat) for all the "maybe-interesting" fields from the original message.

See the definition in the document:
extra: "All anecdotal information, which cannot be parsed into the data harmonization elements. E.g. os.name, os.version, etc. Note: this isonly intended for mapping any fields which can not map naturally into the data harmonization. It is not intended for extending the data harmonization with your own fields."


I hope this clarifies the extra field a bit more.
Yes, we should document it better.
Let's discuss this in the call on Tuesday.

sebix:
> > Why don't you create an extra field in your harmonization? You'll never have collisions and it's much easier to work with. We do the same for our setup.

> I feel that this creates a new Issue: If everyone has his own "harmonisation" data is not harmonised anymore. This will create fragmentation!

The `notification information the certbund_contact bot` is very specific to your workflow. No one else has this (unless the bot is general enough and included in this repo, then we can add such a field here too).

PR #566, opened 2016-07-13
RTIR Collector

sebix:
Collects reports from a RTIR-instance and passes them other bots (parsers). Can also take events.

Used in our setup at cert.at


aaronkaplan:
merge and will massage this some more.

PR #575, opened 2016-07-19
Modify bot: Prepare for more example configs

gsiv:
In order to be able to add more configurations without filling up the bot's root directory I would like to move the configs to an `example` subdirectory.  We already have an additional modify.conf that we would like to add.


aaronkaplan:
looks good for me. @sebix ?

aaronkaplan:
The build is failing here...why?

sebix:
The configuration couldn't be loaded :D

aaronkaplan:
Yeah I saw that in the nose test output. But why?

---
Mobile

sebix:
Because the filename in [intelmq/tests/bots/experts/modify/test_expert.py#L55](https://github.com/certtools/intelmq/blob/master/intelmq/tests/bots/experts/modify/test_expert.py#L55) needs to be adapted accordingly.

aaronkaplan:
ok, travis-CI gave green lights. Thx!

PR #578, opened 2016-07-20
Use destination.fqdn for Botnet-Drone-Hadoop's cc_dns

bernhard-herzog:
The description of the "cc_dns" field of shadowserver's
Botnet-Drone-Hadoop feed is "Fully qualified domain name of the C&C"
which matches destination.fqdn better than destination.reverse_dns.

See also the discussion in https://github.com/certtools/intelmq/pull/555#issuecomment-232702475


aaronkaplan:
thanks!

Issue #556, opened 2016-07-07
Debian-Packages must include bots' documentation files.


bernhardreiter:
In general should the documentation of the bots be included in the package.

Example:
The debian Package intelmq (1.0.0~dev4+intevation.1.2~alpha3) from
https://github.com/Intevation/intelmq/blob/deb-packaging/debian/changelog 
does not package all README.md files from bots. 
E.g. the postgresl/README.md is missing.

@gsiv is this something you can look into?

gsiv:
On Thu, Jul 07, 2016 at 06:29:59AM -0700, Bernhard E. Reiter wrote:
> @gsiv is this something you can look into?

Yes, I'll look into it.

PR #579, opened 2016-07-20
Ubuntu packages: various fixes and improvements

gsiv:
* adjust paths for recently renamed files
* include all bot READMEs (see issue #556)
* use quilt to maintain patches for various files 
  * adjust paths in the docs
  * undo rename of dnspython dependency


debian/intelmq.install
```diff
@@ -1,2 +1,3 @@
-CHANGELOG.md LICENSE COPYRIGHT MANIFEST.in REQUIREMENTS docs/*md usr/share/doc/intelmq/
```
sebix: The LICENSE file is removed here, but not added (in this bunch of commits). Is it still shipped?

gsiv: On Wed, Jul 20, 2016 at 10:01:01AM -0700, Sebastian wrote:
> The LICENSE file is removed here, but not added (in this bunch of
> commits). Is it still shipped?

Good eye!  It is being left out on purpose because it should also be
included in debian/copyright.

Unfortunately, it isn't yet.  I'm going to fix that right away.

debian/conffiles
```diff
@@ -1,9 +1,8 @@
-/opt/intelmq/var/lib/bots/modify/example/modify.conf
```
aaronkaplan: is that now in example/ or in the other directory?

gsiv: On Mon, Jul 25, 2016 at 05:44:42AM -0700, AaronK wrote:
> > @@ -1,9 +1,8 @@
> > -/opt/intelmq/var/lib/bots/modify/example/modify.conf
> 
> is that now in example/ or in the other directory?
> Isn't it called exampleS ?

No, due to setup.py:30 it is indeed
/opt/intelmq/var/lib/bots/modify/example.

> We are unsure what you intend here...

The `conffiles` file merely declares configuration files.  While the
files in example/ will still be there they don't need to be treated
as configuration files by the package manager;¹ hence, the removal
of that line.

A copy of the default.conf example is installed as
/opt/intelmq/var/lib/bots/modify/modify.conf by the package.  This
file is actually used as a configuration file, and thus declared as
one here.

  ¹) We can simplify things, though.  Since the modify bot configs
     also get copied to the /usr/share/doc hierarchy the original
     location has become redundant.  We should probably get rid of
     the whole /opt/intelmq/var/lib/bots/modify/example directory.

PR #581, opened 2016-07-20
Botnet-Drone-Hadoop's count field is actually optional in real data

bernhard-herzog:
In the test data we have, the count field of the shadowserver feed
Botnet-Drone-Hadoop the count field is often empty. Therefore, use
convert_int instead of int as the conversion function.


aaronkaplan:
thanks!

PR #582, opened 2016-07-20
shadowserver's Open-Mongo DB feed does not have a username field

bernhard-herzog:
Remove the rule for the field "username" from the shadowserver parser's
configuration for the Open-Mongo DB feed because it does not actually
exist in the feed. In the new parser fails to parse the feed as a
consequence (which is a good thing!).


aaronkaplan:
thanks!

Issue #587, opened 2016-07-21
Harmonization type for lowercase strings


sebix:
Result of discussion in #585:
Define a new harmonization type for lowercase-only strings. Sanitation automatically converts them to lowercase.

* apply for types in harmonization e.g. protocol names etc.
* eventually remove regex-checks which are now obsolete
* add tests

aaronkaplan:
actually i think I can solve it easier: add a new iregex type in the harmonisation.

sebix:
But don't we want to convert the strings to lower case too? Just ignoring the case could lead to many other potential issues, like duplicates.

bernhardreiter:
bin/intelmq_psql_initdb.py has not been considered.
It does not know the new type and it still has the old UUID type.

PR #588, opened 2016-07-21
add re.IGNORECASE (iregex) support to the harmonisation.conf

aaronkaplan:
add the ignore case thing for regexps for the harmonisation

### DIFF
--- a/intelmq/etc/harmonization.conf
+++ b/intelmq/etc/harmonization.conf
@@ -182,7 +182,7 @@
         },
         "protocol.transport": {
             "description": "e.g. tcp, udp, icmp.",
-            "regex": "^[ -~]{1,10}$",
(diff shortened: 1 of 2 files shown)


aaronkaplan:
comments?

sebix:
I'd like to discuss this and #587 first, please don't merge yet.

aaronkaplan:
ok

aaronkaplan:
discussed this with @sebix : for the case of transport protocol definitons ("udp" vs. "UDP")  we agreed to normalise by lowercasing everything. However, I believe keeping the "iregex" possibility in lib/message.py in a742c034c80249e80d2f280fd0dc8a72236e8a57 is actually handy right now. We can remove it, should it not be needed anywhere else.

PR #589, opened 2016-07-21
Modify bot: Add an additional example config

gsiv:
I'm not crazy about the file name but here it goes. See also PR #575.

### DIFF
--- a/intelmq/bots/experts/modify/examples/morefeeds.conf
+++ b/intelmq/bots/experts/modify/examples/morefeeds.conf
@@ -1,0 +1,239 @@
+{
+"default": {
+    "__default": [{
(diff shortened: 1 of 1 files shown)


aaronkaplan:
thanks!

sebix:
Is this just an extended version of the existing example?

gsiv:
On Thu, Jul 21, 2016 at 08:49:35AM -0700, Sebastian wrote:
> Is this just an extended version of the existing example?

Yeah, pretty much, I guess.  Not having written it I don't know for
for sure, though.  @dmth probably knows the details.

dmth:
yes, it is. A lot of stuff is not required for our usecase.

PR #595, opened 2016-07-22
Deb package: Change fewer intelmq home dir permissions

### LINKED ISSUE #559: Debian-Packages: permissions of /opt/intelmq problems with .ssh/authorized_keys
https://github.com/certtools/intelmq/blob/3e82335a8ee75b525d9c89b87a42c532bd509ed2/debian/intelmq.postinst has the problem that Ubuntu 14.04 sshd does not want to allow connections
via .ssh/authorized_keys if a home directory is group writable.

```sh
        useradd -d /opt/intelmq -U -s /bin/bash intelmq
    chmod -R g+w /opt/intelmq
```

The resulting error message in /var/log/auth.log looks like
```
Jul  7 17:06:24 ubuntu sshd[10993]: Authentication refused: bad ownership or modes for directory /opt/intelmq
```

intelmq-manager will have to write some files in the directory with the "intelmq" group.
Two possible solutions:
 1. check that the top level directory does not need to be written and remove the line
 2. use a different home directory for the intelmq user

@swilde created the file initially.

gsiv:
Only change permissions of its contents, i.e., `var` and `etc`.

This should resolve #559.

### DIFF
--- a/debian/intelmq.postinst
+++ b/debian/intelmq.postinst
@@ -5,7 +5,7 @@ if [ "$1" = "configure" ] ; then
     if ! getent passwd intelmq >/dev/null 2>&1; then
         useradd -d /opt/intelmq -U -s /bin/bash intelmq
     fi
-    chmod -R g+w /opt/intelmq
+    chmod -R g+w /opt/intelmq/*
     chown -R intelmq:intelmq /opt/intelmq
 fi
 


aaronkaplan:
looks ok. thx.

sebix:
Is `/opt/intelmq/etc` writable for www-data?

aaronkaplan:
Yes, I usually do this manually. However in this setup, the problem is that it gets overwritten by the line: 
``bash
chown -R intelmq:intelmq /opt/intelmq
``
So, no... that needs to change in case intelmq-manager is installed. @gsiv

gsiv:
On Fri, Jul 22, 2016 at 09:53:41AM -0700, Sebastian wrote:
> Is `/opt/intelmq/etc` writable for www-data?

It is if www-data is in the group intelmq.

The intelmq-manager deb package adds it for you.

PR #604, opened 2016-07-23
Add script to rewrite config files so travis is happy.

Rafiot:
Very simple script to rewrite the config files a the way the travis tests pass.


aaronkaplan:
should we put this into contrib/ ?

sebix:
Thanks for the fix, I added docs and error handling (and moved the file to intelmq/bin).

PR #605, opened 2016-07-23
Update misp module

Rafiot:
* First commit is a bugfix, feel free to just cherry-pick this one.
* Second commit is required for the 3rd one
* Third commit add a bunch of changes in the MISP bots, but I think they are relevant and useful for most of the intelMQ users.

### DIFF
--- a/intelmq/bots/parsers/misp/parser.py
+++ b/intelmq/bots/parsers/misp/parser.py
@@ -60,60 +60,62 @@ class MISPParserBot(Bot):
     def process(self):
         report = self.receive_message()
         raw_report = utils.base64_decode(report.get('raw'))
-        misp_events = json.loads(raw_report)
+        misp_event = json.loads(raw_report)
 
-        # iterate through the results
-        for misp_event in misp_events:
+        # Set the classifier based on the ecsirt tag
+        classifier = None
+        if misp_event.get('Tag'):
+            for tag in misp_event['Tag']:
+                if tag['name'] in self.MISP_TAXONOMY_MAPPING:
+                    classifier = self.MISP_TAXONOMY_MAPPING[tag['name']]
+                    break
 
-            # Set the classifier based on the ecsirt tag
-            classifier = None
-            if misp_event.get('Tag'):
-                for tag in misp_event['Tag']:
-                    if tag['name'] in self.MISP_TAXONOMY_MAPPING:
-                        classifier = self.MISP_TAXONOMY_MAPPING[tag['name']]
-                        break
+        # get the attributes from the event
+        event_attributes = misp_event['Attribute']
 
-            # get the attributes from the event
-            event_attributes = misp_event['Attribute']
+        # payload type - get malware variant for the event
+        malware_variant = None
+        for attribute in event_attributes:
+            if attribute['category'] == 'Payload type':
+                value = attribute['value'].lower()
+                # TODO: use misp galaxies
+                if value and harmonization.MalwareName.is_valid(value):
+                    malware_variant = value
 
-            # payload type - get malware variant for the event
-            malware_variant = None
-            for attribute in event_attributes:
-                if attribute['category'] == 'Payload type':
-                    value = attribute['value'].lower()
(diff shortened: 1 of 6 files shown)


intelmq/etc/harmonization.conf
```diff
@@ -170,9 +170,14 @@
             "regex": "^[ -~]+$",
             "type": "String"
         },
+        "misp_attribute_uuid": {
```
aaronkaplan: let's call this "misp.attribute_uuid"
and "misp.*"

intelmq/bots/parsers/misp/parser.py
```diff
@@ -60,60 +60,62 @@ class MISPParserBot(Bot):
...
+                # FIXME: Send the whole MISP event with each attribute?
+                event.add('raw', json.dumps(misp_event, sort_keys=True))
+                event.add(self.MISP_TYPE_MAPPING[type_], value)
+                event.add('misp_uuid', misp_event['uuid'])
+                event.add('misp_attribute_uuid', uuid)
+                event.add('comment', comment)
+                event.add('event_description.text', category)
+                event.add('event_description.url', misp_event_url)
+                event.add('malware.name', malware_variant)
+                event.add('classification.type', classifier)
```
aaronkaplan: what are possible values for "classifier"?

aaronkaplan: at this stage in the code, can we also pls set the "classification.taxonomy" field in intelmq?
Basically it is the eCSIRT II taxonomy field (see MISP_TAXONOMY_MAPPING, the left most field.
For example:
```
ecsirt:malicious-code="malware-configuration" 
```
taxonomy = "malicious code"
Please also see https://github.com/certtools/intelmq/blob/master/docs/Data-Harmonization.md#mapping 
)

intelmq/etc/harmonization.conf
Rafiot: misp.attribute_uuid and misp.event_uuid?

aaronkaplan: > On 25 Jul 2016, at 14:20, Raphaël Vinot <notifications@github.com> wrote:
> 
> In intelmq/etc/harmonization.conf:
> 
> > @@ -170,9 +170,14 @@
> >              "regex": "^[ -~]+$",
> >              "type": "String"
> >          },
> > +        "misp_attribute_uuid": {
> 
> misp.attribute_uuid and misp.event_uuid?

I think this makes the most sense.... yes.
What do you think?

intelmq/bots/parsers/misp/parser.py
Rafiot: In the current code, the only possible classifiers you will have are the values in the dict `MISP_TAXONOMY_MAPPING` (based on the ecsirt taxonomy), but it will only work if the MISP events are tagged with the ecsirt taxonomy, any other taxonomy will be ignored.

What do you think of a more generic approach that would use the taxonomies in https://github.com/MISP/misp-taxonomies, and import all those taxonomies by default in intelmq?

intelmq/etc/harmonization.conf
Rafiot: yep, totally. We may want to add more misp.* fields in the future.

aaronkaplan: okay, thanks!
Once this is fixed, I'll merge it in. Thx!

intelmq/bots/parsers/misp/parser.py
aaronkaplan: Probably a very good idea.
However, we are doing a sprint for releasing intelmq 1.0 by 30th of August.
So , we won't add a lot of fundamental changes right now. But yes, please do create an issue and assign it to release 1.1 or 2.0 so that we can include that in the next release.

Rafiot: Sure, no hurry. 

https://github.com/MISP/PyTaxonomies/ <= the code to manage the taxonomies will look like that

aaronkaplan:
thx!

PR #609, opened 2016-07-25
Deb package: Remove redundant modify bot example configs

gsiv:
A regular Python installation places the modify expert bot example
configs in /opt/intelmq/var/lib/bots/modify/example.  The deb package
copies them to the bot's documentation directory in
/usr/share/doc/intelmq.

This patch removes the files from their original location to avoid
duplication.


aaronkaplan:
thx

PR #610, opened 2016-07-25
Shadowserver parsing bug cleanups

aaronkaplan:


SYNchroACK:
Thank you all for working on this one, as far I can see, great work, this is the kind of bot that we need! :) If you guys don't mind, I would like to add a few suggestions that I thought about when I look at code and comments:

* In a overview, I would suggest to have the expert.py with all code and shadowserver.config (json). 

* [import config](https://github.com/certtools/intelmq/blob/master/intelmq/bots/parsers/shadowserver/parser.py#L24) and [config] - (https://github.com/certtools/intelmq/blob/master/intelmq/bots/parsers/shadowserver/config.py#L131) In my perspective we can use a json configuration instead of config.py that will load as a dictionary.

* [IP "sanitizing"](https://github.com/certtools/intelmq/blob/master/intelmq/bots/parsers/shadowserver/config.py#L124) - I think the rfc1918 expert is doing this work. We can give this responsibility to him.

* [set_tor_node](https://github.com/certtools/intelmq/blob/master/intelmq/bots/parsers/shadowserver/config.py#L124) - Tried to understand (dont have access to shadowserver at the moment) and didn't figure out. As you know, there is also a tor_node expert. Should we rely on that bot to set the tor_node value?

Looking forward to receive your feedback on this, the hard work is done. Thank you once more! :)

sebix:
> In a overview, I would suggest to have the expert.py with all code and shadowserver.config (json). 

It would be messy and only partly doable (think of functions) and the configuration is distributed by us anyway. The administrator/user of an intelmq-instance does not need to modify this config.

> IP "sanitizing" - I think the rfc1918 expert is doing this work. We can give this responsibility to him.

Invalid values are never accepted by the system.

> set_tor_node - Tried to understand (dont have access to shadowserver at the moment) and didn't figure out. As you know, there is also a tor_node expert. Should we rely on that bot to set the tor_node value?

The data has such a flag.

PR #617, opened 2016-07-26
Universal shadowserver parser

swilde:
A bunch of fixes for the shadowserver parser.


sebix:
Concerning the raw-field: I created the workaround (`conv_csv_shadowserver` etc) to create an output that is as close as possible to the original.

I agree it would be much easier if the csv-library would provide the original line and/or the line number, and ignored lines. That would be really great. Maybe we can find some library that's doing this, or fork the csv parser ourself and implement this. (In the future, not now as bugfix)

@aaronkaplan @otmarlendl What do you think is the correct approach here?
1. Maintain only the CSV format and minimize code, changing the formatting of the data
2. have an as similar as possible copy of the original data

Concerning merge conflicts: I will merge these fixes together with the ones in https://github.com/certtools/intelmq/tree/shadowserver-bugfixes and resolve conflicts manually anyway. You don't need to fix them beforehand.

Issue #547, opened 2016-06-16
Large reports don't fit into one redis message.


dmth:
As mentioned in the newsletter, large (> 500MB) reports do not fit into the message queue.

We expect an updated redis > 3.2 should work for these large messages.
But this is not yet tested.
- [ ] Who is running a very recent redis and could verify this?
- [x] In addition: the length limit in https://github.com/certtools/intelmq/blob/master/intelmq/etc/harmonization.conf#L336 should be removed.

_EDIT by sebix: checklist_

dmth:
Did you run a test how the application performs with large (>512 MB) reports now? If yes, which Redis version did you use?

aaronkaplan:
We never encountered some large reports like this in practice. So we did not come across it yet. Need to test via generated reports.

sebix:
No success with redis 3.2.1 so far. On server side I get this:

> 7491:M 19 Jul 12:41:46.548 - Protocol error from client: id=38 addr=127.0.0.1:47459 fd=5 name= age=15 idle=0 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=16384 qbuf-free=16384 obl=42 oll=0 omem=0 events=r cmd=lpush
> 7491:M 19 Jul 12:41:46.699 - Protocol error from client: id=39 addr=127.0.0.1:47460 fd=5 name= age=0 idle=0 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=16384 qbuf-free=16384 obl=42 oll=0 omem=0 events=r cmd=NULL

On client side:
```python
>>> r.lpush('a', '01'*260*1024*1024)
Traceback (most recent call last):
  File "<ipython-input-62-2885976fd64e>", line 1, in <module>
    r.lpush('a', '01'*260*1024*1024)
  File "/usr/local/lib/python3.4/dist-packages/redis/client.py", line 1227, in lpush
    return self.execute_command('LPUSH', name, *values)
  File "/usr/local/lib/python3.4/dist-packages/redis/client.py", line 578, in execute_command
    connection.send_command(*args)
  File "/usr/local/lib/python3.4/dist-packages/redis/connection.py", line 563, in send_command
    self.send_packed_command(self.pack_command(*args))
  File "/usr/local/lib/python3.4/dist-packages/redis/connection.py", line 556, in send_packed_command
    (errno, errmsg))
ConnectionError: Error 104 while writing to socket. Connection reset by peer.
```

But also see the docs at http://redis.io/topics/data-types
[...]

PR #629, opened 2016-08-04
Use an unlimited type for the raw field in the events table

bernhard-herzog:
This is a somewhat ad-hoc fix for the problem that the raw field can be
quite large and may not fit into the 2000 characters that are assumed if
no length is specified for a `Base64` field in `harmonization.conf`. There's
probably a better solution but for now we only need it for the raw field
and that's the only field that uses the Base64 type.

See issue #547

A better solution would probably be

```python
        if value['type'] in ('String', 'Base64', 'URL', 'FQDN',
                             'MalwareName', 'ClassificationType'):
            length = value.get('length')
            if length is not None:
                dbtype = 'varchar({})'.format(length)
            else:
                dbtype = 'text'
```

But that would change the SQL types of all fields that do not explicitly have a length in `harmonization.conf`. OTOH, that would actually be correct, since they could be of any length in the event. As long as `intelmq_psql_initdb.py` puts an arbitrary length limit on fields that are actually unbounded it can happen that an event can not be stored in the event database.

### DIFF
--- a/intelmq/bin/intelmq_psql_initdb.py
+++ b/intelmq/bin/intelmq_psql_initdb.py
@@ -29,9 +29,11 @@ def main():
     for field in DATA.keys():
         value = DATA[field]
 
-        if value['type'] in ('String', 'Base64', 'URL', 'FQDN',
+        if value['type'] in ('String', 'URL', 'FQDN',
                              'MalwareName', 'ClassificationType'):
             dbtype = 'varchar({})'.format(value.get('length', 2000))
+        elif value['type'] == 'Base64':
+            dbtype = 'text'
         elif value['type'] in ('IPAddress', 'IPNetwork'):
             dbtype = 'inet'
         elif value['type'] == 'DateTime':


sebix:
I prefer the better solution posted by you. We should then set a length where useful and use text for all others. This would affect the following keys:

- destination.geolocation.cc
- malware.hash.sha1
- destination.as_name
- destination.geolocation.city
- source.geolocation.geoip_cc
- source.geolocation.country
- destination.registry
- source.registry
- classification.taxonomy
- destination.abuse_contact
- feed.name
- source.account
- source.geolocation.city
- malware.version
- status
- misp.event_uuid
- destination.geolocation.country
- source.as_name
- destination.local_hostname
- classification.identifier
- destination.geolocation.region
- malware.hash.md5
- source.geolocation.region
- comment
- malware.name
- source.geolocation.state
- source.geolocation.cc
- protocol.transport
- destination.geolocation.state
- destination.account
- source.local_hostname
- malware.hash
- misp.attribute_uuid
- source.geolocation.cymru_cc
- event_description.target
- event_description.text
- source.abuse_contact
- event_hash

We could also save the raw-field directly without encoding. Postgres supports binary data, that's no problem. (but would break backwards-compatibility)

aaronkaplan:
what about migrations of existing DBs? Will we need a dump & restore mechanism?

sebix:
> what about migrations of existing DBs? Will we need a dump & restore mechanism?

type-changes can be executed on live systems. we can provide such alter-scripts tough. But first we need a consensus.

Some of the fields posted above do have length limits in regex-checks. Will add them first, then we need to deal with only a subset (create new length-restrictions or convert to text).

PR #631, opened 2016-08-09
Fix tests (AS name changed)

Rafiot:


aaronkaplan:
looks good, thanks , pulling in.

PR #636, opened 2016-08-11
Fixed MISP parser issue with raw field

kralca:
Raw field previously contained the entire MISP event for each attribute, resulting in massive duplication. It now just stores the JSON data for the MISP attribute.

### DIFF
--- a/intelmq/bots/parsers/misp/parser.py
+++ b/intelmq/bots/parsers/misp/parser.py
@@ -103,8 +103,7 @@ class MISPParserBot(Bot):
 
                 # Create and send the intelmq event
                 event = Event(report)
-                # FIXME: Send the whole MISP event with each attribute?
-                event.add('raw', json.dumps(misp_event, sort_keys=True))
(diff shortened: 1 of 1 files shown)


sebix:
@Rafiot please review

Rafiot:
It will work. I didn't change it before because I'm not sure what people expect to find in the `raw` key in general.

aaronkaplan:
looks okay for me. It's really the question what you want in the raw field. Potentially you can have both (event and attributes) ... but I leave this decision to the MISP <-> intelmq users. You know best what's needed. Thanks for the PR!!

kralca:
The decision is really a trade-off. If a single MISP event has many attributes then it doesn't make sense to attach the full MISP event to each IntelMQ event. On the other hand, there is some additional information in the MISP event that needs to be extracted and added to the IntelMQ event otherwise it may be lost. Thanks for merging!

kralca:
Thanks @sebix and sorry for missing the test code.

PR #640, opened 2016-08-16
RT collector: use observation time from RT-ticket

sebix:
Fortunately, the time is already UTC (at least in `get_history` as opposed to `get_ticket`m which has local time...)


aaronkaplan:
looks good, thx

PR #648, opened 2016-08-24
Shadowserver config fixups

aaronkaplan:
added open XDMCP (open X11 display manager ports) parser config snippet


sebix:
no tests for all three new feeds

aaronkaplan:
> On 24 Aug 2016, at 19:53, Sebastian <notifications@github.com> wrote:
> 
> no tests for all three new feeds
> 

Not finished yet :)

intelmq/bots/parsers/shadowserver/config.py
```diff
@@ -867,8 +868,40 @@ def validate_ip(value):
         'classification.type': 'vulnerable service',
         'classification.taxonomy': 'Vulnerable',
         'protocol.application': 'xdmcp',
-        'feed.code': 'shadowserver-openmdns',
+        'feed.code': 'shadowserver-openxdmcp',
```
dmth: Are you sure this is correct?

aaronkaplan:
still open.

aaronkaplan:
okay, whatever, merging now, will add tests later. Re-opening an issue for the test cases.

sebix:
This PR has never been reviewed.

And I'm still against setting `feed.code` and `feed.name` in parsers. This is totally against our concept.

aaronkaplan:
okay, so from the git comment log , I only saw that you were against the missing tests.
We can revert / undo the feed.code / feed.name part in the parsers.
But I'd like to understand why? I think after all, we will do exactly that in the parsers of all other feeds (not SS).

dmth:
This merge has breaks compatibility with our branch, due to the alterations at `feed.name`. We use and need this feed to identify the templates that need to be used if a contact is notified. 

After the alteration, shadowserver feeds cannot be distinguished anymore by `feed.name`.

In additon, setting the `Feed.name` in this way contradicts: https://github.com/certtools/intelmq/blob/master/intelmq/bots/parsers/shadowserver/parser.py#L82 where the Feedname is determined by the name which was used to select the correct config (You remember our discussion on this).

We might adapt to use the `feed.code` but also those keys are not the same like  the `feed.name` before
(`shadowserver-sinkhole-http-drone` instead of `Sinkhole-HTTP-Drone`)  there is a lot of migration necessary in order to be capable to use IntelMQ after this change was.

In addition, my gut feeling is that setting the `feed.name` to `shadowserver` is not correct. Shadowserver would be the `feed.provider` (afaik, that field does not exist yet). The `feed.name` would be the name how the provider calls that feed (Sinkhole-HTTP-Drone). I'll call you today if possible. Maybe we can resolve this more simple then.

aaronkaplan:
> On 06 Oct 2016, at 09:47, Dustin Demuth <notifications@github.com> wrote:
> 
> This merge has breaks compatibility with our branch, due to the alterations at feed.name. We use and need this feed to identify the templates that need to be used if a contact is notified.
> 
> After the alteration, shadowserver feeds cannot be distinguished anymore by feed.name.
> 
> In additon, setting the Feed.name in this way contradicts: https://github.com/certtools/intelmq/blob/master/intelmq/bots/parsers/shadowserver/parser.py#L82 where the Feedname is determined by the name which was used to select the correct config (You remember our discussion on this).
> 
> We might adapt to use the feed.code but also those keys are not the same like the feed.name before
> (shadowserver-sinkhole-http-drone instead of Sinkhole-HTTP-Drone) there is a lot of migration necessary in order to be capable to use IntelMQ after this change was.
> 
> In addition, my gut feeling is that setting the feed.name to shadowserver is not correct. Shadowserver would be the feed.provider (afaik, that field does not exist yet). The feed.name would be the name how the provider calls that feed (Sinkhole-HTTP-Drone). I'll call you today if possible. Maybe we can resolve this more simple then.


okay... point taken.
Yeah, I remember the discussion. Let's try to solve this on the phone.
I am best reachable after 14:00

Best,
a.

Issue #647, opened 2016-08-23
Modify Expert: How to extend the bot "the right way": Replacement of substrings


dmth:
Hey folks,

**Problem**
We are currently facing a problem with our modify-expert configuration. For all events which carry the string `bitdefender-*` as a `malware.name` we need to remove the part `bitdefender-`. Normally one could do that with a `str.replace()`, but that is not possible here.

We tried to use this approach in our configuration, (as visible in https://github.com/certtools/intelmq/blob/master/intelmq/bots/experts/modify/examples/morefeeds.conff) but sadly it does not work. The format method does not allow the use of functions within the format string.

**Solution?**
A possible solution: @bernhard-herzog extended the modify-bot in order to allow the reuse of match-groups (see below).

**Questions**
We are not sure if such an extension is the right way to go. It makes the configuration file even harder to read and increases complexity of the modfiy expert.

Might a "Str-Replace"-Expert which can only fulfill the task of replacing Stings in Events be the better way to go?

What is your Opinion? Do you have a similar use-case?

BR
Dustin

**Patch**

```
diff --git a/intelmq/bots/experts/modify/examples/morefeeds.conf b/intelmq/bots/experts/modify/examples/morefeeds.conf
index d16572e..61aac92 100644
--- a/intelmq/bots/experts/modify/examples/morefeeds.conf
+++ b/intelmq/bots/experts/modify/examples/morefeeds.conf
@@ -224,9 +224,9 @@
             "classification.identifier": "Dridex"
         }],
     "bitdefender" : [{
-            "malware.name": "bitdefender-$"
+            "malware.name": "bitdefender-(.*)$"
         }, {
-            "malware.name": "{msg[malware.name].replace('bitdefender-', '')}"
+            "malware.name": "{matches[malware.name][1]}"
         }]
     },
 "SecurityScorecard": {
diff --git a/intelmq/bots/experts/modify/expert.py b/intelmq/bots/experts/modify/expert.py
index 0a687cc..5b54a54 100644
--- a/intelmq/bots/experts/modify/expert.py
+++ b/intelmq/bots/experts/modify/expert.py
@@ -9,6 +9,15 @@ from intelmq.lib.bot import Bot
 from intelmq.lib.utils import load_configuration
 
 
+class MatchGroupMapping:
+
+    def __init__(self, match):
+        self.match = match
+
+    def __getitem__(self, key):
+        return self.match.group(key)
+
+
 class ModifyExpertBot(Bot):
 
     def init(self):
[...]

Issue #662, opened 2016-09-01
modify bot: overhaul/redesign improving usability


bernhardreiter:
There are a number of issues about the modify export bot, which may be best
solved by a larger redesign. This issue aims to bundle the aspects.

The issues are:
1. Executing of some groups and rules is non-deterministic, because it depends on iterating over
  python dictionaries, which has an undefined order.
2. The configuration file is too hard to read. (This is the configuration representation of the logic.)
    Reported as #404 (and seen in other problems with getting modify bot to do as it would)
3. It is hard to understand, what the bots will do (this is execution logic as opposed to 2.)

The discussion in #647 has shown some of the problems and a potential solution rewriting
the bot, which is discussed here: https://github.com/certtools/intelmq/compare/master...sebix:dev-modify-bot-regex-replace?expand=1

bernhardreiter:
Any opinions? Maybe @swilde ?

aaronkaplan:
@bernhardreiter basically the modify bot's syntax is really not readable.
We need a simple:

```
IF /pattern/ AND /pattern/ AND ... THEN
   update(key, new_value)
END IF
```

Structure.
It needs to be readable.

aaronkaplan:
Basically this issue is a duplicate of #404

bernhardreiter:
@aaronkaplan in order to make it more readable, I've proposed the simplification above.
Its logic is much clearer and comes a log closer to your IF_THEN pattern. However I believe
the IF_THEN pattern does not translate nicely into json. My proposal kept the idea of keeping the rules
as a configuration in a json file.

@sebix: What do you mean by "fix sub-list syntax", in my proposal there would be one 'list' which has elements of "(conditions and actions)".

aaronkaplan:
> On 05 Oct 2016, at 22:49, Bernhard E. Reiter <notifications@github.com> wrote:
> 
> @aaronkaplan in order to make it more readable, I've proposed the simplification above.
> Its logic is much clearer and comes a log closer to your IF_THEN pattern.

It was an example for *readable* syntax.

I seriously doubt that JSON is the right format for describing any of this logic.

> However I believe
> the IF_THEN pattern does not translate nicely into son.

Forget JSON here please

> My proposal kept the idea of keeping the rules
> as a configuration in a json file.
> 
Nope. I really don't want to edit if-then rules in JSON.
Not readable at all.

bernhardreiter:
[...]

PR #663, opened 2016-09-01
Modify Bot: (major) improvements.

bernhardreiter:
Closes https://github.com/certtools/intelmq/issues/647 .

Contains:
-  the new way to  access match groups from the conditions
- overhauled description, now matching the behaviour
- more unit tests: the first for the morefeeds.conf examples.


docs/Bots.md
```diff
@@ -350,15 +355,43 @@ The configuration is called `modify.conf` and looks like this:
...
+We have an event with `feed.name = Spamhaus Cert` and `malware.name = confickerab`. The expert loops over all sections in the file and eventually enters section `Spamhaus Cert`. First, the default condition is checked, it matches! Ok, going on. Otherwise the expert would have selected a different section that has not yet been considered. Now, go through the rules, until we hit the rule `conficker`. We combine the conditions of this rule with the default conditions, and both rules match! So we can apply the action: `classification.identifier` is set to `conficker`, the trivial name.
 
-Assume we have an event with `feed.name = Spamhaus Cert` and `malware.name = feodo`. The default condition matches, but no others. So the default action is applied. The value for `classification.identifier` is `{msg[malware.name]}`, this is [standard Python string format syntax](https://docs.python.org/3/library/string.html#format-string-syntax). Thus you can use any value from the processed event, which is available as `msg`.
```
sebix: I'd keep the link to python's docs, as bot all users of this software may be aware of the string format syntax including it's possibilities.

intelmq/bots/experts/modify/examples/morefeeds.conf
```diff
@@ -3,9 +3,7 @@
     "__default": [{
             "classification.identifier": "",
             "malware.name": ".*"
-        }, {
-            "classification.identifier": "{msg[malware.name]}"
```
sebix: Why has this line been removed?

intelmq/bots/experts/modify/expert.py
```diff
@@ -19,60 +33,73 @@ def matches(self, identifiers, event, *rules):
...
+                if match is None:
+                    return None
+                else:
+                    matches[name] = match
 
-        return True
+        return matches
 
-    def apply_action(self, event, action):
+    def apply_action(self, event, action, matches={}):
```
bernhard-herzog: What's the purpose of the default value for the matches parameter? It's not needed. All places where the method is called provide a value for this parameter. Also, the default value is mutable, which can lead to subtle problems ([see Python FAQ](https://docs.python.org/3/faq/programming.html#why-are-default-values-shared-between-objects))

docs/Bots.md
bernhardreiter: The link to the python do has been moved to a better section, away from the examples to the place where the actions are described (line 384).

sebix: Ah, ok. Missed that one.

intelmq/bots/experts/modify/examples/morefeeds.conf
bernhardreiter: See the rationale in https://github.com/certtools/intelmq/pull/663/commits/8a775c55041b78c73eedaa6f6bc112774226549c  It never worked reliably.
If this functionality is still needed, it can be covered by adding a second modify bot.

sebix: ok, will be solved by #662

intelmq/bots/experts/modify/expert.py
bernhardreiter: Good catch, corrected!

(Details: the parameters were missing in the second call initially, so adding a default parameter was my first fix. Later I did add the parameter in the second call and thought leaving it in would make it more robust.)

docs/Bots.md
bernhardreiter: No problem, thanks for reviewing!

PR #667, opened 2016-09-06
Modify Bot: (minor) improvements DOC

bernhardreiter:
Better examples in morefeeds.conf that are also part of the test. In addition improved the DOC section that describes the behaviour.


sebix:
Fine for me.

PR #669, opened 2016-09-08
XMPP Collector and Output, JSON parser

sebix:
Based on #500 

The json parser is just copying the data. Validation is done by the message class itself (which I fixed now to validate data given on initialization)

Known issue: all collected messages are duplicated (I have no idea why) (EDIT: Fixed)


intelmq/bots/collectors/xmpp/collector.py
```diff
@@ -21,48 +41,87 @@ def init(self):
...
+            self.logger.info("Disconnected from XMPP")
+
+            super(XMPPCollectorBot, self).stop()
+        else:
+            self.logger.info("There was no XMPPClient I could stop.")
+
+    def log_message(self, msg):
+        self.logger.debug("Received Stanza: %r from %r", msg['body'],
+                          msg['from'])
+        self.logger.info("Stanza received")
```
dmth: whoa. After some reconsideration: Let's make this debug. I guess there might be heavy load on the logs once this might get into the wild...

sebix: Right, but as we have a debug line anywhere in the line above, let's just remove it?

I think we should also limit the length of the logged stanza to 500 or so.

dmth:
How does this "duplication" manifest?
Does it only arise when the JSON Parser is used in combination with XMPP?

Is the "base64" Decoding right at this place? I remember of troubles whilst I was creating the Parser in the first place as I couldn't figure out if it's encoded / decoded by serialize/unserialize.

How does the json-string in report["raw"] look like?



Your PR also contains changes on 

intelmq/bots/parsers/danger_rulez/parser.py 
intelmq/lib/message.py
intelmq/lib/test.py 

and a bunch of tests... is this intended?

intelmq/bots/collectors/xmpp/collector.py
dmth: ack

sebix:
> How does this "duplication" manifest?

I posted the message once in the chatroom and I got this:
```
xmpp-collector: Received Stanza: '{"__type": "Event", "classification.type": "unknown", "source.asn": 456, "source.ip": "127.0.0.1"}' from test@conference.jabber.at/sebix
xmpp-collector: Stanza received
xmpp-collector: Sending message.
xmpp-collector: Received Stanza: '{"__type": "Event", "classification.type": "unknown", "source.asn": 456, "source.ip": "127.0.0.1"}' from test@conference.jabber.at/sebix
xmpp-collector: Stanza received
xmpp-collector: Sending message.
```
This does not happen for direct messages.

It's not only logged two times, it's also two times in the queue.

> Does it only arise when the JSON Parser is used in combination with XMPP?

No, always.

> Is the "base64" Decoding right at this place? I remember of troubles whilst I was creating the Parser in the first place as I couldn't figure out if it's encoded / decoded by serialize/unserialize.

Current version handles this correctly.

> How does the json-string in report["raw"] look like?

This is what's logged by the json-parser as incoming: `{'raw': 'eyJfX3R5cGUiOiAiRXZlbnQiLCAiY2xhc3NpZmljYXRpb24udHlwZSI6ICJ1bmtub3duIiwgInNvdXJjZS5hc24iOiA0NTYsICJzb3VyY2UuaXAiOiAiMTI3LjAuMC4xIn0=', 'time.observation': '2016-09-09T07:25:33+00:00'}`

This is exactly this string here `{"__type": "Event", "classification.type": "unknown", "source.asn": 456, "source.ip": "127.0.0.1"}`

> Your PR also contains changes on 
> 
> intelmq/bots/parsers/danger_rulez/parser.py 
> intelmq/lib/message.py
> intelmq/lib/test.py 
> 
> and a bunch of tests... is this intended?

Yes it is. The `Message.__init__(message)` (ans therefore also Report. Event, MessageFactory.unserialize) didn't check the validity of fields in the given `message` dict/tuple. Instead of implementing this in the parser, I did it in the message-class itself. Which is the more correct way to do it I think.

dmth:
Hi, long shot: is it possible that BOTH Event-handlers fire when a groupchat_message is received?

```
self.xmpp.add_event_handler("message", self.log_message)
self.xmpp.add_event_handler("groupchat_message", self.log_message)
```

dmth:
YES! Thats it!.

> First, we register an event handler for the groupchat_message event inside the bot’s __init__ function.
>
> Note
>
> We do not register a handler for the message event in this bot, but if we did, the group chat message would have been sent to both handlers.

http://sleekxmpp.com/getting_started/muc.html

dmth:
So, @sebix and me tested some communication between Instances of IntelMQ via XMPP
```
System A - GER
XMPP Collector -> JSON Parser -> XMPP Output
     /\                                                                  \/
--------------------------------------------------------------------
XMPP Output <- JSON Parser <- XMPP Collector
System B - AT

```

dmth:
Most likely still missing: Validate SSL-Certificates when connecting to XMPP-Server in:

 - [ ] Output Bot
 - [ ] Collector Bot

sebix:
Yep, need to set `xmpp.xmpp.ca_certs` to set the path to the ca-certificates file.

sebix:
There's no `default_context`.

Looking at [sleekxmpp's code](https://github.com/fritzy/SleekXMPP/blob/master/sleekxmpp/xmlstream/xmlstream.py#L524) it seems we really need to pass the path.

[`ssl.get_default_verify_paths()`](https://docs.python.org/3/library/ssl.html?highlight=context#ssl.get_default_verify_paths) gives some paths. `ssl.get_default_verify_paths().openssl_cafile` gives `'/usr/lib/ssl/cert.pem'`, a non-existant file here. `openssl_capath` points to `/usr/lib/ssl/certs` which is a symlink to `/etc/ssl/certs`. On Debian the file is called `ca-certificates.crt`, on centos it's `ca-bundle.crt`.

dmth:
My intention was something like https://docs.python.org/3.4/library/ssl.html#ssl.create_default_context

sebix:
And what to do with the resulting `SSLContext` object?

sebix:
I think this PR is complete now.

PR #676, opened 2016-09-12
ENH: Filterexpert can now use RegEx

dmth:
We enhanced the Filterexpert in order to be capable of using regular expressions.

If a value is set in `filter_regex`, the field  `filter_value` is interpreted as RegEx. The expert will always use pythons `re.search` to filter the values.

If no value is set in `filter_regex`, the common behavior `==` is used.

This change should not be incompatible with existing configurations.


intelmq/tests/bots/experts/filter/test_expert_regex_search.py
```diff
@@ -0,0 +1,35 @@
...
+
+import intelmq.lib.test as test
+from intelmq.bots.experts.filter.expert import FilterExpertBot
+
+EXAMPLE_INPUT = {"__type": "Event",
+                 "classification.type": "defacement",
+                 "time.source": "2005-01-01T00:00:00+00:00",
+                 "time.observation": "2015-09-12T00:00:00+00:00",
+                 "feed.name": "test-feed",
+                 "raw": "fds56gf4jh4jhgh4j6"
```
sebix: This is an invalid base64-value and will fail once #669 is merged.

Issue #668, opened 2016-09-07
Very long BOTS file


sykaeh:
I was just setting up IntelMQ and IntelMQ Manager on our servers and I think it is a great tool, so thank you! As I was adding new nodes I noticed that the list of collectors and parsers is really long. What are the guidelines (if there are any) to adding a new collector to the BOTS file? 

Should there be a new entry for each possible feed (like abuse.ch and blocklist.de)? Or only if there is a separate module for it (i.e. not just using "intelmq.bots.collectors.http.collector_http" with a different URL)?

If the decision is to have a new entry for each feed URL, then I would suggest organizing it in a clearer way (especially in IntelMQ Manager).

dmth:
Hi @sykaeh thank you for bringing up this topic.
My personal preference is the configuration of a "generic" http-url fetcher which allows to copy and paste the URL into it, without the need of creating a seperate entry in the BOTS file.

sykaeh:
That would be my preference as well. My suggestion would be to keep these generic collectors:
 * Fileinput
 * MISP Generic
 * Generic Mail URL Fetcher
 * Generic Mail Attachment Fetcher
 * Generic URL Fetcher **(new)**
 * Request Tracker
 * XMPP collector

And add only the specific collectors that have their own module:
 * AlienVault OTX
 * BitSight Ciberfeed Stream
 * Blueliv Crimeserver
 * N6stomp

I'd be happy to make the changes this is something that is wanted.

In order not to loose all of the feed URLs we could put them in a list in the documentation somewhere, so that everyone can choose themselves whether or not they are relevant for them.

aaronkaplan:
Thanks Sybil for commenting.
I also agree.

We currently know that the config structure right now is sub-optimal.
We were discussing changing that into a bots.d/* kind of config snippet layout.
Similar to /etc/foobar.d/* where  you can have a small config file for each relevant bot.

Playing together with that is the concept of inheritance and having more abstract classes (like HTTP collector) and simply instantiating them with actual parameters. This will make the config much leaner.

However, we are not there yet and struggling to clean up the last bugs which stop us from putting a release 1.0 tag on it.
We will address these config layout issues in release 1.1

That being said, I agree with all your comments below and I'd love to hear good concrete suggestions on how we should structure things in 1.1

Best,
a.

aaronkaplan:
Ha!

Saw your answer just now.

[...]

Issue #644, opened 2016-08-18
Errors when already configured bots gain additional options through upgrade


gsiv:
It appears that when bots receive additional options through an upgrade,
previously configured bots produce errors because of those new, unconfigured
options.

At least, this is what happened after the mail collector bots gained the two
new chunking options:

```
2016-08-05 17:38:13,111 - Shadowserver-Open-SNMP - INFO - Reading email report
2016-08-05 17:38:13,112 - Shadowserver-Open-SNMP - INFO - Downloading report from https://[…]
2016-08-05 17:38:15,883 - Shadowserver-Open-SNMP - INFO - Report downloaded.
2016-08-05 17:38:15,884 - Shadowserver-Open-SNMP - ERROR - Bot has found a problem.
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/intelmq/lib/bot.py", line 121, in start
    self.process()
  File "/usr/lib/python3/dist-packages/intelmq/bots/collectors/mail/collector_mail_url.py", line 89, in process
    for report in generate_reports(template, io.BytesIO(resp.content), self.parameters.chunk_size,
AttributeError: 'Parameters' object has no attribute 'chunk_size'
2016-08-05 17:38:15,884 - Shadowserver-Open-SNMP - INFO - Current Message(event): None.
2016-08-05 17:38:15,885 - Shadowserver-Open-SNMP - INFO - Bot will restart in 15 seconds.
2016-08-05 17:38:18,144 - Shadowserver-Open-SNMP - ERROR - Received KeyboardInterrupt.
2016-08-05 17:38:18,145 - Shadowserver-Open-SNMP - INFO - Bot stopped.
2016-08-05 17:38:18,145 - Shadowserver-Open-SNMP - ERROR - Exiting.
2016-08-05 18:02:27,511 - Shadowserver-Open-SNMP - INFO - MailURLCollectorBot initialized with id BSI-Shadowse
ver-Open-SNMP and version 3.4.3 (default, Oct 14 2015, 20:28:29) as process 11604.
2016-08-05 18:02:27,512 - Shadowserver-Open-SNMP - INFO - Bot is starting.
2016-08-05 18:02:27,515 - Shadowserver-Open-SNMP - INFO - Bot starts processings.
2016-08-05 18:02:27,516 - Shadowserver-Open-SNMP - INFO - Pipeline ready.
2016-08-05 18:02:27,516 - Shadowserver-Open-SNMP - INFO - Start processing.
[...]

Issue #552, opened 2016-06-30
Enable separate packaging of bots by allowing addition and removals to the config.


bernhardreiter:
The problem:
Some bots have different requirements from the core of intelmq. Example: a contact bot with postgresql.
It makes sense to enable separate packaging of bots. With current intelmq (26d112873381f91cb6c6fedb5ddef2f32b4bf46d) there is a file https://github.com/certtools/intelmq/blob/master/intelmq/bots/BOTS
which holds basic info about each bot.

Adding or removing a bot would need to add or remove a section in this file.
As discussed in April 2016 on http://lists.cert.at/pipermail/intelmq-dev/2016-April/thread.html
people agree that this situation should be improved with a number of suggestions.

Previous related issues are #368 and #267 
(I lack the permission rights to rename and refocus  #368, so I am opening this one)

aaronkaplan:
+1

Issue #368, opened 2015-10-08
Where bot configs should be stored in repo?


SYNchroACK:

sebix:
I assume this question arose as the docs are now centralized in a separate combined file?

SYNchroACK:
The following information represents an example of what we discussed

**Modify Expert Bot**
* * *
**Files location in repository:**

* bot documentation: intelmq/bots/experts/modify/modify.md
* bot config: intelmq/bots/experts/modify/modify.conf

**Files location in installation folder:**

* bot documentation: /opt/intelmq/docs/bots/modify.md
* bot config: /opt/intelmq/conf/bots/modify.conf


Is it ok?

sebix:
Yes, I'm fine with it.

bernhardreiter:
For  completeness: I also believe that the documentation for a bot should be close to the code (and the default config, and the test code for this bot), generally a bot directory should contain everything.

A technical remark: an alternative to  ` intelmq/bots/experts/modify/modify.md` is `intelmq/bots/experts/modify/README.md` making it easier to collect the documentation.

sebix:
On 09/27/2016 08:36 AM, Bernhard E. Reiter wrote:
>
> For completeness: I also believe that the documentation for a bot
> should be close to the code (and the default config, and the test code
> for this bot), generally a bot directory should contain everything.
>
Also makes a separation of bots easier.
>
> A technical remark: an alternative to
> |intelmq/bots/experts/modify/modify.md| is
> |intelmq/bots/experts/modify/README.md| making it easier to collect
> the documentation.
>
ACK

PR #706, opened 2016-09-22
Fixes certtools/intelmq#668.

sykaeh:
Reorganization of BOTS file and more documentation

### DIFF
--- a/intelmq/bots/BOTS
+++ b/intelmq/bots/BOTS
@@ -23,7 +23,7 @@
                 "mail_password": "<password>",
                 "mail_ssl": true,
                 "mail_user": "<user>",
-                "rate_limit": "60",
+                "rate_limit": 60,
                 "subject_regex": "<subject>"
             }
         },
@@ -39,7 +39,7 @@
                 "mail_password": "<password>",
                 "mail_ssl": true,
                 "mail_user": "<user>",
-                "rate_limit": "60",
+                "rate_limit": 60,
                 "subject_regex": "<subject>",
                 "url_regex": "http://"
             }
@@ -66,7 +66,6 @@
                 "rate_limit": 3600
             }
         },
-
         "Request Tracker": {
             "description": "Request Tracker Collector fetches attachments from an RTIR instance and optionally decrypts them with gnupg.",
             "module": "intelmq.bots.collectors.rt.collector_rt",
@@ -110,7 +109,7 @@
             "parameters": {
                 "api_key": "<insert your api key>",
                 "feed": "AlienVault OTX",
-                "rate_limit": "3600"
+                "rate_limit": 3600
             }
         },
         "BitSight Ciberfeed Stream": {
@@ -127,7 +126,7 @@
             "parameters": {
                 "api_key": "<insert your api key>",
                 "feed": "Blueliv Crimeserver",
-                "rate_limit": "3600"
+                "rate_limit": 3600
             }
         },
         "N6stomp": {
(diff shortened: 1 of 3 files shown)


sykaeh:
The build failed because the order of the collectors is not alphabetical in the BOTS failed. I changed to be generic collectors and then specific collectors (as discussed in #668). Should I change it to be alphabetical again? Or do we want to change the test?

dmth:
Thanks for your work!
From my point of view the test is not useful. On the contrary: It prohibits a user freindly ordering of fields, for instance: First: Username, Second: Password

sykaeh:
Yes, I agree that the ordering can be done more user friendly when it is not alphabetical. However, I do think that a test to check whether the BOTS file is valid JSON should be done. 

Are there any other checks that would be useful? Possibly a check to see whether there is a section for all of the bot types (i.e. collectors, parsers, outputs, experts)?

intelmq/bots/BOTS
```diff
@@ -653,6 +103,45 @@
                 "xmpp_server": "xmpp server",
                 "xmpp_user": "xmpp username"
             }
+        },
+        "AlienVault OTX": {
```
sebix: Why has this been moved from the top to the bottom? This list should be sorted (also see bin/rewrite_config_files.py)
EDIT: Haven't seen the discussion yet.

intelmq/bots/BOTS
```diff
@@ -653,6 +103,45 @@
...
+                "http_url": "http://alerts.bitsighttech.com:8080/stream?key=<api>"
+            }
+        },
+        "Blueliv Crimeserver": {
+            "description": "Blueliv Crimeserver Collector is the bot responsible to get the report through the API.",
+            "module": "intelmq.bots.collectors.blueliv.collector_crimeserver",
+            "parameters": {
+                "api_key": "<insert your api key>",
+                "feed": "Blueliv Crimeserver",
+                "rate_limit": "3600"
```
sebix: This should be an integer, not string.

intelmq/bots/BOTS
```diff
@@ -653,6 +103,45 @@
...
                 "xmpp_user": "xmpp username"
             }
+        },
+        "AlienVault OTX": {
+            "description": "AlienVault OTX Collector is the bot responsible to get the report through the API. Report could vary according to subscriptions.",
+            "module": "intelmq.bots.collectors.alienvault_otx.collector",
+            "parameters": {
+                "api_key": "<insert your api key>",
+                "feed": "AlienVault OTX",
+                "rate_limit": "3600"
```
sebix: integer please

intelmq/bots/BOTS
sykaeh: I was not aware that `bin/rewrite_config_files.py` existed. Do you still think that it should be ordered alphabetically rather than generic and specific?

intelmq/bots/BOTS
sykaeh: This was like this before already. I only moved it, but of course I will change it.

intelmq/bots/BOTS
sykaeh: This was like this before already. I only moved it, but of course I will change it.

sykaeh:
> A check if the module actually exists and the the description exists and is not empty. For collectors: Parameter feed exists.

Where exactly is the description used? Most of the parsers have the same nondescript text anyways: "<Parser name> is the bot responsible to parse the report and sanitize the information." Wouldn't it make more sense to move the description of the bots to the documentation instead of this file (like the docs/Bots.md file)?

sebix:
> Where exactly is the description used? Most of the parsers have the same nondescript text anyways: " is the bot responsible to parse the report and sanitize the information."

Yes, it's in mostly pretty useless unfortunately.

> Wouldn't it make more sense to move the description of the bots to the documentation instead of this file (like the docs/Bots.md file)?

The docs/Bots.md file is also not a good place. We had the idea to save the following data per bot into a readme or as docstring:
* parameters+descriptions+default value
* description (i.e. readme)

Then we can always generate a combined documentation (sphinx et al.) and the bots always have their default values and they can easily be upgraded.

Also see #368 and #644 

I will have time to address these issues as recently as in October.

intelmq/bots/BOTS
sebix: > Do you still think that it should be ordered alphabetically rather than generic and specific?

The configuration has a lot of shortcomings, this PR addresses one of them. It does not solve the grouping/sorting ultimately, but it's better than before.

So, yes, I think we can drop the sorting now.

Please more opinions @certtools/intelmq-contributors concerning this change (not so much the technical questions, but the change in general).

aaronkaplan:
---
Mobile

> On 22.09.2016, at 20:02, Sybil Ehrensberger <notifications@github.com> wrote:
> 
> A check if the module actually exists and the the description exists and is not empty. For collectors: Parameter feed exists.
> 
> Where exactly is the description used? Most of the parsers have the same nondescript text anyways: " is the bot responsible to parse the report and sanitize the information." Wouldn't it make more sense to move the description of the bots to the documentation instead of this file (like the docs/Bots.md file)?
> 

+1

aaronkaplan:
---
Mobile

> On 22.09.2016, at 17:15, Sybil Ehrensberger <notifications@github.com> wrote:
> 
> Yes, I agree that the ordering can be done more user friendly when it is not alphabetical. However, I do think that a test to check whether the BOTS file is valid JSON should be done.
> 
> Are there any other checks that would be useful? Possibly a check to see whether there is a section for all of the bot types (i.e. collectors, parsers, outputs, experts)?
> 
+ Valid JSON syntax check

sebix:
> + Valid JSON syntax check
Once we parse it to check anything (indentation, existence of fields),
we always check the syntax implicitly.

aaronkaplan:
---
Mobile

> On 23.09.2016, at 11:20, Sebastian <notifications@github.com> wrote:
> 
> 
> > + Valid JSON syntax check
> Once we parse it to check anything (indentation, existence of fields),
> we always check the syntax implicitly.
> 

Excellent.

bernhardreiter:
@sebix, you wrote:

> The docs/Bots.md file is also not a good place. We had the idea to save the following data per bot [..]

If we separate the documentation, we should also separate the rest of the bot infos and complete #552. However this may be too much for the 1.0 release and you did already tag #552 for 1.1, so may suggestion for 1.0 and this issue at hand is: put documentation either into docs/Bots.md (to be moved later when everything is moved with #552) or link stuff from docs/Bots.md to the README.md in the directory (I think I will do so for for two files while fixing #711).

intelmq/bots/BOTS
sykaeh: Any ideas on how to drop the sorting from the test while still checking for indentation and other style specifications?

bernhardreiter: @sykaeh to drop the sorting you have change test_BOTS_syntax() in intelmq/tests/test_conf.py . From reading the code it seems that you would need to replace the call to the custom to_json() function with an explicit call that does not sort.

sykaeh: Yes, but if you do not sort, then Python outputs the content in an arbitrary order (which is not the same as the input file) and then you cannot compare it to the original file.

dmth: I think @bernhardreiter might want to go this way:

```
import collections
...

def to_unsorted_json(obj):
    """
    Transforms object into JSON with intelmq-style.
    """
    return json.dumps(obj, indent=4, sort_keys=False,
                      separators=(',', ': ')) + '\n'

...
interpreted = json.loads(fcontent, object_pairs_hook=collections.OrderedDict)
self.assertEqual(to_unsorted_json(interpreted), fcontent)
```

PR #718, opened 2016-10-04
Introduction of the Enabled-Switch for Collectors.

### LINKED ISSUE #715: [3] "Deactivated Switch" for Collectors.
Since Bots were split into different classes, it's possible to enhance certain types of bots. That is a very good development.

I'd like to add a "Deactivated Switch" to the CollectorBot, in order to enable the deactivation of Collectors. When a bot is starting / restarted it checks if it was deactivated, and automatically stops if this is the case.

This is useful for a temporary deactivation of collectors , for instance when feeds are providing erroneous data. Restarting the whole bot-net would not start such bots, once a "deactivated switch" is in place.

dmth:
This introduces the enabled-switch for all Collector-Bots.
By doing so it fixes #715.

The Collectorbot was extended with an own start-function
which checks if the enabled parameter is present and was set
to a representation of False. To check the "representation of
False" the function to_bool was introduced in utils.py, which
maps well-known-text to booleans. It has a a fallbackmode parameter
which determines if the function returns true of false if no
mapping  was found.

As long as to_bool does not evaluate to False the bot will be started.

### DIFF
--- a/docs/Bots.md
+++ b/docs/Bots.md
@@ -9,6 +9,10 @@
 <a name="collectors"></a>
 ## Collectors
 
+* All Collector-Bots can be extended with an `enabled` parameter in `runtime.conf`. If the parameter is set to `True`
+  (which is assumed as defaut if it is missing) the bot will start. If the parameter was set to `False`, the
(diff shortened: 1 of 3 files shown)


sebix:
I pushed a generalized version to master.

Issue #267, opened 2015-07-10
Configurations - Hierarchy configurations


SYNchroACK:
- generic
- bot type
- bots

sebix:
Could you please expand the description of this issue? It is totally unclear what is meant here.

SYNchroACK:
The idea is remove the defaults.conf file and have the runtime.conf in the following format:

**Current config:**
```
{
    "malware-domain-list-collector": {
        "error_procedure": "retry",
        "error_retry_delay": 30,
        "error_log_message": true,
        "error_dump_message": true,
        "url": "http://www.malwaredomainlist.com/updatescsv.php",
        "feed": "Malware Domain List",
        "rate_limit": 3600
    },
    "malware-domain-list-parser": {
        "error_procedure": "pass",
        "error_retry_delay": 30,
        "error_log_message": true,
        "error_dump_message": true
    },
    "file-output": {
        "error_procedure": "retry",
        "error_retry_delay": 30,
        "error_log_message": true,
        "error_dump_message": true,
        "file": "/opt/intelmq/var/lib/bots/file-output/events.txt"
    }
}
```

**Proposal:**
```
{
    "__default__": {
        "broker": "redis",
        "error_procedure": "retry",
        "error_max_retries": 3,
        "error_log_message": true,
        "error_log_exception": true,
        "error_dump_message": true,
        "error_retry_delay": 15,
        "exit_on_stop": true,
        "rate_limit": 0,
        "load_balance": false,
        "source_pipeline_host": "127.0.0.1",
        "source_pipeline_port": "6379",
        "source_pipeline_db": "2",
        "destination_pipeline_host": "127.0.0.1",
        "destination_pipeline_port": "6379",
        "destination_pipeline_db": "2",
        "http_proxy": null,
        "https_proxy": null,
        "http_user_agent": null
    },
    "Collectors": {
        "__default__": {
            "error_procedure": "pass",
            "error_retry_delay": 1000,
            "error_log_message": false,
            "error_dump_message": false,
        },
        "malware-domain-list-collector": {
            "error_procedure": "retry",
            "error_retry_delay": 30,
            "error_log_message": true,
            "error_dump_message": true,
            "url": "http://www.malwaredomainlist.com/updatescsv.php",
            "feed": "Malware Domain List",
            "rate_limit": 3600
        }
    },
    "Parsers": {
        "__default__": {
            "error_procedure": "retry",
            "error_retry_delay": 2323,
            "error_log_message": false,
[...]

PR #719, opened 2016-10-05
Remove startup.conf

sebix:
next step of certtools/intelmq#267

It's still supported, but will be dropped after the feature freeze.


dmth:
The file should also be removed from "examples"

sebix:
See certtools/intelmq-manager#85 for the changes in the manager.

dmth:
hi, @sebix what's the state of this PR? From my understanding it's still needed.

sebix:
I will merge it this week tough there are not reviews.

sebix:
Merged.

PR #724, opened 2016-10-06
deduplicator: Improve speed and use full sha256 hash

bernhard-herzog:


dmth:
**Notes for Migration** After applying this PR, the deduplication of events will not work properly for a short time, as two different kinds of hashing are used.  This will lead to duplicated events.

bernhard-herzog:
We could add an optional compatibility mode: For each event, compute the old hash and check whether that is present in the cache and discard the event if it is.

The old hash can easily be computed from the new one: `hash(int(newhash, 16))`

sebix:
In dcdf799f73a737647727ff04a873daf2c095b8e0 I changed the deduplicator bot to be backwards compatible by computing the old short hashes. Was easier than I initially thought.

intelmq/lib/message.py
```diff
@@ -221,18 +221,27 @@ def __get_type_config(self, key):
         return class_name
 
     def __hash__(self):
+        return int(self.hash(), 16)
```
wagner-certat: Shouldn't this be just `self.hash()`?

bernhard-herzog: The `hash`  method returns a string, but `__hash__` has to return an int. This conversion from string to int is the only part left in `__hash__` from its original version.

Issue #574, opened 2016-07-19
dump files: locking mechanism


sebix:
#573 emits a warning in intelmqdump if the bot is running. But this is not enough to mitigate race-conditions, as a bot could be started during recovering. Both bot and intelmqdump have to lock the file before working on it.

https://pypi.python.org/pypi/filelock/

sebix:
Other possibilities:

 - One line per dump (in JSON-format), allows easy appending
 - an error pipeline queue
 - one file per dump

wagner-certat:
> One line per dump (in JSON-format), allows easy appending

Does not help either. Locking file descriptors can be done with: https://docs.python.org/3.5/library/fcntl.html#fcntl.flock

wagner-certat:
Rewriting intelmqdump is something for the next feature release

wagner-certat:
fixed in 1b18e23e41fe63251034af0573ec596f8e645d9d

PR #730, opened 2016-10-11
intelmqdump improvements

sebix:
- Handle EOF for user input
- Hide the possibly very long list after show command
- deny recovery for running bots (see also certtools/intelmq#574)
- handle missing ids for `r` command
- boost recovery performance by saving files only once at the end


sebix:
I added a commit which allows a safe abort of the program during recovering or in case of errors (e.g. redis has problems because of the huge amount of data). Was very useful in our setup.

PR #740, opened 2016-10-13
Generic Database lookup expert

sebix:
This PR is based on the commits in PR #739 

It extends the contact db lookup bot by providing a more generic interface. It's able to do lookups on multiple columns and can replace multiple values.


sebix:
merged in e3656e5

PR #745, opened 2016-10-17
Modify expert: convert dict-based structure to list-based

sebix:
Fixes the unpredictable ordering, discovered by @dmth a while ago (I haven't found an issue for it)


dmth:
Clarification: I didn't discover this. I think it was @bernhard-herzog, @bernhardreiter or @swilde

dmth:
Nevertheless you assigned this PR to me: What shall I do with this PR?

dmth:
Ok, had a short chat with @sebix next todos are:
- [ ] Review this PR. 
- [ ] Check Documentation

bernhardreiter:
There is no documentation in the diff as far as I can see. I suggest to rereview, once
documentation is there. Overall I believe the old implementation should be kept until #662
is solved, otherwise you are creating another variant, which does not help much.

sebix:
Documentation updated.

> Overall I believe the old implementation should be kept until #662
is solved, otherwise you are creating another variant, which does not help much.

I consider the non-deterministic behavior a bug, while the configuration format is a feature-request. A fix for first is a must-have, the second is optional.

dmth:
I don't share the opinion to keep the old implementation. I consider the new Configuration much easier to read and write. 

The documentation should contain a more clear note, that the the actions of rules are applied in order of their appearance in the config file, except for the actions in the `__default` rule, which is only applied if no other rule was chosen.

Did you convert the config files by hand or with a script? Do you mind to share this script in order to enable a much easier migration of existing configurations?

By the look of it, I did not run / test the code, it's ok.
In order to apply this in our production, we need to migrate our configuration, though.

sebix:
On 10/20/2016 09:36 AM, Dustin Demuth wrote:
>
> I don't share the opinion to keep the old implementation. I consider
> the new Configuration much easier to read and write.
>
We could also write the converted configuration to the file. Thus we
have a clean upgrade path.

bernhardreiter:
> I don't share the opinion to keep the old implementation. I consider the new Configuration much easier to read and write.

Because of the pending change to an even easier configuration with #662 you are putting work into a dead horse. And you go against @aaronkaplan plan to not have an easier syntax that keeps json.

bernhardreiter:
> We could also write the converted configuration to the file. Thus we
> have a clean upgrade path.

You are changing the behaviour of the configuration. Ff the original configuration was correct,
the rules did not have side effects otherwise it was broken. In that case your conversion script tries to interpret the intentions of the author of the original configuration. Not a clean solution in my opionion.

dmth:
Whilst discussing this, @bernhardreiter mentioned the question, if grouping and the `__default` rule  are really needed. Keeping them means to keep a high level of complexity. He suggested to remove them in https://github.com/certtools/intelmq/issues/662#issuecomment-244015676 


My opinion on the groups is that those improve readability and the `__default` condition is can be used in order to not repeat oneself.

So the configuration is perceived in very different ways.

Nevertheless: @sebix Do you know of use-cases where the grouping is required?

sebix:
I will discuss it with @aaronkaplan.

Concerning the grouping: I kept it as nobody objected earlier in the discussion.

sebix:
I now removed the grouping feature.

dmth:
@sebix this PR would be linked with the morefeeds config from https://github.com/certtools/intelmq/pull/764

If this code is merged, the file from https://github.com/certtools/intelmq/pull/764 would also require an overhaul. OR this PR also contains the lowercase writing of the `classification.identifier`

docs/Bots.md
```diff
@@ -305,62 +305,68 @@ The modify expert bot allows you to change arbitrary field values of events just
...
-    "conficker": [{
-            "malware.name": "^conficker(ab)?$"
-        }, {
+[
+    {
+        "rule": "Standard Protocols http",
+        "action": {
+            "protocol.application": "http"
+        },
+        "selection": {
```
aaronkaplan: let's call this "if-regexp" or just "if"

docs/Bots.md
```diff
@@ -305,62 +305,68 @@ The modify expert bot allows you to change arbitrary field values of events just
...
-            "feed.name": "^Spamhaus Cert$"
-        }, {
-            "classification.identifier": "{msg[malware.name]}"
-        }],
-    "conficker": [{
-            "malware.name": "^conficker(ab)?$"
-        }, {
+[
+    {
+        "rule": "Standard Protocols http",
```
aaronkaplan: let's call this "rule-name" or "rule name".
It's important for the reader to know that this might be ignored.

docs/Bots.md
```diff
@@ -305,62 +305,68 @@ The modify expert bot allows you to change arbitrary field values of events just
...
+        "action": {
```
aaronkaplan: Let's call this "then"

docs/Bots.md
```diff
@@ -305,62 +305,68 @@ The modify expert bot allows you to change arbitrary field values of events just
...
-The dictionary on the first level holds sections to group the rules.
-In our example above we have two sections labeled `Spamhaus Cert` and `Standard Protocols`.
-All sections will be considered, but in undefined order.
-
-Each section holds a dictionary of rules, consisting of *conditions* and *actions*.
-`__default` indicates an optional default rule. If a default rule exist, the section
-will only be entered, if its conditions match. Actions are optional for the default rule.
+In our example above we have five groups labeled `Standard Protocols http`,
+`Spamhaus Cert conficker`, `bitdefender`, `urlzone` and `default`.
+All sections will be considered, in the given order.
```
aaronkaplan: might want to add ... "given order (from top to bottom)"

intelmq/bots/experts/modify/examples/default.conf
```diff
@@ -1,154 +1,256 @@
...
-        }, {
-            "classification.identifier": "conficker"
-        }],
-    "urlzone": [{
-            "malware.name": "^urlzone2?$"
-        }, {
+[
+    {
+        "rule": "Abuse.ch zeus",
+        "action": {
```
aaronkaplan: need to adapt all the key names here of course.

intelmq/bots/experts/modify/expert.py
```diff
@@ -23,16 +23,25 @@ def __getitem__(self, key):
         return self.match.group(key)
 
 
+def convert_config(old):
+    config = []
+    for groupname, group in old.items():
+        for rule_name, rule in group.items():
+            config.append({"rule": groupname + ' ' + rule_name,
+                           "selection": rule[0],
```
aaronkaplan: need to adapt the names here

sebix:
Merged conflicts and renamed the keys

ping @dmth and @aaronkaplan

sebix:
Ping again at @dmth and @aaronkaplan

dmth:
From my POV that's ok. We need to make sure that this PR is on our Changelogs.

wagner-certat:
please review @aaronkaplan 

This PR now waits since nearly 3 months.

PR #735, opened 2016-10-11
Shadowserver config fixups

aaronkaplan:
Partially fixes #733

PR #748, opened 2016-10-18
Shadowserver new configs

cert-lv:
Added Shadowserver's configs for:
- Compromised-Website
- Open-Netis
- NTP-Version
- Sandbox-URL
- Spam-URL
- Vulnerable-ISAKMP
- Botnet-CCIP
- Accessible-RDP
- Open-LDAP
- Blacklisted-IP
- Accessible-Telnet
- Accessible-CWMP

And edited config for **Open-mDNS**.

We do not use a field **classification.taxonomy** because it's automatically added by **Taxonomy** expert-bot later.

P.S. sorry for 6 commits, some of them was for reverting changes back.

Issue #725, opened 2016-10-07
Extension of the DHO by a Field `feed.provider`


dmth:
Currently the DHO can only reflect the `Name`, `Url` and `Code` of a feed, it is not possible to store the provider / feeder of an information.

We think its necessary to extend the DHO by such a field, an also enhance the documentation how these fields are used.

For instance:

`feed.provider` : shadowserver <- Set by Parser
`feed.name`: Sinkhole-HTTP-DRONE <- Set by Parser
`feed.code`: MyTrustableSource <- Set by user in Runtime.conf
`feed.url`: url://to.feed.description <- Set by Parser

sykaeh:
I think having a field `feed.provider` is a good idea, but shouldn't the `feed.provider`, `feed.name` and `feed.url` be set in the collector?

dmth:
In the collector they have to be fields which are configurable by the user. As you can use collectors for different feeds. If those parameters are set in the parser, the user configuration is not necessary (at least for the specialised parsers like shadowserver), because these fields should be known by parser or its specific configuration.

One idea is that the field `feed.provider` is also available in the collector, but the parser would ensure, that the field is populated if no value was set in the collector.

The values of some parameters like `feed.name` are evaluated (at least in our use case) in later steps, for instance when generating e-mails, we rely on the fact that these values can always be determined. If they become a user-configurable parameter, it might happen, that the event cannot be processed at the end.

sebix:
I think it boils down to: Can one collector collect only one feed or multiple?

sykaeh:
It's actually both whether the collector and the parser can collect/parse one feed or multiple. 

For example in the Shadowserver case, it would be easy to use one collector to collect all feeds, but you could also use one collector per feed type. The same is true for the parser. You could use one Shadowserver parser to parse all of the feeds (assuming you know what the `feed.name` field is already filled out) or you could have a separate parser for each feed type.

It would be possible to have one collector per Shadowserver feed and then only one parser that relies on the `feed.name` to determine how to parse the feed. However, the opposite is not possible (one collector, many parsers).

IMHO, the information about the feed belongs in the collector, as that is where you are actually retrieving the feed. The parser only knows what the feed is based on the current configuration (or possibly based on values in the feed itself).

PR #752, opened 2016-10-20
Shadowserver config fixups

sebix:
Rebased from #735


dmth:
Thanks for giving this a fresh start.

Questions so far:
 1. Do you want to stick to the approach, that no `feed.name` is set by the parser? Have we agreed that this should always be done by a collector as a user-configurable parameter? See https://github.com/certtools/intelmq/issues/725 for a start of this discussion
 2. Do we want to keep the `feedname` <-> `feed.name` ambiguity or can we resolve this somehow? For instance by resolving `feed.name` from the report send by the collector if no explicit `feedname` is present (see line https://github.com/certtools/intelmq/blob/56d18d31b27185cd247a85addb907af7581f048f/intelmq/bots/parsers/shadowserver/parser.py#L31 where the correct parser configuration is discovered by using this parameter)
 3. The Overwrite / Override Change is not applied to this PR, might be reasonable to do.

dmth:
@sebix I've no objections for this PR, I think we can deal with the alterations which happened here.

I want to get this one going, in order to be able of merging the other parser configurations contributed  in https://github.com/certtools/intelmq/pull/748

The part where `feed.code` is set should be removed now.
https://github.com/certtools/intelmq/blob/master/intelmq/bots/parsers/shadowserver/parser.py#L162


What Intevation needs to do for their setup to keep this working afterwards:
 - Ensure Override / Overwrite is set to "True" in order to set the correct `feed.name` by the parser to `self.parameters.feedname`

dmth:
@sebix @aaronkaplan When will this PR be merged?

sebix:
> Do you want to stick to the approach, that no feed.name is set by the parser? Have we agreed that this should always be done by a collector as a user-configurable parameter? See #725 for a start of this discussion

> Do we want to keep the feedname <-> feed.name ambiguity or can we resolve this somehow? For instance by resolving feed.name from the report send by the collector if no explicit feedname is present (see line https://github.com/certtools/intelmq/blob/56d18d31b27185cd247a85addb907af7581f048f/intelmq/bots/parsers/shadowserver/parser.py#L31 where the correct parser configuration is discovered by using this parameter)

I think @aaronkaplan can best answer this.

The overwrite/override PR has been merged just now.

dmth:
Hi,

The questions I mentioned above should not stop you from merging this PR.

Nevertheless
> The part where feed.code is set should be removed now.
> https://github.com/certtools/intelmq/blob/master/intelmq/bots/parsers/shadowserver/parser.py#L162

might need another look. The code is superfluos right now.

sykaeh:
I would really appreciate it if this PR was merged soon as well. There seem to be a lot of open PR with regards to the Shadowserver parser.

1. A compromise might be to set the `feed.name` if there currently is no `feed.name`. 

2. I think using the `feed.name` from the report if no feedname parameter is specified is a great idea. That why every organization can decide themselves whether they want to specify the feed name in the collector or the parser. However this would mean moving the selection of the correct config from the `init()` block to the `parse()` block.

dmth:
@sebix Thanks for merging this!

> However this would mean moving the selection of the correct config from the init() block to the parse() block. 

If we can assumpt, that every report contains the same feed.name and it does not change between reports, we only need to lookup the config once.
It might be tested if the `feed.name` from the report, differs from the `feed.name` that was received from the previous report. If yes: Load a new Config, if no: Stick to the previos config.  This would enable ONE Parser for different collectors.

I think we can be capable of providing both approaches.

Issue #733, opened 2016-10-11
feed.* naming: clean up after PR #648


aaronkaplan:
#648 added a "feed.name" to the config.py. However, as it turned out with a discussion between @dmth and me, we actually had two different concepts in mind:
- feeder (i.e. the source of the data feed) - let's call it now "feed.provider"
  and
- one or multiple feeds ("feed.name") that a "feed.provider" might provide.

We need to separate these concepts.

Action items:
- [x] revert back the addition of "feed.name" to config.py
- [ ] add a "feed.provider" to the DHO
- [ ] clarify DHO description
- [ ] add a "feed.provider" to the config.py  to the collector
- [ ] adapt test cases

sykaeh:
A separate issue for `feed.provider` has been opened, see #725.

PR #759, opened 2016-10-27
Add "feed.provider" to the DHO

sykaeh:
This implements #725 (also mentioned in #733) and includes some typo fixes.

This only adds `feed.provider` to the DHO and implements adding it to the report if the parameter `provider` is set in a collector (analogous to `feed.code`). Whether this value is set in the parsers is still open for discussion.

### DIFF
--- a/docs/Harmonization-fields.md
+++ b/docs/Harmonization-fields.md
@@ -14,7 +14,7 @@ Harmonization field names
 |Destination|destination.as_name|[String](#string)|The autonomous system name to which the connection headed.|
 |Destination|destination.asn|[Integer](#integer)|The autonomous system number from which originated the connection.|
 |Destination|destination.fqdn|[FQDN](#fqdn)|A DNS name related to the host from which the connection originated. DNS allows even binary data in DNS, so we have to allow everything. A final point is stripped, string is converted to lower case characters.|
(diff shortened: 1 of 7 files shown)


sykaeh:
Does anyone understand why the cymru whois expert bot test failed on Python3.4?

dmth:
Thank you for implementing this! 

> Does anyone understand why the cymru whois expert bot test failed on Python3.4?

AFAIK the tests make a real lookup at team cymru. If the response differs, the test fails. I don't know why especially the 3.4 Test fails. Maybe there was a short Hickup?

aaronkaplan:
Hi @sykaeh ,

thanks for your submission.
It seems the cymru whois unit test simply failed since something changed in the output format there:
```
-  'destination.as_name': 'Local Datacenter Soluções em Comunicação Ltda.,,, BR',
?                                                                         --
+  'destination.as_name': 'Local Datacenter Soluções em Comunicação Ltda., BR',
```

So, I'll fix that test and we'll be fine again and re-run the tests.
Thanks
a

sykaeh:
Previously only one of the three Python versions failed. It seems like Cymru is not always returning the same value. That's why I had changed it. I now changed it back. We'll see if it works.

aaronkaplan:
@sykaeh okay, got it. Maybe we then have to re-write the test. Anyway. Thanks! It seems to have done the trick for now. Merging this PR.

Issue #693, opened 2016-09-17
[13] Use init systems to manage bots and botnet


sebix:

ondj:
I think intelmqctl should be startup system independent because, for example, I use intelmqctl for starting and stopping bots running under supervisord in Docker container.

sebix:
Sure, providing a configuration for supervisord is no problem. But as default, I would not use supervisor, as it's not packaged on lts distros.

sebix:
Moving this to 1.0, see https://github.com/certtools/intelmq/pull/763#issuecomment-260991676

sebix:
To create/update the systemd/supervisord config we could implement the functionality in intelmqctl e.g. `intelmqctl create-scripts` which will then "convert" the runtime config into systemd services/supervisord config (depending on configuration).

However, this needs to be called as root, as `/lib/systemd/system`. Or we place them under a intelmq-writeable directory and use the full path when calling `systemctl`.

ondj:
And what about to use [systemd generator](https://www.freedesktop.org/software/systemd/man/systemd.generator.html)?

But maybe it is not suitable, because:

> Generators are run very early at boot and cannot rely on any external services. They may not talk to any other process. That includes simple things such as logging to syslog(3), or systemd itself (this means: no systemctl(1))! Non-essential file systems like /var and /home are mounted after generators have run. Generators can however rely on the most basic kernel functionality to be available, including a mounted /sys, /proc, /dev, /usr.

wagner-certat:
Thanks for the pointer!

We definitely need `/etc` to load the runtime.

Also important:

> If you are careful, you can implement generators in shell scripts. We do recommend C code however, since generators are executed synchronously and hence delay the entire boot if they are slow.

wagner-certat:
@certtools/intelmq-contributors 
I think we should postpone this to a later release. The current base is very stable and we postponed the 1.0 release already for too long. As introducing such a change right before beta/RC-releases means a lot of discussions, work and rash decisions, I'd rather not do this now.

I changed my opinion in this regard, this change is not *necessary* for the next stable release. Big changes in the process management implicates some months of intense discussions, implementation and testing.

wagner-certat:
Solved by #1360 and #953

PR #763, opened 2016-11-08
Write pidfiles by bots

### LINKED ISSUE #710: [3] pidfiles should be written by bots, not intelmqctl
Currently, the intelmqctl tool does write and delete the pidfiles, this should be done by the bots themselves. 

As it is now, pidfiles of died bots are deleted when ctl checks the bot's status the next time. They should be removed when the bots exits instead.

sebix:
Fixes #710


ondj:
I think this is not a good solution because it prevents to use a different method for managing bot processes, like by systemd or supervisord, which handles PID files by own.

sebix:
Moving the pidfile handling from the ctl to the bots is a bugfix. Otherwise you may end up with multiple instances of the same bot, see #710 and certat/intelmq#92

By no means I would prefer dropping pidfiles at all, the concept is flawed. See #693

aaronkaplan:
After discussing with @sebix I believe it is much more elegant to have this done properly with systemd.
Can you please:
  1. check how this works with systemd
  2. provide a sample config for systemd
  3. update the documentation
  4. get rid of the pid file code all together.

PR #765, opened 2016-11-09
Bots reloading harmonization correctly now

sebix:
Needed a change in way how we initialized events, but it's more direct now.


aaronkaplan:
looks ok. thx!

Issue #737, opened 2016-10-12
Add client side certificate to HTTP collector


sykaeh:
To be able to use the generic HTTP collector with the N6 HTTP feeds, we need to be able to specify the client side certificate. Currently there is no parameter for that. The `requests` library supports two methods:

> You can also specify a local cert to use as client side certificate, as a single file (containing the private key and the certificate) or as a tuple of both file's path.

For our use case a single file is enough, but maybe we should allow the tuple as well? 

It also begs the question if we should refactor the collectors that make HTTP requests so that there is less duplicated code and more consistent support for the various options (basic authentication, proxies, headers, SSL cert verification). Currently the following collectors make HTTP requests:
- HTTPCollectorBot
- HTTPStreamCollectorBot
- MailURLCollectorBot
- RTCollectorBot

cert-lv:
It would be great to have universal HTTP Collector, but currently `requests` library doesn't support certificates with a password, so there is no easy place to type a password in if you have that. Be ready to implement a temporary workaround.

More info: https://github.com/kennethreitz/requests/issues/1573

Another option is to use `pycurl` library.

sebix:
How important are password-protected for your current needs? File passwords needs to be placed in the config anyway.

Providing client certificates can be implemented very easily. One method is enough - they are equivalent anyway.

sykaeh:
We do not need support for password-protected certificates.

aaronkaplan:
I agree with @sykaeh here. For now IMHO it is enough to support password-less client certificates.
You can convert a password protected client certificate via openssl to a password-less. one.
I 100% agree with @sykaeh approach here. Let's add it in a generic way.

Issue #757, opened 2016-10-27
Clearly define all parameters used in a bot


sykaeh:
Recently I have been creating, modifying and using a lot of bots to fit our setup. And one of the things that has been bothering me is how inconsistently the bot's parameters are defined/documented. Most of the parameters for a BOT are listed in the BOTS file, however I have noticed that the list of parameters is not always complete (especially for parameters that are processed elsewhere like `code` or `accuracy` in collector bots). Also see #668, for a related discussion about the BOTS file.

Additionally, there is no way to explain what the parameter is, whether it is required or optional and what type the parameter is (needed for certtools/intelmq-manager#81). So ideally, you should be able to specify the following items for each parameter:
- variable name
- human readable name
- description
- required/optional
- type (string, boolean, etc.)
- default value (if appropriate)

Other requirements:
- It should be possible to inherit common parameters (e.g. feed parameters (`name`, `code`, `accuracy`, `provider`), collector parameters (`rate_limit`), HTTP request parameters (`http_username`, `http_password`, `http_proxy`), logging defaults, etc.)
- The information is specific to the implementation of the bot and the code relies on it, so it should be in a file close the the code and not in some huge configuration file (like the BOTS file currently)
- Possibility to add a description of the bot so we can remove the BOTS file entirely.
- (optional) Possibility to group parameters together by name (i.e. feed parameters, mail config parameters, etc.)

As far as I can tell, the BOTS file is only used by the intelmq-manager (and possibly as a reference guide of what to put in `runtime.conf` if you do not use the intelmq-manager).

My current suggestion is to put all of this information in a python file (either directly in the bot, in `__init__.py` or in a separate file in the bots module). This allows easy implementation of inheritance and I personally find it just as easy to write Python code as JSON files). The intelmq-manager would then read the information through the intemqctl (or a something similar) thus allowing to get a list of all of the currently installed bots and their configuration settings without having to list all of them in a single file.

sykaeh:
I also just noticed that issue #644 is related.

sykaeh:
I'm not entirely sure what this would mean for the `defaults.conf` file as I do not quite understand how it relates to the `runtime.conf` and `BOTS` file.

sebix:
defaults and runtime hold the same parameters. runtime overrides defaults.
BOTS is only for the manager, it's not used anywhere else.

SYNchroACK:
[...]

PR #767, opened 2016-11-09
Support client certificates for HTTP requests

sykaeh:
This fixes certtools/intelmq#737

I also used this opportunity to consolidate the parameters that are supplied for HTTP requests and added the parameter parsing in the general CollectorBot. 

There also seemed to be inconsistencies whether `http_ssl_proxy` or `https_proxy` should be used. I replaced all of them with `https_proxy` (seemed to be more common and documented).


sebix:
* Renaming parameters means breaking installations
* The newly added parameters are not documented in docs/Bots.md
* http_proxy and https_proxy are provided by defaults as None, you explicitly change it to empty string :(
* As they are always provided, no need to use the getattr workaround
* ssl_cl_cert is not self-explaining, better use ssl_client_cert

aaronkaplan:
overall : looks ok. Did a brief review of the code. Please let's test this internally and merge.
Thanks @sykaeh

sykaeh:
> Renaming parameters means breaking installations

I think for the sake of simplicity and clarity it is worth it. Also, if we change it before 1.0, we do not have to try to maintain the inconsistencies in the future. 

> The newly added parameters are not documented in docs/Bots.md

There are a lot of things that are not documented in docs/Bots.md. I have now added all of the generic collectors along with their parameters. As I have mentioned before in #757, it would be nice to have a better solution to this. Also, I do not understand what exactly is meant by `name`, `lookup` and `public` in the Bots.md file.

> http_proxy and https_proxy are provided by defaults as None, you explicitly change it to empty string :(

Sorry, I hadn't noticed that. I changed it to `null`.

> As they are always provided, no need to use the getattr workaround

I just copied what was done before, but I agree and have now changed it.

> ssl_cl_cert is not self-explaining, better use ssl_client_cert

Same, I just copied what was done before, but I changed it to ssl_client_cert

Issue #688, opened 2016-09-17
Tests: Do not depend on installed files


sebix:
Some tests (bin) depend on files in `/opt/intelmq` (ROOT_DIR), this requirement should be removed.

aaronkaplan:
seems a quite important one if we want to follow the LSB.

Issue #687, opened 2016-09-17
[13] Tests: Skip if no redis or no internet


sebix:
If we have a confined build environment we have to skip all test requiring internet access.

Same for redis if we cannot start the service.

sebix:
For redis this partially works:

    INTELMQ_SKIP_REDIS=1 test-command

And this in front of every class needing redis.

    @unittest.skipIf(os.environ.get('INTELMQ_SKIP_REDIS'), 'Skipping without running redis.')

TODO: Mock the cache.

aaronkaplan:
seems very relevant for the unit tests?

sebix:
What do you mean?

aaronkaplan:
It was a question ;)

PR #769, opened 2016-11-10
Tests: More intuitive, less load and more tests

sebix:
* Reduces the number of tests run -> faster (while coverage remains)
* The bots are only executed if the user requests it -> more intuitive
* tests for more bots (parsers and outputs)
* allows skipping tests requiring redis and/or internet and/or databases (currently only postgres) #687
* postgres tests disabled by default
* Dependency on /opt/intelmq removed #688


sebix:
It's finished, open for comments now!

PR #770, opened 2016-11-11
Run bots directly as executable

sebix:
This proposal removes the `if __name__ == "__main__":` block in every bot and creates executables for each bot. They can now be called directly from the commandline with their module name, e.g. with `$ intelmq.bots.parsers.json.parser json-parser`.

For the packages we could move these executables to e.g. `/var/lib/intelmq/bin/` and add this to the PATH of the user intelmq (in order to not add all executables to the PATH of ordinary users).

I kindly ask for comments and feedback.


bernhard-herzog:
On the whole I think this goes in the right direction. In particular it's a good idea not to try to execute modules in packages as scripts. However:

1. The `Bot.run` method could be a `classmethod`.

    It would be much simpler:

    ```python
    @classmethod
    def run(cls):
        if len(sys.argv) < 2:
            exit('No bot ID given.')
        instance = cls(sys.argv[1])
        instance.start()
    ````

2. Listing all the bots in the `entry_points` parameter of `setup` seems awkward.

    Are they supposed to be managed manually? Might be better to derive them automatically from
   `BOTS`.

3. `intelmqctl start` still tries to execute the module with `python -m`

sebix:
> The Bot.run method could be a classmethod.

I thought of this possibility, but assumed it was not applicable here. Thanks!

> Listing all the bots in the entry_points parameter of setup seems awkward.

Yeah, I just listed it there as kind of proof of concept. To generate the list I went through all modules, using `utils.get_botname_from_module` (see here: https://github.com/sebix/intelmq/commit/760c8fd7992b064a1a0fa861a7dc8a4efcb0f703). Using the bundled BOT file totally makes sense, but we don't have the name of the class there :(
Using the `utils.get_botname_from_module` workaround does not work before installation. But we could put something into the `__init__.py` of the bots (e.g. a __all__).

> intelmqctl start still tries to execute the module with python -m

I knew I forgot something :)

sebix:
> Yeah, I just listed it there as kind of proof of concept. To generate the list I went through all modules, using `utils.get_botname_from_module` (see here: sebix@760c8fd). Using the bundled BOT file totally makes sense, but we don't have the name of the class there :(

In every bot we could define a variable pointing to the relevant class, e.g. `BOT =  AlienVaultOTXCollectorBot`

dmth:
Hi, I didn't look into your code, but would it be possible to derive the bots from the `bots`-directory instead of the `BOTS`-File?
This might start to pave the way to something like a `bots.d` which enables automatic detection of bots without the need of a `BOTS` file

sebix:
> Hi, I didn't look into your code, but would it be possible to derive the bots from the bots-directory instead of the BOTS-File?

They only idea I can come up with is: Iterate over all the files, read them, find out if they are really bots (some are internal or extrernal libs) and find the Name of the class by some regex or similar.

aaronkaplan:
I like this idea. Thx @bernhard-herzog

sebix:
@bernhard-herzog @dmth I now added a variable `BOT` to all bots holding the value of the relevant class. This avoids the ugly magic.

bernhard-herzog:
Merging this would not be a problem for us, I think. It seems that simply replacing the 'id __name__ = "__main__"' construct with BOT = ... in out bots would suffice to adapt our code.

PR #775, opened 2016-11-16
Downgraded logging in Shadowserver parser.

dmth:
This commit contains two changes
 1. The loglevel was downgraded to DEBUG when a value could not
    be inserted into an IntelMQ-DHO field, but was added to
    extras instead.
    This happens quite often for the url, cc_geo, cc_dns
    fields in botnet_drone_hadoop configuration.
    Keeping the loglevel at INFO would have blown up the
    logfiles without need.
 2. Enhanced the configuration of the botnet_drone_hadoop
    parser in order to "guess" the destination.url from
    the hostname and url field of the shadowserver data.


dmth:
From my point of view, this is ok now. Please comment.

aaronkaplan:
apart from my minor comment, I am fine to merge this in. Thanks!

sebix:
> A url is a complete URL (including scheme (|http://|) ). What you
> probably mean is a /path/.
>
The URL we receive from shadowserver has only the path. That's why we
are constructing the URL here.

intelmq/bots/parsers/shadowserver/config.py
```diff
@@ -98,7 +98,22 @@ def convert_int(value):
...
 
-def convert_host_and_url(value, row):
+def convert_hostname_and_url(value, row):
+    """
+    URLs are split into hostname and path, we can also guess the protocol here.
+    but only guess if the protocol is in a set of known good values.
+    """
+    if row['application'] in ['http', 'https', 'irc']:
+        if row['hostname'] and row['url']:
+            return row['application'] + '://' + row['hostname'] + row['url']
```
dmth: To be sure we are talking about the same things:

```
       |------------------ Schema-specific Part   ------------------|

 https://max:muster@www.example.com:8080/index.html?p1=A&p2=B#ressource
 \___/   \_/ \____/ \_____________/ \__/\_________/ \_______/ \_______/
   |      |    |           |         |       |          |         |
Scheme    | Password      Host      Port     Path     Query    Fragment
        User
```
Source https://de.wikipedia.org/wiki/Uniform_Resource_Locator


Yes, the Shadowserver `url` field (`row[`url`]) only includes the `path` part of the url.
The full URL can only be determined when the `host` part ist known, maybe the `port` might also be of use.

Guessing the correct `scheme` is tricky here, as this could only be deduced from the `applciation` field and not from other things. But `application` field is `None` in most cases, or contains funky values like `3.390`. So I limited the guessing to http, https and irc, in all other cases simply the `path` part is returned by this function which will cause a failure when the element is added to the event, resulting in a `url` field in the events extra dict.

Damn, that's hard to read...

aaronkaplan:
@sebix yes, I get that :) I just wanted to point out that probably we should name the things accordingly. But in this case, it's a minor thing.
Apart from that I am OK with the PR. From my pov we can merge it.

sebix:
On 11/18/2016 10:25 AM, AaronK wrote:
>
> @sebix <https://github.com/sebix> I just wanted to point out that
> probably we should name the things accordingly.
>
'url' is the column name in the feed. Renaming the column internally
would result in more confusing IMHO.

aaronkaplan:
> On 18 Nov 2016, at 10:47, Sebastian <notifications@github.com> wrote:
> 
> On 11/18/2016 10:25 AM, AaronK wrote:
> >
> > @sebix <https://github.com/sebix> I just wanted to point out that
> > probably we should name the things accordingly.
> >
> 'url' is the column name in the feed. Renaming the column internally
> would result in more confusing IMHO.
> 

okay, point taken & agreed.
good. Sebix, do you want to merge? Should I?

PR #776, opened 2016-11-16
Redis UNIX socket support

ondj:
UNIX socket can be faster (see http://redis.io/topics/benchmarks) and sometimes is easier to connect Redis with a socket instead of TCP.


ondj:
@sebix Documentation updated

aaronkaplan:
looks good. Was it tested? If yes and if it works, lets merge it and ... also provide sample configs with the unix socket part. 
Great addition!!

Issue #384, opened 2015-10-30
Contribute: Feeds List


SYNchroACK:
If you want to **contribute** creating some bots in order to collect more information, pick one of these feeds:
- [ ] ~~1d4                   https://1d4.us/archive/~~ offline
- [ ] Abuse.ch (RSS)        https://sslbl.abuse.ch/blacklist/
- [x] Abuse.ch (RSS)        https://feodotracker.abuse.ch/blocklist/
- [x] ~~Abuse.ch (RSS)        https://palevotracker.abuse.ch/blocklists.php~~ discontinued
- [ ] ~~Abuse.ch (RSS)        https://spyeyetracker.abuse.ch/blocklist.php~~ discontinued
- [x] Abuse.ch (RSS)        https://zeustracker.abuse.ch/blocklist.php
- [ ] ~~Abusix                http://abusix.org/service/spamfeeds~~ offline
- [ ] ADB Lock Plus         https://easylist-msie.adblockplus.org/malwaredomains_full.tpl
- [ ] ~~Affairedhonneur       http://qwe.affairedhonneur.us/depqfie59y~~ offline
- [ ] ~~Animus Project        https://github.com/animus-project/threat_data~~ offline
- [ ] Anti Hacker Alliance https://anti-hacker-alliance.com/
- [ ] ~~Arbor FastFlux        http://atlas.arbor.net/summary/fastflux~~ discontinued
- [ ] APWG
- [ ] Bad IPs               https://www.badips.com
- [x] Bambenek Consulting   http://osint.bambenekconsulting.com/feeds/
- [ ] Berkeley              https://security.berkeley.edu/aggressive_ips/ips
- [ ] Binary Defense        https://www.binarydefense.com/
- [ ] ~~Bit Cash              http://bitcash.cz/misc/log/blacklist~~ redirects
- [ ] ~~Blade-Defender        http://www.blade-defender.org/eval-lab/~~ offline
- [x] Blueliv               https://www.blueliv.com/enterprise-solution/#threat-intelligence-feed
- [x] BlutMagie             http://torstatus.blutmagie.de/ip_list_all.php/Tor_ip_list_ALL.csv in tor_nodes included
- [ ] BlutMagie             http://torstatus.blutmagie.de/ip_list_exit.php/Tor_ip_list_EXIT.csv
- [ ] BlutMagie             http://torstatus.blutmagie.de/query_export.php/Tor_query_EXPORT.csv
[...]

PR #785, opened 2016-11-22
Nothink DNS Attack Feed parser

jgedeon120:
Feed parser for Nothink's DNS Attack feed.  Referenced in #384


intelmq/bots/parsers/nothink/parser_dns_attacks.py
```diff
@@ -0,0 +1,39 @@
...
+        else:
+            line = line.strip('"').split('","')
+            event = Event(report)
+            event.add('time.source', line[0] + '+00:00')
+            event.add('source.ip', line[1])
+            event.add('source.asn', line[2])
+            event.add('source.as_name', line[3])
+            if FQDN.is_valid(line[4]):
+                event.add('source.reverse_dns', line[4])
+            if len(line[5]) == 2:
```
sebix: Are you checking this to filter out `UNK`? Are something else too?

If not, better check directly for `!= 'UNK'`. I guess UNK is unknown.

intelmq/bots/parsers/nothink/parser_dns_attacks.py
```diff
@@ -0,0 +1,39 @@
...
+            self.tempdata.append(line)
+
+        else:
+            line = line.strip('"').split('","')
+            event = Event(report)
+            event.add('time.source', line[0] + '+00:00')
+            event.add('source.ip', line[1])
+            event.add('source.asn', line[2])
+            event.add('source.as_name', line[3])
+            if FQDN.is_valid(line[4]):
```
sebix: I see only two invalid values: `.` and `n/a`. I'd prefer a simple check for these two instead of the more complex FQDN check.

intelmq/tests/bots/parsers/nothink/test_parser_dns_attacks.py
```diff
@@ -0,0 +1,51 @@
...
+        cls.bot_reference = NothinkDNSAttackParserBot
+        cls.default_input_message = EXAMPLE_REPORT
+
+    def test_event(self):
+        """ Test if correct Event has been produced """
+        self.run_bot()
+        self.assertMessageEqual(0, EXAMPLE_EVENT)
+
+if __name__ == '__main__':
+    unittest.main()
```
sebix: Missing newline at the end.

intelmq/bots/parsers/nothink/parser_dns_attacks.py
jgedeon120: Looking at the list there are ., n/a, and bka.  I went with the FQDN is valid to ensure if the field wasn't valid then the other information would not be lost due to the failure on one field.

intelmq/bots/parsers/nothink/parser_dns_attacks.py
sebix: > ISO3166-1 alpha-2

How do you know this is ISO3166-1 alpha-2?

intelmq/tests/bots/parsers/nothink/test_parser_dns_attacks.py
jgedeon120: I'm not sure how I missed that.  I will get that corrected at my next push when the other comments are also adjusted.

intelmq/bots/parsers/nothink/parser_dns_attacks.py
sebix: Well, I know that we want alpha-2, but is this also the case for the feed?

jgedeon120: Yes it looks that way as the other valid ones are two character.  Also for alpha-3 UNK is a code that is agreed not to be used.

PR #789, opened 2016-11-24
Faster DateTime.__parse method for isoformat

ondj:
Parsing dates with dateutil is very slow. Because harmonization DateTime format is normalized to format generated by `datetime.isoformat` method, this patch adds method which can parse this format much faster and if it is not isoformat, then is passed to dateutil.


sebix:
Thanks!

PR #790, opened 2016-11-24
Cache also invalid responses from DNS server in reverse_dns bot

ondj:
For 60 or 3600 seconds, because in some cases response can take 5 seconds and it can slow down the whole queue.

### DIFF
--- a/intelmq/bots/experts/reverse_dns/expert.py
+++ b/intelmq/bots/experts/reverse_dns/expert.py
@@ -64,8 +64,10 @@ class ReverseDnsExpertBot(Bot):
                         result = None
                         raise ValueError
                 except (dns.exception.DNSException, ValueError) as e:
-                    # Set longer TTL for 'DNS query name does not exist' error
-                    ttl = 3600 if isinstance(e, dns.resolver.NXDOMAIN) else 60
+                    # Set default TTL for 'DNS query name does not exist' error
+                    ttl = None if isinstance(e, dns.resolver.NXDOMAIN) else \
+                        getattr(self.parameters, "cache_ttl_invalid_response",
+                                60)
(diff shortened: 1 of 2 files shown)


intelmq/bots/experts/reverse_dns/expert.py
```diff
@@ -61,8 +64,10 @@ def process(self):
                         result = None
                         raise ValueError
                 except (dns.exception.DNSException, ValueError) as e:
-                    if isinstance(e, dns.resolver.NXDOMAIN):
-                        continue
+                    # Set longer TTL for 'DNS query name does not exist' error
+                    ttl = 3600 if isinstance(e, dns.resolver.NXDOMAIN) else 60
```
sebix: Why are you using another TTL for NXDOMAIN as for regular responses?

And I'd like to make the TTL in case of other errors configurable too, e.g. with `cache_ttl_invalid_response = 60` for the reverse dns.

intelmq/bots/experts/reverse_dns/expert.py
```diff
@@ -61,8 +64,12 @@ def process(self):
                         result = None
                         raise ValueError
                 except (dns.exception.DNSException, ValueError) as e:
-                    if isinstance(e, dns.resolver.NXDOMAIN):
-                        continue
+                    # Set default TTL for 'DNS query name does not exist' error
+                    ttl = None if isinstance(e, dns.resolver.NXDOMAIN) else \
```
sebix: Why do you use `None` now instead of `redis_cache_ttl`?

ondj: Because `None` means default TTL and default TTL is `redis_cache_ttl`.

ondj:
Thanks!

Issue #792, opened 2016-11-24
intelmqdump no longer working


sykaeh:
The commit af1bd9645408cad99e2b7325a406d541e3796384 broke intelmqdump, specifically the following section in intelmqdump.py:

```
if answer not in available_opts:
     print('Action not allowed.')
     continue
```

The problem is that `available_opts` is not the correct thing to test for. 
```
available_opts = ['(s)how by ids', '(r)ecover by ids', 'delete (e)ntries', 'recover (a)ll', '(d)elete file', '(q)uit']
```
The correct array should be: `['s', 'r', 'e', 'a', 'd', 'q']`. Also, it needs to be `answer[0]` since answer is an array.

Also, I saw that some `TODOs` in `intelmq/bots/collectors/mail/lib.py` where done, however I have not found a place where that file is actually used. I think it can be deleted.

PR #794, opened 2016-11-24
Fixes intelmqdump issue

sykaeh:
Fixes certtools/intelmq#792


intelmq/bin/intelmqdump.py
```diff
@@ -198,7 +198,7 @@ def main():
         else:
             if not answer:
                 continue
-        if answer not in available_opts:
+        if len(answer) == 0 or answer[0] not in ['s', 'r', 'e', 'a', 'd', 'q']:
```
sebix: In case of corrupted files this does allow invalid methods, see [line 175](https://github.com/sykaeh/intelmq/blob/master/intelmq/bin/intelmqdump.py#L175)

intelmq/bots/collectors/mail/lib.py
```diff
@@ -1,50 +0,0 @@
-# -*- coding: utf-8 -*-
```
sebix: I guess this change is accidentally in this commit/PR as it has nothing to do with issue #792?

sykaeh: As I already mentioned in #792:

> Also, I saw that some TODOs in intelmq/bots/collectors/mail/lib.py where done, however I have not found a place where that file is actually used. I think it can be deleted.

sebix:
Thanks!

PR #796, opened 2016-11-28
Do not log password parameters

ondj:


sebix:
Why do we want to hide some parameters? The debug log level is for *debugging*.

dmth:
Without looking at the code, replacing the `password` value in the logs with an obfuscated thing is ok, IMHO. Exchange of debugging-logs is a little bit safer now, as you lower the risks of sharing passwords without intent.

sebix:
> Without looking at the code, replacing the password value in the logs with an obfuscated thing is ok, IMHO. Exchange of debugging-logs is a little bit safer now, as you lower the risks of sharing passwords without intent.

Some feeds authenticate via user-agent, others with get-parameters, or auth-tokens or auth-keys. So to do it right, we obfuscate all URLs and user-agents, everything containing `pass`, `auth` or `token`. Sooo, what about a whitelisting approach...? Or remove the logging of parameters altogether?

dmth:
> Some feeds authenticate via user-agent, others with get-parameters, or auth-tokens or auth-keys. So to do it right, we obfuscate all URLs and user-agents, everything containing pass, auth or token. 

True, but do we need to go for the 100% solution or is it ok, to start working with 5% and enhance the solution step by step?

> Sooo, what about a whitelisting approach...? Or remove the logging of parameters altogether?
Remove logging of all Parameters might also be insufficient, as this line:
```
 self.logger.info("Downloading report from %s" %
                         self.parameters.http_url)
```
in https://github.com/certtools/intelmq/blob/master/intelmq/bots/collectors/http/collector_http.py might log the tokens you already mentioned. 

So let's start with small steps here...

sykaeh:
I also think that not logging the actual passwords is a good idea. I always get a bad feeling when I see passwords in plaintext.

wagner-certat:
@dmth It's not sufficient to modify this one line. In debug mode, we also log the full events. You need to remove them too.

I agree that logging credentials is bad, but we shouldn't do it only in some places and ignore others. This will create the wrong impression that it's fine to share full logs, because credentials apparently aren't logged. IMHO we shouldn't deceive the users.

dmth:
I see. But it's not about "deceiving"  (that's a very strong word!) the users.

I neither see the necessity to reach the 100% solution, nor I'm sure we could reach such as solution right now. 
We should start with a 1% solution and increment it.
This will make it better one step at a time. With this PR we can take the first step of the journey.

Issue #655, opened 2016-08-31
cleanmx-phishing parser - wrong key


SYNchroACK:
https://github.com/certtools/intelmq/blob/master/intelmq/bots/parsers/cleanmx/parser_phishing.py#L17

In that line virusname column from the source means "target" field like in PhishTank. Anyone can double check if I'm right?!

dmth:
As I do not have access to the cleanmx feeds, I cannot validate this.

dmth:
Had a look at today's data: From my point of view the field "virusname" in csv does not make sense at all and is misleading:
Example-Values: `eBay`, `Google`, `Free`, `Microsoft`, `Paypal`, an many more...

Using the field `malware.name` is wrong in this context. `event_description.target` might be a better place for this information. On the other hand, the real target might be the end-user and not the organistation reported in the `virusname` field.

sebix:
> On the other hand, the real target might be the end-user and not the organistation reported in the `virusname` field.

The user is always implicitly the target, so I think it's not useful to mention it explicitly.

I agree `event_description.target` is better.

PR #798, opened 2016-11-30
ENH+TST: Revised cleanmx parser

sebix:
fixes certtools/intelmq#655


dmth:
https://github.com/certtools/intelmq/issues/818 Seems like the BOTS file was forgotten here.

PR #804, opened 2016-12-02
File output bot check if file still exists

franznemeth:
I am writing this PR because of how we use intelmq.

We are using TI gathered with intelmq for our SEM platform. 
The problem is that after reading the output file the SEM platform deletes the file which leads to the file-output bot writing into a file which doesnt exist on the system anymore.


dmth:
Hi,
Thank you very much for your contribution.

Currently your patch is in conflict with master, as the way to start bots has changed. @sebix can tell you more if you are interested.

You can fix the conflict by replacing the `if __name__ == "__main__":` Block with `BOT = FileOutputBot`

franznemeth:
Hi,

Yes it would make sense to set the logging to debug in the try block.

I will implement the changes tonight, should I open a new PR and close this one?

sebix:
> I will implement the changes tonight, should I open a new PR and close this one?

Just push to your branch (which is master in your case, **I recommend to not use master for PRs**, see also https://github.com/certtools/intelmq/wiki/Developers-Guide#workflow)

I will fix merge conflicts when merging anyway.

intelmq/bots/outputs/file/output.py
```diff
@@ -14,12 +15,21 @@ def init(self):
...
-        self.file.write("\n")
-        self.file.flush()
-        self.acknowledge_message()
+
+        try:
+            os.stat(self.parameters.file)
+        except FileNotFoundError:
+            self.logger.debug("Reopening %r file." % self.parameters.file)
+            self.file = io.open(self.parameters.file, mode='at', encoding="utf-8")
+            self.logger.info("File %r is open." % self.parameters.file)
```
sebix: The very same line in init() also uses INFO-level.

These 3 lines could be replaced by simply calling `self.init()`.

intelmq/bots/outputs/file/output.py
```diff
@@ -14,12 +15,21 @@ def init(self):
...
+        
 
         event_data = event.to_json(hierarchical=self.parameters.hierarchical_output)
-        self.file.write(event_data)
-        self.file.write("\n")
-        self.file.flush()
-        self.acknowledge_message()
+
+        try:
+            os.stat(self.parameters.file)
```
sebix: If the file has been deleted between the `os.stat()` and the write here, this will fail again. I think it is better to simply try to write.

sebix:
You are very far behind master, which wouldn't be a problem, but you now used class name from master's version (`FileOutputBot`) for `BOT = ` but left the classname `FileBot`. This won't work.

To bring your version up-to-date with master, please do a **merge**. I have summarized the workflow here: https://github.com/certtools/intelmq/wiki/Developers-Guide#workflow and there are very good tutorials out there.

Now to the issue I have raised: You are doing `os.stat()` and then you are writing. What happens if the file is being deleted inbetween? Think of a `os.remove(self.parameters.file)` after the stat.

franznemeth:
Thanks for your help!

The removal of the file between os.stat() and the write could be solved by using "with". For that I would have to move the opening of self.parameters.file from  __init__ to the process method.
Would that be acceptable?

wagner-certat:
I'd just do:

```python
try:
    # write
except FileNotFoundError:
    self.init()
else:
    # acknowledge
```

intelmq/bots/outputs/file/output.py
```diff
@@ -1,9 +1,12 @@
 # -*- coding: utf-8 -*-
 import io
+import sys
+import os
 
 from intelmq.lib.bot import Bot
 
 
+
```
wagner-certat: pep8: too many blank lines

intelmq/bots/outputs/file/output.py
```diff
@@ -1,9 +1,12 @@
 # -*- coding: utf-8 -*-
 import io
+import sys
```
wagner-certat: unused import

intelmq/bots/outputs/file/output.py
```diff
@@ -13,12 +16,17 @@ def init(self):
 
     def process(self):
         event = self.receive_message()
-
         event_data = event.to_json(hierarchical=self.parameters.hierarchical_output)
-        self.file.write(event_data)
-        self.file.write("\n")
-        self.file.flush()
         self.acknowledge_message()
```
wagner-certat: acknowledge before data is actually written

wagner-certat:
merged with modifications: 3c6f491eba747c9d3e13ac70e11ef271b22a42b4

PR #814, opened 2016-12-09
Bot process manager

ondj:
Moved bot process management code to `BotProcessManager` class. Now it can be possible to simply change way how bot processes are managed and use a different system (for example systemd or supervisor).


sebix:
Thanks and sorry for the delay!

Issue #779, opened 2016-11-16
CSV parsers: ignore comments


sebix:
The python csv parser can't ignore comment lines starting with # and has problems with empty lines at the beginning. Should be fixed directly in `ParserBot.parse_csv`

PR #821, opened 2016-12-13
CSV parser: ignore empty lines and comments

sebix:
fixes certtools/intelmq#779

The generic csv parser can now ignore lines starting with a defined list of characters to ignore comments.

Also, the message is stripped, so newlines at the beginning and the end are removed, which too poses problems onto the csv parser of python.

### DIFF
--- a/intelmq/lib/bot.py
+++ b/intelmq/lib/bot.py
@@ -451,6 +451,8 @@ class Bot(object):
 
 
 class ParserBot(Bot):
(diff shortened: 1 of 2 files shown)


wagner-certat:
Could someone review this please?

SYNchroACK:
Code makes sense, only conceptual things that I will review later but is not enough to to block the merge.

PR #823, opened 2016-12-13
Minor fixes

dmth:

Issue #822, opened 2016-12-13
Generic CSV-Parser broken when reading column-definition from a string value


dmth:
Currently the generic CSV Parser does only parse one line correctly, when the columns are a configured as a string (this especially happens when you are using the manager for example).

Currently the first line (not header) is read and resolved according to the column definition. All following lines are not resolved and added as raw.

If the following patch is applied all lines will be read and resolved again:

```
diff --git a/intelmq/bots/parsers/generic/parser_csv.py b/intelmq/bots/parsers/generic/parser_csv.py
index 29ee44a..ae4e78d 100644
--- a/intelmq/bots/parsers/generic/parser_csv.py
+++ b/intelmq/bots/parsers/generic/parser_csv.py
@@ -30,7 +30,7 @@ def parse(self, report):
         self.columns = self.parameters.columns
         # convert columns to an array
         if type(self.parameters.columns) is str:
-            self.columns = map(str.strip, self.columns.split(","))
+            self.columns = self.columns.split(",")
 
         if hasattr(self.parameters, 'type_translation'):
             self.type_translation = json.loads(self.parameters.type_translation)
```

I have no Idea where the `map` statement originated from. My original code-fragment did not contain it.

dmth:
https://github.com/Intevation/intelmq/commit/22c09ef060215316736c5e3f4c1c92ee719259e5 fixes the problem

PR #824, opened 2016-12-14
CSV parser: Fix parsing of column names given as 

sebix:
And add a test with more than one line, would have detected this issue earlier.

fixes certtools/intelmq#822
closes certtools/intelmq#823


intelmq/bots/parsers/generic/parser_csv.py
```diff
@@ -24,17 +24,19 @@
...
-        if type(self.parameters.columns) is str:
-            self.columns = map(str.strip, self.columns.split(","))
+        if type(self.columns) is str:
+            self.columns = [column.strip() for column in self.columns.split(",")]
 
         if hasattr(self.parameters, 'type_translation'):
             self.type_translation = json.loads(self.parameters.type_translation)
 
+    def parse(self, report):
+        print(self.columns)
```
dmth: Why the print?

sebix: :/

Issue #825, opened 2016-12-15
Clarification about "enabled" switch


sykaeh:
Issue #715 introduced an enabled switch and as far as I understood this was another parameter of the bot (i.e. like `rate_limit` or `hierarchical_output`) that could be used for any bot (or is it only collectors?). This is also documented in [Bots.md](https://github.com/certtools/intelmq/blob/b5c87cafb6be758963714978fb114d3a0dc39239/docs/Bots.md).

However, when I tried to use the enabled parameter, it did not do anything. Looking into the source code, the changes in 06cde54d914ff4d025d012980fa003fe318cf990 seem to show that the enabled flag is one level higher (on the same level as `name` and `module`). Is there an error in the code or in the documentation?

dmth:
I cannot tell you why it's not working. The changeset you are referencing ist for intelmqctl. The switch you most likely refer to was introduced here https://github.com/certtools/intelmq/commit/9d0db8aa1c9ce325e81d7841834034f3f6f4ccd7 and was genralised here https://github.com/certtools/intelmq/commit/2828cf6b90f5d18c07ca4246bf408117f22fcbe2

ondj:
I think that the problem is in [this code](https://github.com/certtools/intelmq/blob/66eb8f2bbac9cfbd538529b5e5b70ae37976cd00/intelmq/bin/intelmqctl.py#L463):
```python
    def botnet_start(self):
        botnet_status = {}
        for bot_id in sorted(self.runtime_configuration.keys()):
            if self.runtime_configuration[bot_id].get('enabled', True):
```

Because `enabled` is loaded directly from `runtime_configuration`  and not from `parameters` array.

sebix:
As enabled is not related to the parameters for the bot itself, but rather for the initialization (just as module), I think this behavior is fine.

ondj:
Sorry, now I get it. So maybe it could help when `enabled` will be added to all bots in BOTS file with `true` value?

sykaeh:
So, to summarize:
1. The enabled switch is only used in the `intelmqctl` (it was removed from the other places in 408b969)
2. The enabled switch is part of the initialization of the bot, so it should be added at that level. For example (in `runtime.conf`):
```
    "shadowserver-parser": {
        "group": "Parser",
        "name": "ShadowServer",
        "module": "intelmq.bots.parsers.shadowserver.parser",
        "description": "ShadowServer parses the shadowserver reports.",
        "parameters": {
            "feedname": ""
[...]

PR #830, opened 2017-01-04
'enabled' parameter clarification

SYNchroACK:
related to #825 

@sykaeh please check and give your feedback :) Thank you!

### DIFF
--- a/docs/User-Guide.md
+++ b/docs/User-Guide.md
@@ -135,6 +135,8 @@ All bots inherit this configuration parameters and they can overwrite them using
 
 Small extract:
 
+* `enabled`: This option defines if a bot is enable to run or not. It's useful when `intelmqctl` is executed to start all bots except the ones defined as disabled.
+     * **`true/false`** - allows the bot to run or not
 * `logging_handler`: Can be one of `"file"` or `"syslog"`.
 * `logging_level`: Defines for all system the level of logging that will be use by all bots and intelmqctl tool. Possible values are: `"CRITICAL"`, `"ERROR"`, `"WARNING"`, `"INFO"` and `"DEBUG"`.
 * `logging_path`: If `logging_handler` is `file`. Defines for all system the logs folder that will be use by all bots and intelmqctl tool. Default value is: `/opt/intelmq/var/log/`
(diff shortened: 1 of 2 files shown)


SYNchroACK:
@sebix I tried locally fix the config test but no luck. If you have chance, can you double-check this one? Thank you mate.

sykaeh:
@SYNchroACK I thought we decided **not** to add it `defaults.conf` since it is not like the other parameters and just clarify it in the documentation.

(The config test fails because `defaults.conf` needs to be sorted in alphabetical order)

wagner-certat:
That's not how the parameter works and is intended to work.

> @sebix I tried locally fix the config test but no luck. If you have chance, can you double-check this one? Thank you mate.

It's simply not sorted.

SYNchroACK:
> @SYNchroACK I thought we decided not to add it defaults.conf since it is not like the other parameters and just clarify it in the documentation.

I added it to default.conf because its the only way for IntelMQ-Manager to get it loaded on the interface.
I think with this approach is 99% easy for the user to understand and proceed without need to learn another way to do the things...the user will simply configure the bot with this parameter like other parameter. I just did not find any clear and easy solution from the user point of view. @sykaeh go ahead and suggest another approach (the pull req will not merge without your feedback to my comment) :)

wagner-certat:
On 01/10/2017 08:29 PM, SyNchroAcK wrote:
>
> I added it to default.conf because its the only way for
> IntelMQ-Manager to get it loaded on the interface.
>
You just need to add it in the manager:
*
https://github.com/certtools/intelmq-manager/blob/1a7bfabd93fa781772c1abf09a00f8841babf63c/intelmq-manager/js/main.js#L96
*
https://github.com/certtools/intelmq-manager/blob/1a7bfabd93fa781772c1abf09a00f8841babf63c/intelmq-manager/js/runtime.js#L22
*
https://github.com/certtools/intelmq-manager/blob/1a7bfabd93fa781772c1abf09a00f8841babf63c/intelmq-manager/js/static.js#L32

SYNchroACK:
Closed due a conflit with my branchs on forked repo. Re-opened on #838

PR #837, opened 2017-01-11
Remove new lines from mail subject in mail collector

ondj:
E-mail subject can contain newlines characters, but in e-mail format, it is only separator with no meaning. And without removing, subject regex can stop working anytime for a long subject.

### DIFF
--- a/intelmq/bots/collectors/mail/collector_mail_attach.py
+++ b/intelmq/bots/collectors/mail/collector_mail_attach.py
@@ -29,7 +29,7 @@ class MailAttachCollectorBot(CollectorBot):
 
                 if (self.parameters.subject_regex and
                         not re.search(self.parameters.subject_regex,
-                                      message.subject)):
+                                      re.sub("\r\n\s", " ", message.subject))):
                     continue
 
                 for attach in message.attachments:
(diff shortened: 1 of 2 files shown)


wagner-certat:
this does not work for me:
```python
>>> re.sub("\r\n\s", "", "fooo\n\r bar")
'fooo\n\r   bar'
```
And the mails in my inbox don't have a `\r`, also maybe use `\s*`?

dmth:
For reference the RFC: https://www.ietf.org/rfc/rfc2822.txt

We need to unfold the header here:

"Unfolding is accomplished by simply removing any CRLF
   that is immediately followed by WSP.  Each header field should be
   treated in its unfolded form for further syntactic and semantic
   evaluation."

aaronkaplan:
For me this PR is OK. I have no issues with it. From my side , we can merge.

wagner-certat:
@dmth 

> EDIT: Maybe not. Proper unfolding should be done by IMBOX, at least I suppose it does, why does your header contain these Characters?

From imbox I get this as subject:

    'very longvery longvery longvery longvery longvery longvery longvery\r\n longvery longvery longvery longvery longvery longvery longvery longvery\r\n longvery longvery longvery longvery longvery longvery longvery longvery\r\n longvery long'

Applying the proposed substitution:

    'very longvery longvery longvery longvery longvery longvery longvery longvery longvery longvery longvery longvery longvery longvery longvery longvery longvery longvery longvery longvery longvery longvery longvery longvery long'

Issue #799, opened 2016-11-30
Single parser doing multiple feeds


jgedeon120:
The developers guide has a section in it under Directory layout in the repository that discusses having multiple parsers for the different feeds coming from the same root source.  The example given is Abusech and the different types of feeds for Zeus.  Recently using the IntelMQ Manager UI started getting a little long on the parsers.  I started to start combining parsers into a single parser for the different types of feeds within the same root source.  Is this not advisable or wanted? Looking at the experts they seem to be created using if key in event so combining the feeds and then having the expert after the parser will add a little overhead if multiple collectors are pointed to the same parser.  An example of this is displayed below in the screen shot.

![screen shot 2016-11-23 at 7 52 29 pm](https://cloud.githubusercontent.com/assets/8508738/20775898/54e9d26a-b72b-11e6-8ab2-346705109fe4.png)

sebix:
This is very feasible to have, yes. We already have some parsers which process different feeds with the same provider, as the format, data and thus processing is very similar. We also have generic parsers (currently only a minimal json and a quite sophisticated csv) which could be used in more places, at least if we do not need to convert data - which is often the case.

dmth:
The approach is similar to approach 2) sketched in https://github.com/certtools/intelmq/pull/752#issuecomment-259384416

sykaeh:
Since this is generally already possible, are there any specific parsers that you were thinking of combining (that haven't been done already)? Otherwise I think we could close this issue.

wagner-certat:
Are there any open questions here?

PR #839, opened 2017-01-11
Developers-Guide Improvements

SYNchroACK:
* layout rules - relate to #799 
* document restructured

### DIFF
--- a/docs/Developers-Guide.md
+++ b/docs/Developers-Guide.md
@@ -1,12 +1,45 @@
 **Table of Contents**
 
-1. [Intended Audience](#audience)
-2. [Goals](#goals)
-2. [Code and Repository Rules](#code-and-repository-rules)
-2. [System Overview](#system-overview)
-3. [Bot Developer Guide](#bot-developer-guide)
+* [Intended Audience](#intended-audience)
+  * [Goals](#goals)
+* [Development Environment](#development-environment)
+  * [Installation](#installation)
+  * [Testing](#testing)
+* [Development Guidelines](#development-guidelines)
+  * [Coding-Rules](#coding-rules)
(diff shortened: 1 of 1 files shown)


docs/Developers-Guide.md
```diff
@@ -203,9 +203,10 @@ intelmq/
     system.conf
 ```
 
-Assuming you want to create a bot for 'Abuse.ch Zeus' feed. It turns out that here it is necessary to create different parsers for the respective kind of events (C&C, Binaries, Dropzones). Therefore, the hierarchy ‘intelmq/bots/parser/abusech/parser.py’ would not be suitable because it is necessary to have more parsers, as mentioned above. The solution is to use the same hierarchy with an additional "description" in the file name, separated by underscore. Also see the section *Directories and Files naming*.
+Scenario:
+Assuming that you want to create a bot for 'Abuse.ch Zeus' feed. It turns out, just as an example, that here it is necessary to create different parsers for the respective kind of events (C&C, Binaries, Dropzones). Therefore, the hierarchy ‘intelmq/bots/parser/abusech/parser.py’ would not be suitable because it is necessary to have more parsers, as mentioned above. The solution is to use the same hierarchy with an additional "description" in the file name, separated by underscore. Also see the section *Directories and Files naming*.
 
-Example:
+Example: (the following filepaths does NOT exist, they are just for learning purposes)
```
wagner-certat: Why don't you replace the old examples with current ones? E.g. intelmq/bots/parsers/abusech/parser*.py

SYNchroACK:
Folks, I merged this one since it's just documentation. Please review if you have a chance.
Apart the possible small typos/errors, is useful to merge now.

PR #838, opened 2017-01-11
'enabled' parameter clarification

SYNchroACK:
related to #825

PR #840, opened 2017-01-11
Clarify 'enabled' switch

sykaeh:
Fixes certtools/intelmq#825.

After much discussion and misunderstanding, I think this is what we decided on in #825. This  PR replaces #830 and #838. 

Changes for the intelmq-manager are still needed. See issue certtools/intelmq-manager#92

### DIFF
--- a/docs/Bots.md
+++ b/docs/Bots.md
@@ -5,9 +5,13 @@
 3. [Experts](#experts)
 4. [Outputs](#outputs)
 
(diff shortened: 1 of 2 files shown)


SYNchroACK:
@wagner-certat already reviewed, I also checked. Assuming your perspective to solve the issue, this pull request is correct and that's why I'm merging now. Although, as soon as possible, I will revisit this issue on a concept perspective.

Thank you @sykaeh  :)

PR #841, opened 2017-01-11
Message class: deprecations and new methods

wagner-certat:
* deprecates Message.contains
* deprecate force parameter in Message.add and add new overwrite parameter
* adds Message.is_valid
  (Note: I couldn't reduce the duplicate code in .add and .is_valid here, because some values are ignored while all other are explicitly invalid.)
* Added the raise_failure parameter to Message.add

Please review and give feedback if this is ok. Then I will do the replacements in the code base.

### DIFF
--- a/intelmq/bots/experts/abusix/expert.py
+++ b/intelmq/bots/experts/abusix/expert.py
@@ -34,7 +34,7 @@ class AbusixExpertBot(Bot):
                 ip = event.get(ip_key)
                 email = self.lookup(ip)
                 if email:
-                    event.add(abuse_contact_key, email, force=True)
+                    event.add(abuse_contact_key, email, overwrite=True)
 
         self.send_message(event)
         self.acknowledge_message()
--- a/intelmq/bots/experts/abusix/expert.py
+++ b/intelmq/bots/experts/abusix/expert.py
@@ -30,7 +30,7 @@ class AbusixExpertBot(Bot):
             abuse_contact_key = key + "abuse_contact"
             if abuse_contact_key in event and not self.parameters.overwrite:
                 continue
-            if event.contains(ip_key):
+            if ip_key in event:
                 ip = event.get(ip_key)
                 email = self.lookup(ip)
                 if email:
--- a/intelmq/lib/message.py
+++ b/intelmq/lib/message.py
@@ -155,6 +155,8 @@ class Message(dict):
         self.add(key, value, force=True, sanitize=sanitize)
 
     def contains(self, key):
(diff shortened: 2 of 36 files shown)


intelmq/lib/message.py
```diff
@@ -106,14 +106,47 @@ def __init__(self, message=(), auto=False, harmonization=None):
     def __setitem__(self, key, value):
         self.add(key, value)
 
-    def add(self, key, value, sanitize=True, force=False, ignore=()):
-        if not force and key in self:
-            raise exceptions.KeyExists(key)
+    def is_valid(self, key, value, sanitize=True):
+        """
+        Checks if a value is valid for the key.
```
SYNchroACK: Add a phrase regarding `sanitize` parameter as a dev documentation.

SYNchroACK: **Suggestion:** Add a phrase regarding sanitize parameter as a dev documentation.

intelmq/lib/message.py
```diff
@@ -106,14 +106,47 @@ def __init__(self, message=(), auto=False, harmonization=None):
...
+        if sanitize:
+            value = self.__sanitize_value(key, value)
+        valid = self.__is_valid_value(key, value)
+        if valid[0]:
+            return True
+        return False
 
-        if value in ["-", "N/A"]:
+    def add(self, key, value, sanitize=True, force=False, overwrite=False, ignore=(),
+            raise_failure=True):
```
SYNchroACK: **Suggestion:** Add the same kind of documentation as you have in `is_valid` method with special attention to `raise_failure` parameter.

wagner-certat:
> Should we replace all bots code which is using the is_valid method from each harmonization type (String, FQDN, URL, etc)?

Yes, I will do so. If the library changes are ok.

intelmq/bots/experts/abusix/expert.py
```diff
@@ -30,7 +30,7 @@ def process(self):
             abuse_contact_key = key + "abuse_contact"
             if abuse_contact_key in event and not self.parameters.overwrite:
                 continue
-            if event.contains(ip_key):
+            if ip_key in event:
```
aaronkaplan: I really prefer this syntax: if ip_key in event:
That's may more pretty.

SYNchroACK:
> Yes, I will do so. If the library changes are ok.

Yap, everything ok @wagner-certat

wagner-certat:
I did all replacements, please have a quick glance.

SYNchroACK:
> I did all replacements, please have a quick glance.

Good to go! Nice work

PR #851, opened 2017-01-20
Update config.py

aaronkaplan:
According to BSI, it makes more sense to have the machine name along next to the WORKGROUP  name in the extra field and not as a separate field. 
Also the machine name is not really an identifiable item. In other feeds it is. But here it's not.
Hence -> move it to extra.


dmth:
I think this change is ok.

wagner-certat:
Merged

PR #853, opened 2017-01-20
More testdata for shadowserver parsers + Script to generate data

dmth:
This PR provides some anonymized testdata for some common shadowserver parsers.

In addition, it adds an awk script that can be used to generate anonymized datasets from original shadowserver data.


aaronkaplan:
Looks ok for me. It's just test data. Did not try the awk script yet however.

wagner-certat:
Why 93 times basically the same lines? Just four, five lines covering all the special cases is just fine. I don't see any reason why we should have identical lines with different city names.

SYNchroACK:
> Why 93 times basically the same lines? Just four, five lines covering all the special cases is just fine. I don't see any reason why we should have identical lines with different city names.

also have the same doubt. :/

wagner-certat:
This PR adds *2219* lines to add test data for *26* feeds. Do you want to do speed measurements? I think this has nothing to do with unittests.

dmth:
You are right, It's to much for unit-tests, but for development and testing of more complex scenarios, such as notifications and aggregation, more testdata is very helpful.

aaronkaplan:
> On 17 Mar 2017, at 12:39, Dustin Demuth <notifications@github.com> wrote:
> 
> You are right, It's to much for unit-tests, but for development and testing of more complex scenarios, such as notifications and aggregation, more testdata is very helpful.
> 

How about moving a large set of test data to some other place and keeping a small and manageable set of test data for the unit tests?

dmth:
> How about moving a large set of test data to some other place and keeping a small and manageable set of test data for the unit tests?

Works for us.  What's the status of the homepage? I think this might be a candidate.

bernhardreiter:
Personally I tend to recommend putting the new testfiles as is. They may inspire more automated testing (which is good even if it is not unit-testing) and they belong close to the parsers, because they should be current and match the parsers.  

A better organisation of files will happen once it is possible to have a set up bots bundled together separately in one subdir, so we can have bots which are updated separately. As long as this does not happend, I believe it is better to keep files close to where they should be maintained.

wagner-certat:
What are the feeds for the following files:
* drone.csv (It's not Sinkhole-HTTP-Drone, that's another file)
* open-db2-discovery-service.csv We don't have this feed in our parser at all: https://db2scan.shadowserver.org/
* vulnerable-nat-pmp.csv 
* vulnerable-nat-pmp-systems.csv

We have a Open-NATPMP, but not two?

wagner-certat:
I deactivated the mentioned tests. If you know the answers for https://github.com/certtools/intelmq/pull/853#issuecomment-306523063 we can activate them again.

dmth:
Sorry for not responding here. As I cannot answer your question, I'm fine with this approach.

Issue #666, opened 2016-09-06
Timeout for bots


sebix:
In another program but in a similar context we encountered a problem with long-running HTTP-requests. More specifically, in a request the data rate drops significantly sometimes to a very low rate. As a consequence the job is not finishing and blocks also other runs (the program was started regularly with cron).

This shows that we need something like a timeout for all kind of lookups/requests. This can be solved by spawning a thread which is to be killed after some time. But python does not support killing threads :( We can still kill the whole process and the init-system (e.g. systemd) automatically spawns a new process.
Maybe there is a better solution?

Setting timeouts for a request itself is not useful in the case described above, data was still flowing. However we can set timeouts for creating the connection and the read. But these are very simple fixes.

aaronkaplan:
How do you handle a time out then?

dmth:
If this concerns the HTTP-Timeouts of the HTTP-Collectors, those can be set quite simple:

https://requests.readthedocs.io/en/master/user/quickstart/#timeouts
https://requests.readthedocs.io/en/master/user/advanced/#timeouts

> By default, requests do not time out unless a timeout value is set explicitly.

and 

> The timeout value will be applied to both the connect and the read timeouts. Specify a tuple if you would like to set the values separately:

aaronkaplan:
> On 06 Sep 2016, at 14:11, Sebastian <notifications@github.com> wrote:
> 
> In another program but in a similar context we encountered a problem with long-running HTTP-requests. More specifically, in a request the data rate drops significantly sometimes to a very low rate. As a consequence the job is not finishing and blocks also other runs (the program was started regularly with cron).
> 
> This shows that we need something like a timeout for all kind of lookups/requests. This can be solved by spawning a thread which is to be killed after some time. But python does not support killing threads :( We can still kill the whole process and the init-system (e.g. systemd) automatically spawns a new process.
> Maybe there is a better solution?
> 
Does not sound very nice.
I prefer @dmth's approach.

And if data is flowing (but every slowly) it is not something we can really handle right now.
What we *can* do is to record the events/sec rate in each bot and visualise that.
This will at least allow the operator to determine the source of the performance problem.

My 2 cents,
a.

sebix:
> If this concerns the HTTP-Timeouts of the HTTP-Collectors, those can be set quite simple:

This is true, but not the case which I described above. If the rate is very low, the timeout does not come into effect. (This is scenario is *not* hypothetical)

sebix:
[...]

PR #835, opened 2017-01-10
AnubisNetworks Cyberfeed bot revised and added HTTP timeouts

wagner-certat:
* dropped the collector, just use the generic http stream collector (pycurl is no longer needed)
* extended the parser to use more fields
* readded generic http stream collector to docs

PR #859, opened 2017-01-25
Timeouts for some collectors using "requests"

### LINKED ISSUE #854: Generic Mail URL Fetcher hangs forever on Downloading report
On our customers production system it was discovered today, that some of the collector_mail_url bots were hanging (four out of 19).  The symptoms were Identical in all cases:

* The last log message was:
  ```INFO - Downloading report from https://example.com/valid/download/url```
* intelmqctl status showed the bot as "is running"
* stracing the hanging bots process showed an read in progress:  ```read(7,```
* `ss -tpn` showed an hanging connection:
   ```CLOSE-WAIT 97     0             10.0.0.1:60822        10.0.0.2:993    users:(("python3",5627,5))```
  but as you can see this is IMAP, not the download, I suspect the imap part of the bot was waiting for the dl to finish.

Maybe the bots should get a hard timeout for all network operations, so that if something gets stuck they get restarted (some kind of watchdog).?

dmth:
This PR adds the capability of setting a timeout to the collector_http and collector_mail_url.

If a timeout occurs, the connection is tested again for two times. If the request fails three time in a row, the collector will start processing again, when the `rate_limit` interval was exceeded.

This PR is in conflict with https://github.com/certtools/intelmq/pull/835#discussion_r96223317
This PR fixes #854 and #666


intelmq/bots/collectors/http/collector_http.py
```diff
@@ -29,11 +29,24 @@ def process(self):
...
+                resp = requests.get(url=self.parameters.http_url, auth=self.auth,
+                                    proxies=self.proxy, headers=self.http_header,
+                                    verify=self.http_verify_cert,
+                                    cert=self.ssl_client_cert,
+                                    timeout=self.http_timeout_sec)
+
+            except requests.exceptions.Timeout:
+                timeoutretries += 1
+                self.logger.warn("Timeout whilst downloading the report.")
+
```
wagner-certat: The loop can be interrupted after the first successful run.

intelmq/bots/collectors/mail/collector_mail_url.py
```diff
@@ -46,11 +48,29 @@ def process(self):
...
+                                timeoutretries += 1
+                                self.logger.warn("Timeout whilst downloading the report.")
+
+                        if timeoutretries >= 3:
+                            self.logger.error("Request timed out three times in a row. ")
+                            erroneous = True
+                            # The download timed out too often, leave the Loop.
+                            continue
+
+                        self.logger.debug(resp.content)
```
dmth: uuuuh yes.

intelmq/bots/collectors/mail/collector_mail_url.py
```diff
@@ -46,11 +48,29 @@ def process(self):
...
+                                                    headers=self.http_header,
+                                                    verify=self.http_verify_cert,
+                                                    cert=self.ssl_client_cert,
+                                                    timeout=self.http_timeout_sec)
+
+                            except requests.exceptions.Timeout:
+                                timeoutretries += 1
+                                self.logger.warn("Timeout whilst downloading the report.")
+
+                        if timeoutretries >= 3:
```
dmth: Would be reasonable.

dmth:
To Merge this PR  intelmq/bots/collectors/http/collector_http_stream.py  and intelmq/bots/collectors/rt/collector_rt.py  need to be adapted to the new parameters

intelmq/bots/collectors/http/collector_http.py
```diff
@@ -29,11 +29,25 @@ def process(self):
...
 
-        resp = requests.get(url=self.parameters.http_url, auth=self.auth,
-                            proxies=self.proxy, headers=self.http_header,
-                            verify=self.http_verify_cert,
-                            cert=self.ssl_client_cert,
-                            timeout=self.http_timeout)
+        timeoutretries = 0
+        resp = None
+
+        while timeoutretries < self.http_timeout_max_tries and resp is None:
```
SYNchroACK: @wagner-certat can you validate this with me:

Is the Error Handling features documented [here](https://github.com/certtools/intelmq/blob/master/docs/User-Guide.md#error-handling) capable to do the same thing as this pull request want to do? I'm not 100% sure but it seems the same.

wagner-certat: Yeah, except that this is specific to timeouts. However, as the error_max_retries can be set to 3 for theses both collectors, I actually think that keeping the number of configuration parameters low is more important than handling the errors differently. So maybe just let the request.get raise the tiumeouterror?

wagner-certat:
> to fix the timeout issue there is only need to add 3 lines as far I can see:

Why do you catch and raise the exact same error at all?

SYNchroACK:
> Why do you catch and raise the exact same error at all?

:D:D:D:D heh I only added that just to have a specific message, if it's important but in practice, I agree with you @wagner-certat , there is no need to fix for that

dmth:
From my point of view it's better to have an explicit timeout behaviour than using the built in error handling.
`error_max_retries` depicts a behaviour which is occurring when the bot  fails due to a most likely internal error or because some data-format has changed.
`http_timeout_max_tries` on the other hand clearly states a behaviour which only occurs when talking to an external service.

Based on my experience users, and administrators as well, might want to differentiate this behaviour.


What might be a step to simplify configuration?
If `http_timeout_max_tries` is not set, use `error_max_retries` as fallback.


BTW: The Guide https://github.com/certtools/intelmq/blob/master/docs/User-Guide.md#error-handling is buggy:
> `error_max_retries` - in case of an error and the value of the `error_procedure` option is `retry` 

But: `error_procedure` does only list `stop` and `pass`

wagner-certat:
> BTW: The Guide https://github.com/certtools/intelmq/blob/master/docs/User-Guide.md#error-handling is buggy:
> > `error_max_retries` - in case of an error and the value of the `error_procedure` option is `retry` 
> 
> But: `error_procedure` does only list `stop` and `pass`

Thanks, fixed: 1f51da9dbe41657476c8aa55d10535a06e5133b3

swilde:
FWIW I'm with @dmth here:
problems preventing a download from finishing are quite common and often related to infrastructure not under control of the intelmq-admin.  Handling of such situations should be configurable with out much side effects.  Networking related errors are always something to expect, in contrast I'd consider `error_max_retries` a configuration option for handling of less expected errors.

dmth:
I've upgraded the documentation for the http_max_retries parameter. In addition I renamed the parameter for the stream collector in order to use the same name.
However, the stream collector does not honor the max_retries parameter.

wagner-certat:
This PR now proposes a mix of `http_timeout` and `http_timeout_sec`:

```python
> grep http_timeout * -r
bots/collectors/http/collector_http_stream.py:http_timeout: tuple of two floats or float
bots/collectors/http/collector_http_stream.py:                               timeout=self.http_timeout)
bots/collectors/http/collector_http.py:http_timeout: tuple of two floats or float
bots/collectors/http/collector_http.py:                                    timeout=self.http_timeout_sec)
Binary file bots/collectors/http/__pycache__/collector_http.cpython-34.pyc matches
Binary file bots/collectors/http/__pycache__/collector_http_stream.cpython-34.pyc matches
Binary file bots/collectors/rt/__pycache__/collector_rt.cpython-34.pyc matches
bots/collectors/rt/collector_rt.py:                                    timeout=self.http_timeout)
Binary file bots/collectors/mail/__pycache__/collector_mail_url.cpython-34.pyc matches
bots/collectors/mail/collector_mail_url.py:                                                    timeout=self.http_timeout_sec)
etc/defaults.conf:    "http_timeout_sec": 30,
lib/bot.py:        self.http_timeout_sec = getattr(self.parameters, 'http_timeout_sec', None)
```

dmth:
> This PR now proposes a mix of http_timeout and http_timeout_sec:

The mix is not intentional. I updated the RT collector and documentation correspondingly.

dmth:
@aaronkaplan  I really wonder why this cannot be merged...

 1. It provides a more clear naming for the timeout parameters (http_timeout_sec instead of http_timeout)
 2. It enables an administrator to control how often downloads are retried (http_timeout_max_tries)
 3. Experienced administrators told us, this is a reasonable way to go. (https://github.com/certtools/intelmq/pull/859#issuecomment-277277763)
 
From reviewing the latest commits,  it looks like `intelmq/bots/experts/ripencc_abuse_contact/expert.py` (a1a09ac67c025335ea203f3b79a639010abd6abd) and `intelmq/bots/outputs/restapi/output.py` (a918ec7ceeb2331dbf23d1b2cb39dd32776cc85e) would also require such renaming now.  Right now I'm not sure if I want to keep this PR up to date. It's a little bit wearing....

```
grep -r http_timeout **/*.py 
intelmq/bots/collectors/http/collector_http.py:    http_timeout_sec: tuple of two floats or float
intelmq/bots/collectors/http/collector_http.py:    http_timeout_max_tries: an integer depicting how often a connection attempt is retried
intelmq/bots/collectors/http/collector_http.py:        while timeoutretries < self.http_timeout_max_tries and resp is None:
intelmq/bots/collectors/http/collector_http.py:                                    timeout=self.http_timeout_sec)
intelmq/bots/collectors/http/collector_http.py:        if resp is None and timeoutretries >= self.http_timeout_max_tries:
intelmq/bots/collectors/http/collector_http_stream.py:    http_timeout_sec: tuple of two floats or float
intelmq/bots/collectors/http/collector_http_stream.py:                               timeout=self.http_timeout_sec)
intelmq/bots/collectors/mail/collector_mail_url.py:                        while timeoutretries < self.http_timeout_max_tries and resp is None:
intelmq/bots/collectors/mail/collector_mail_url.py:                                                    timeout=self.http_timeout_sec)
intelmq/bots/collectors/mail/collector_mail_url.py:                        if resp is None and timeoutretries >= self.http_timeout_max_tries:
intelmq/bots/collectors/rt/collector_rt.py:                                    timeout=self.http_timeout_sec)
intelmq/bots/experts/ripencc_abuse_contact/expert.py:                                timeout=self.http_timeout)
intelmq/bots/experts/ripencc_abuse_contact/expert.py:                                timeout=self.http_timeout)
intelmq/bots/experts/ripencc_abuse_contact/expert.py:                                timeout=self.http_timeout)
intelmq/bots/outputs/restapi/output.py:                              timeout=self.http_timeout,
intelmq/lib/bot.py:        self.http_timeout_sec = getattr(self.parameters, 'http_timeout_sec', None)
intelmq/lib/bot.py:        self.http_timeout_max_tries = getattr(self.parameters, 'http_timeout_max_tries', 1)
intelmq/lib/bot.py:        self.http_timeout_max_tries = self.http_timeout_max_tries if self.http_timeout_max_tries >= 1 else 1
```

intelmq/bots/collectors/http/collector_http.py
aaronkaplan: +1

wagner-certat:
>  This PR fixes #854 and #666 

This PR will not fix #666

wagner-certat:
Would be great if someone from @Intevation could check my changes. I'll then merge it.

dmth:
> Would be great if someone from @Intevation could check my changes. I'll then merge it.

Thanks for enhancing the code. Looks good, please go ahead.

Issue #653, opened 2016-08-29
gethostbyname Bot - improvements


SYNchroACK:
I think make sense to change the gethostbyname bot code like the following:

```
       ......

            if event.contains(key_ip):
                continue
            try:
                ip = socket.gethostbyname(event.get(key_fqdn))
                if intelmq.lib.harmonization.IPAddress.is_valid(ip, sanitize=True):
                    event.add(key_ip, ip, sanitize=True)
            except socket.gaierror:
                pass

        self.send_message(event)
        self.acknowledge_message()

       ......
```

What do you guys think? No point to get errors from fqdn's that doesn't translate to an IP.

sebix:
Should we ignore all errors? E.g. EAI_MEMORY is a valid one.

See https://github.com/python/cpython/blob/master/Modules/socketmodule.c#L7549 for a full list.

proposal:

 - `EAI_ADDRFAMILY`: raise
 - `EAI_AGAIN`: raise
 - `EAI_BADFLAGS`: raise
 - `EAI_FAIL`: pass (no record)
 - `EAI_FAMILY`: raise
 - `EAI_MEMORY`: raise
 - `EAI_NODATA`: pass (no record)
 - `EAI_NONAME`: pass (no record)
 - `EAI_OVERFLOW`: raise
 - `EAI_SERVICE`: raise
 - `EAI_SOCKTYPE`: raise
 - `EAI_SYSTEM`: pass (no record)

But as I do not have much experience with sockets, please correct me if I'm wrong :)

kralca:
socket.gaierror is raised when the name resolution fails (which @SYNchroACK 's proposed code handles). Without the check, any attempt to resolve a domain that cannot be resolved causes the bot to crash. I'd suggest implementing @SYNchroACK's proposed solution in the first instance as this is the most likely error to occur.

sebix:
@kralca There are different kinds of [gaierror](https://docs.python.org/3/library/socket.html?highlight=socket.gaierror#socket.gaierror). IMHO we should not treat a non-existing record the same way as a problem in our side (implementation, library, setup, etc.).

SYNchroACK:
@sebix cool. In your perspective, how should be treat in this case?

sebix:
@SYNchroACK above I listed all possible errors and how we could treat them (raising an error or treating as normal answer). Please review this list if you have more knowledge of sockets and dns answer codes.

wagner-certat:
Still waiting for feedback here.

[...]

PR #502, opened 2016-04-29
fqdn2ip expert bot

slashtolau:
fqdn2ip expert bot to resolve ip from fqdn where no ip field exists.

url2fqdn was used as a template for this bot.

PR #865, opened 2017-01-30
Treat non existing records in GethostbynameExpert

wagner-certat:
fixes certtools/intelmq#653


wagner-certat:
> from my point of view would make sense to call this bot FQDNtoIP or Host2IP.

Also See: https://github.com/certtools/intelmq/pull/502#r62932343

SYNchroACK:
> Also See: #502 (comment)

If was already discussed, I will not start that discussion again. Thats ok.
Thank you for point it out.

Issue #864, opened 2017-01-30
Shadowserver Open-LDAP key mapping


dmth:
Whilst looking at shadowserver's Open-LDAP we discovered that the mapping  `dns_host_name` to `destination.local_hostname` might not be optimal. `source.local_hostname` would be a better choice for this feed.

https://github.com/certtools/intelmq/blob/master/intelmq/bots/parsers/shadowserver/config.py#L1171

We are going to provide a PR for this.

Slightly relates to #833

wagner-certat:
You are right, this should be attributed to the source.

dmth:
Migrations required for installations of IntelMQ after this Change:

1) Modify EventDB in order to move all events which were classified the old way.
2) Update Filter-Experts which depend on this Key.

We are also going to provide a statement for 1)

swilde:
You can fix existing event databases with this statement:
```
UPDATE events
   SET "source.local_hostname"="destination.local_hostname",
       "destination.local_hostname"=DEFAULT
 WHERE "feed.name"='Open-LDAP' AND "source.local_hostname" IS NULL;
```

PR #866, opened 2017-01-30
Changed keymapping in Shadowserver Open-LDAP config

dmth:
This fixes https://github.com/certtools/intelmq/issues/864 by moving "destination" to "source".

# Required Migrations

**Only necessary if you are running a setup that is already using the Shadowserver Open-LDAP, or has been using this feed**

1) This commit will require migrations in the EventDB, an example SQL-statement will be provided in issue #864, soon
2) If you are using Filter-Experts looking for the key `destination.local_hostname`, you need to update them to `source.local_hostname`


SYNchroACK:
Thank you @dmth :)

dmth:
https://github.com/certtools/intelmq/issues/864#issuecomment-276119697 contains instructions for the migration which is required.

Issue #357, opened 2015-10-06
create an example *.conf setup which works out of the box


aaronkaplan:
We need an example config setup which works just straight out of the box.
I'll add it to an example/ subdirectory.

SYNchroACK:
The following information represents an example for runtime configuration:

**Runtime configuration**
* * *
**File location in repository:** intelmq/conf/examples/runtime.conf
**File location in installation folder:** /opt/intelmq/conf/examples/runtime.conf

Is it ok?

aaronkaplan:
important for usability.
The current example set is quite small. Need to include at least 5 OSINT feeds.

PR #867, opened 2017-01-30
Provide bigger & diversified example config

wagner-certat:
This extends the current example configuration. Now 4 different free feeds are configured: Abuse.ch, Spamhaus, malware domain list (was already default) and malc0de.
Also, some very useful experts: deduplicator, taxonomy, url2fqdn, gethostbyname, taxonomy and cymru.
The only output is still the file, all other would need proper configurations.
Additionally I added two instances for the gethostbyname (because of the lookups) to demonstrate a useful example to load balancing.

fixes certtools/intelmq#357

### DIFF
--- a/intelmq/etc/pipeline.conf
+++ b/intelmq/etc/pipeline.conf
@@ -1,6 +1,52 @@
 {
+    "abusech-domain-parser": {
+        "destination-queues": [
+            "deduplicator-expert-queue"
+        ],
+        "source-queue": "abusech-domain-parser-queue"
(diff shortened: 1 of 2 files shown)


SYNchroACK:
@wagner-certat everything good and Thank you for this work. Only think that I would suggest you to change is the bot-ids of gethostbyname bots. For clarification sake, please use the following two IDs:
* gethostbyname-1-expert
* gethostbyname-2-expert

Cheers

Issue #807, opened 2016-12-07
IntelMQ wiki install and dev guides now HTTP 404


mars01:
Are there new links by any chance?

Thanks!

sebix:
Which links do you mean?

sykaeh:
I think @espressobeanie means the links on the main page of the project. If you click on the links on https://github.com/certtools/intelmq then you get 404 Errors:

https://github.com/certtools/intelmq/blob/master/User-Guide.md
https://github.com/certtools/intelmq/blob/master/Developers-Guide.md
https://github.com/certtools/intelmq/blob/master/Data-Harmonization.md

This happens because the README is actually in the docs folder and not in the root directory of the repo.

sebix:
Ah, this is a result of #806 

Any ideas how to solve this without having two almost identical documents?

SYNchroACK:
(sharing my notes after my review on this)

**Read the Docs**
Issue: main page [here](http://intelmq.readthedocs.io/en/latest/) does not exists. Currently we are using docs/README.rst and all other docs are in markdown format.

**pypi**
Issue: links related to User-Guide, etc are broken. Check [here](https://pypi.python.org/pypi/intelmq/1.0.0.dev6)

wagner-certat:
> **Read the Docs**
Issue: main page [here](http://intelmq.readthedocs.io/en/latest/) does not exists. Currently we are using docs/README.rst and all other docs are in markdown format.

I wasn't aware of this Read the docs page, who's maintaining this? A conversion using pandoc before pushing fixes this easily.
The Readme is in RST because pypi does not understand markdown.

Concerning the links on pypi: I will fix this by replacing the links during runtime in setup.py.

SYNchroACK:
> I wasn't aware of this Read the docs page, who's maintaining this?

@aaronkaplan

> A conversion using pandoc before pushing fixes this easily.
> The Readme is in RST because pypi does not understand markdown.
> 
> Concerning the links on pypi: I will fix this by replacing the links during runtime in setup.py.

Cool.

PR #872, opened 2017-01-30
Documentation: Fixed links and docs for intelmqctl

wagner-certat:
This will fix the links in the README, #807 

Also adds a new file explaining the behavior of current intelmqctl with examples.


wagner-certat:
Two of the tests fail because of temporary connection failures, nothing relevant.

PR #875, opened 2017-02-01
Deduplicator Bot with dedup modes

SYNchroACK:
# Deduplicator

Bot responsible for ignore duplicated messages. The bot can be configured to perform deduplication just looking to specific fields on the message using one of two modes: whitelist / blacklist. 

## Parameters for "fine-grained" deduplication

* `filter_type`: type of the filtering ("blacklist", "whitelist") which will be applied
using the parameter `filter_keys` to generate an hash without fields that are not important
to detect duplication.
* `filter_keys`: string with multiple keys separated by comma. Please note that `time.observation` key will not be considered even if defined, because the system always ignore that key.

#### Example 1

The bot with this configuration will detect duplication only based on `source.ip` and `destination.ip` keys.

```
"parameters": {
    ...
    "filter_type": "whitelist",
    "filter_keys": "source.ip,destination.ip",
}
```

#### Example 2

The bot with this configuration will detect duplication based on all keys, except `source.ip` and `destination.ip` keys.

```
"parameters": {
    ...
    "filter_type": "blacklist",
    "filter_keys": "source.ip,destination.ip",
}
```


bernhard-herzog:
I think the documentation of `filter_type` should be improved. I was confused about its purpose until I looked at the implementation. How about:

 * `filter_type` ("blacklist", "whitelist"): How to interpret `filter_keys` when deciding whether an event has already been seen. If "whitelist", only the fields listed in `filter_keys` will be considered. If "blacklist" all fields *except* those in  `filter_keys` will be considered. Note that `time.observation` will never be considered.

SYNchroACK:
@bernhard-herzog check the new update please. Thank you for your feedback! :)

intelmq/lib/message.py
```diff
@@ -304,17 +304,32 @@ def __get_type_config(self, key):
     def __hash__(self):
         return int(self.hash(), 16)
 
-    def hash(self, blacklist=frozenset()):
+    def hash(self, filter_type="blacklist", filter_keys_list=frozenset()):
```
wagner-certat: You are swapping the arguments here while they can be used positional.

intelmq/lib/message.py
```diff
@@ -304,17 +304,32 @@ def __get_type_config(self, key):
...
+        filter_keys_list parameter will be considered as the keys to ignore
+        or the only ones to consider. If given, the filter_keys_list
         parameter should be a set.
         """
+
+        if filter_type not in ["whitelist", "blacklist"]:
+
+            raise exceptions.InvalidArgument('filter_type',
+                                             got=filter_type,
+                                             expected='<whitelist> or'
```
wagner-certat: Should be a list of values: `['whitelist', 'blacklist']`, the formatting is done by InvalidArgument anyway.

intelmq/lib/message.py
```diff
@@ -304,17 +304,32 @@ def __get_type_config(self, key):
...
-    def hash(self, blacklist=frozenset()):
+    def hash(self, filter_type="blacklist", filter_keys_list=frozenset()):
         """Return a sha256 hash of the message as a hexadecimal string.
-        The hash is computed over almost all key/value pairs. The only
-        keys omitted are 'time.observation' and all keys contained in
-        the optional blacklist parameter. If given, the blacklist
+        The hash is computed over almost all key/value pairs. Depending on
+        filter_type parameter (blacklist or whitelist), the keys defined in
+        filter_keys_list parameter will be considered as the keys to ignore
+        or the only ones to consider. If given, the filter_keys_list
```
wagner-certat: You removed the note that `time.observation` will be always ignored.

intelmq/bots/experts/deduplicator/README.md
```diff
@@ -1,3 +1,53 @@
...
+
+### Parameters for connecting to cache
+* `redis_cache_host` - redis host to connect (e.g. `127.0.0.1`)
+* `redis_cache_db` - redis db to connect (e.g. `6`)
+* `redis_cache_ttl` - ttl (in seconds) for each entry inserted on cache (e.g. `86400`)
+* `redis_cache_password` - password to access redis (by default is None)
+
+### Parameters for "fine-grained" deduplication
+
+* `filter_type`: type of the filtering which can be "blacklist" or "whitelist". The filter type will be used to define how deduplicator bot will interpret the the parameter `filter_keys` in order to  decide whether an event has already been seen or not, i.e., duplicated event or a completely new event.
```
wagner-certat: double whitespace `order to  decide whether`

intelmq/bots/experts/deduplicator/expert.py
```diff
@@ -14,13 +32,14 @@ def init(self):
                            getattr(self.parameters, "redis_cache_password",
                                    None)
                            )
-        self.ignore_keys = set(k.strip() for k in
-                               self.parameters.ignore_keys.split(','))
+        self.filter_keys_list = set(k.strip() for k in
```
wagner-certat: it's not a list, it's a set^^

CHANGELOG.md
```diff
@@ -35,6 +35,8 @@ v1.0 (in development, master branch)
 - ENH: New parameter and field named accuracy to represent the accuracy of each feed
 - Consistent naming "overwrite" to switch overwriting capabilities of bots (as opposed to override)
 - Renamed http_ssl_proxy to https_proxy
+- deduplicator bot has a new parameter to configure deduplication mode
```
dmth: and how is this parameter called?

wagner-certat: Thanks, fixed in #883

Issue #879, opened 2017-02-09
XMPP Bots: Log Messages are misleading


dmth:
The XMPP - Bots (Collector and Output) both log a Message `Connected to Server.`,  even if sleekxmpp is not capable of connecting to the XMPP Server.

This is Misleading.

I'm going to propose a fix for this.

wagner-certat:
Fixed by #880

PR #880, opened 2017-02-09
ENH: logging in XMPP Bots

dmth:
https://github.com/certtools/intelmq/issues/879

More clear logging  messages. In addition the XMPP Bots stop, if a connection cannot be established.

### DIFF
--- a/intelmq/bots/collectors/xmpp/collector.py
+++ b/intelmq/bots/collectors/xmpp/collector.py
@@ -132,14 +132,17 @@ class XMPPCollectorBot(CollectorBot):
             # Set CA-Certificates
             self.xmpp.ca_certs = ca_certs
 
-        self.xmpp.connect(reattempt=True)
-        self.xmpp.process()
+        if self.xmpp.connect(reattempt=False):
+            self.xmpp.process()
+            # Add Handlers and register Plugins
+            self.xmpp.register_plugin('xep_0030')  # Service Discovery
+            self.xmpp.register_plugin('xep_0045')  # Multi-User Chat
 
-        # Add Handlers and register Plugins
-        self.xmpp.register_plugin('xep_0030')  # Service Discovery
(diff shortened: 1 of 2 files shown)


intelmq/bots/collectors/xmpp/collector.py
```diff
@@ -185,9 +190,12 @@ def log_message(self, msg):
...
         # now it's up to a parser to do the interpretation of the message.
         if raw_msg:
-            report = self.new_report()
-            report.add("raw", raw_msg)
-            self.send_message(report)
+            try:
+                report = self.new_report()
+                report.add("raw", raw_msg)
+                self.send_message(report)
+            except:
```
wagner-certat: Did you see a case where this happened? Which Error is thrown? Please specify it.

dmth: For instance: if the "strip_message" parameter is not set and the message is like:
```
 {...normalEventAsJSon}
^"Space-Character"
```

wagner-certat: This does not trigger the error.

However, if there's any Exception inside this function, sleekxmpp will be kicked from the server because it would send something weird (probably the traceback?). For (better) error messages and debugging, I've put the whole function into an try-except block to log these exceptions.

wagner-certat:
@dmth Could you check my changes? Then I'll merge this.

wagner-certat:
See also f6e0b43688be0c2e155175a988cf61937706d760 which stops the threads correctly (otherwise we have multiple xmpp threads running and message are sent multiple times).

dmth:
https://github.com/certtools/intelmq/pull/880#issuecomment-279487337
I've had a look at your commit. Seems fine for. But I didn't test it yet.

dmth:
Hi, I did a quick function test today. Collectors are capable to collect, outputs are capable to send, works as expected.

dmth:
Feel free to merges this code.

PR #884, opened 2017-02-14
Changelog and NEWS

### LINKED ISSUE #883: Add a NEWS file
# Add a NEWS file...
**...containing a list of changes and Instructions on necessary migrations or adaptions to the configuration**

This is necessary for operators/administrators in order to become aware of breaking changes, such as:

* https://github.com/certtools/intelmq/pull/882

or new features or changes to the DHO like:

* https://github.com/certtools/intelmq/pull/861
* https://github.com/certtools/intelmq/pull/866

wagner-certat:
Extended the changelog and added a NEWS file

I digged through diffs and commit messages to hopefully get most of the changes. Please speak up if something is missing.

Fixes #883


dmth:
Right now I expect that a lot of IntelMQ's exist, which do not need the full set of statements mentioned in the NEWS file. It's hard to figure out what's required to change in this case. Normally this might be done using Tags/Versions but that's not possible yet. We might use dates, though, for example:


## Since 2017-01-18:
When running an IntelMQ older than this date, you need to do: FOO,
The Feature BAZ is supported again.

## since 2016-12-12
When running an IntelMQ older than this date, you need to do: BAR

## since 2016-12-11
When running an IntelMQ older than this date, you need to do: BAZ
**EDIT on 2017-01-18**: When upgrading to a Version newer than 2017-01-18 this feature is supported again.

wagner-certat:
>  This assumes that they are all running the same revision from timestamp X. This is not the case in IntelMQ.

It's the last stable version, i.e. 0.9. All version inbetween are development versions.

dmth:
> It's the last stable version, i.e. 0.9. All version inbetween are development versions.

I have a very bad feeling about this, as 0.9 is from 21 Jun 2015. I don't think that anyone is still running this.

It's OK to mention the DEV Releases in the Changelogs. They are fixed points in time and can be used for reference. This allows an administrator a much better view.

wagner-certat:
I have split the upgrade instructions into sections (one per dev version).

dmth:
Thank's for your work, the changelog file is still missing, though.

From my point of view it's good enough right now, but you need to state more clear which changes happend in which revisions (so maybe just alter the header in the changelog from "v1.0 (in development, master branch)" to "Between 0.9 and 1.0.0.dev6"

wagner-certat:
Merging because of no negative (blocking) feedback.

Issue #857, opened 2017-01-25
Decide on docstring style for sphinx-apidoc / autodoc


bernhardreiter:
Intelmq is implemented in python3 thus I consider it agreeable that we use www.sphinx-doc.org as processor to generate our code documentation. (As opposed to the overview documentation which is independent and currently done in some Markdown variant.)

There seem be several styles for docstrings around (see http://www.sphinx-doc.org/en/stable/ext/autodoc.html#module-sphinx.ext.autodoc)
 * reStructuredText (the default of sphinx)
 * NumPy style
 * Google style

http://www.sphinx-doc.org/en/stable/ext/napoleon.html#module-sphinx.ext.napoleon is the sphinx plugin adding the two additional styles and comparing them. As we are python3, **my proposal is:**

 * Declare **module-sphinx.ext.napoleon "google style"** the default autodoc style for intelmq
 * Declare **python3 type annotations** as the way types SHOULD be documented

bernhardreiter:
Up to now there are not many autodoc docstrings in the codebase (I have looked at a small sample of files). However it has been discussed as a way to document bots in the future in the comments of https://github.com/certtools/intelmq/pull/706 .

wagner-certat:
I guess this is the doumentation for the Goolge style? https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments

I'm fine with porting the docstrings to this style and then generate a proper configuration using sphinx.

Yet I used the numpy style for the libs because I know it (I simply used it in previous projects related to numpy/scipy).

bernhardreiter:
@wagner-certat this is the link to the general description. While we shall more specifically use "sphinx-napoleon-google-type-annotation" style as described in: http://www.sphinx-doc.org/en/stable/ext/napoleon.html#type-annotations

(The first part of the description you've linked is a python standard recommendation anyway.)

I don't know how many docstrings we'll have to convert, but we'll certainly should say which is the prefered style (and put this in the right place for coding conventions).

What do others think?

dmth:
I don't have a favourite style, but I concur we should agree on one style, and define this in the Contributing-Guidelines.

If in doubt, we should stick to the style which is already used (it's numpy, isn't it?)

wagner-certat:
> If in doubt, we should stick to the style which is already used (it's numpy, isn't it?)

It's numpy yes, but it doesn't matter. There are only a few docstrings, which can easily be converted.

[...]

PR #887, opened 2017-02-15
Code documentation style converted

wagner-certat:
See #857 

All docstrings are using the "sphinx-napoleon-google-type-annotation" style as described in: http://www.sphinx-doc.org/en/stable/ext/napoleon.html#type-annotations.
Type annotations are also added for functions with (converted) docstrings.

@bernhardreiter please review


bernhardreiter:
Just took a few minutes and browsed the code, all section I've looked at looked good and improved. Of course I could not do a full check nor did I check that all comments have been converted. 

Next steps would be to run Python3 on it, run a sphinx on it, run a type checker on it. And check if there are other comments which may need conversion. 

However it seems I won't be able to do this in the next days, thus I recommend committing anyway if someone can confirm that the code still run, because the situation then could have only improved.

PR #891, opened 2017-02-21
New expert 'field reducer'

wagner-certat:
This bot deletes fields from events, either using a whitelist or a blacklist.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ v1.0 (in development, master branch)
 - ENH: added bots.parsers.spamhaus.parser_cert
 - ENH: added bots.parsers.fraunhofer.parser_dga
 - ENH: added bots.experts.certat_contact.expert
+- ENH: added bots.experts.reducer.expert
 - MAINT: renamed bots.parsers.spamhaus.parser to bots.parsers.spamhaus.parser_drop
 - Dropped dragon research group feeds: discontinued
 - changed configuration syntax for bots.experts.modify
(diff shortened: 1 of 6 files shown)


wagner-certat:
Build fails because a apt mirror is down :/

dmth:
Why do you need this?

wagner-certat:
We will send data to a contact via REST. But we do not send all data of course, just some fields which are shareable.

intelmq/bots/experts/reducer/expert.py
```diff
@@ -0,0 +1,38 @@
...
+        if isinstance(self.keys, str):
+            self.keys = [key.strip() for key in self.keys.split(',')]
+
+    def process(self):
+        event = self.receive_message()
+
+        if self.type == 'whitelist':
+            new_event = Event()
+            for key in self.keys:
+                new_event.add(key, event[key], sanitize=False)
```
aaronkaplan: we noticed: what if the the key is not in the (old) event?

aaronkaplan: maybe take a look at pythons' event.fromkeys(keys...)? (for the whitelist part at least)

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ v1.0 (in development, master branch)
 - ENH: added bots.parsers.spamhaus.parser_cert
 - ENH: added bots.parsers.fraunhofer.parser_dga
 - ENH: added bots.experts.certat_contact.expert
+- ENH: added bots.experts.reducer.expert
```
aaronkaplan: how about calling it "field_reducer"? Why?

Because a "reducer" sounds like the reducer part of map-reduce.
That's seomthing else.

However, TBD... not sure... anyway, not a big issue.

wagner-certat:
Renamed the bot to *Field Reducer*

PR #894, opened 2017-02-27
Idea export and separate files output bot

pharook:
Idea expert bot + bot for saving events into separate files (based on discussions from IHAP in Prague).

Idea expert bot is in fact export - it transforms IntelMQ event data into [Idea](https://idea.cesnet.cz/) format in "output" field (also added into Harmonization).

Files output bot saves incoming events into separate files in safe manner - create file in temporary directory and atomically move into destination dir. It tries to play well with the rest of the system, so it can be used to export IntelMQ events (supports hierarchical_output, just like existing File output bot), however allows to be configured to save specific key, like, say, "output". :)

I tried to write at least basic run/conversion/comparison unit test for both of them.


intelmq/bots/experts/idea/expert.py
```diff
@@ -0,0 +1,186 @@
+# -*- coding: utf-8 -*-
+from collections import Sequence, Mapping
+from base64 import b64decode
+from uuid import uuid4
+from urllib.parse import quote_plus
+
+from intelmq.lib.bot import Bot
+from intelmq.lib.message import Event
```
wagner-certat: imported but unused

intelmq/bots/experts/idea/expert.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+def quot(s):
+    return quote_plus(s or "")
```
wagner-certat: uncovered by the testcase

intelmq/bots/experts/idea/expert.py
```diff
@@ -0,0 +1,186 @@
...
+
+    type_to_source_type = {
+        # Added nonstandard Dropzone, MalwareConf, will consider adding to Idea spec
+
+        "phishing": "Phishing",
+        "dropzone": "Dropzone",
+        "botnet drone": "Botnet",
+        "malware configuration": "MalwareConf",
+        "c&c": "CC"
+    }
```
wagner-certat: there are much more types, see harmonization.py for a complete list.

intelmq/bots/outputs/files/output.py
```diff
@@ -0,0 +1,77 @@
...
+        os.rename(path.join(self.tmp, tmpname), path.join(self.tmp, newname))
+        nf = io.open(fd, "w", encoding="utf-8")
+        return nf, newname
+
+    def process(self):
+        event = self.receive_message()
+        if self.parameters.single_key:
+            event_data = str(event.get(self.parameters.single_key))
+        else:
+            event_data = event.to_json(hierarchical=self.parameters.hierarchical_output)
```
wagner-certat: This line is untested.

intelmq/bots/BOTS
```diff
@@ -500,6 +500,13 @@
             "module": "intelmq.bots.experts.gethostbyname.expert",
             "parameters": {}
         },
+        "IDEA": {
+            "description": "Converts events into the IDEA format.",
+            "module": "intelmq.bots.experts.idea.expert",
+            "parameters": {
+                "test_mode": false
```
wagner-certat: AFAIU the test mode just changes the category to test. What do you need that for in this bot?

We have a test type and taxonomy for this purpose.

intelmq/bots/experts/idea/expert.py
pharook: Thx, fixed

intelmq/bots/experts/idea/expert.py
pharook: Ok, I've added some fields into test events

intelmq/bots/experts/idea/expert.py
pharook: Thx, I've missed 'dga domain', 'proxy' and 'other' (probably older version). I'm considering writing unit test, which would diff list in harmonization.py with translation in Idea translator, would that make sense?

intelmq/bots/experts/idea/expert.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def _init_translation(self):
```
pharook: Than there would be this big initialization, which would visually overwhelm the call to super(). Now the two distinct actions taken (data initialization and call to ancestor __init__) are clearly visible. However no strong opinion, do you insist? :)

intelmq/bots/experts/idea/expert.py
```diff
@@ -0,0 +1,186 @@
...
+            # extra - too informal, will consider based on real world data
+
+            "Format": lambda s: "IDEA0",
+            "Description": lambda s: "%s: %s" % (s["feed.name"], s.get("event_description.text", s.get("comment", s.get("classification.type", "unknown")))),
+            "Category": [lambda s: self.type_to_category[s.get("classification.type", "unknown")], lambda s: "Test" if self.parameters.test_mode else None],
+            "DetectTime": "time.observation",
+            "EventTime": "time.source",
+            "ID": lambda s: str(uuid4()),
+            "RelID": lambda s: "hash:" + quot(s["event_hash"]),
+            "Confidence": lambda s: min(1, max(0, s["feed.accuracy"]/100)),
```
pharook: Fixed (also in test cases, which are probably not covered by checker)

intelmq/bots/outputs/files/output.py
pharook: Fixed, however I'm not sure if it is clean enough. Deducing from code, cls.sysconfig is meant to be set in set_bot(), however then it would not be possible to test various configs (which is necessary here). Usually this is solved by another test class with different fixtures, however that is not possible in IntelMQ as it requires just one test class per bot, with exact name. So I resorted to setting self.sysconfig in each test.

intelmq/bots/BOTS
pharook: It does not change, it _appends_ additional category (in Idea there is possibility for event to have multiple categories). It is handy when setting up new client - all outgoing events are as they should, however they are also marked as "Test", so if they end up in database or recipients, everybody is able to easily filter them out.

pharook:
Regarding nonintuitive/long variable names: could you please be more specific?
Regarding documentation: is there also any policy or recommendation about where to put particular bot within its category? Some bots seem ordered, some not.

intelmq/bots/experts/idea/expert.py
wagner-certat: yes, we do net check the style for the tests currently

intelmq/bots/experts/idea/expert.py
wagner-certat: There is no need to call `super()` in `init()`

intelmq/bots/outputs/files/output.py
wagner-certat: that's fine

wagner-certat:
>  Regarding documentation: is there also any policy or recommendation about where to put particular bot within its category? Some bots seem ordered, some not.

Thanks, fixed in master

intelmq/bots/experts/idea/expert.py
wagner-certat: yes, then we are aware of this when we add a new type.

intelmq/bots/BOTS
wagner-certat: Personally I do not like this at all. With the same argument we can add such a test-flag to every bot.

Are there any other opinions?

wagner-certat:
> Regarding nonintuitive/long variable names: could you please be more specific?

Hm, not sure what I was thinking about.

intelmq/bots/BOTS
pharook: The additional Test category is the standard practice in Idea/Warden community (as are multiple categories). Is there any other more standard or general way to put in two categories in IntelMQ?

intelmq/bots/experts/idea/expert.py
pharook: Are you sure? AFAIU Bot.__init__ would not get called then, and there is all the initialization of logging and configuration in there...

intelmq/etc/harmonization.conf
```diff
@@ -338,6 +338,10 @@
         "time.source": {
             "description": "Time reported by a source. Some sources only report a date, which may be used here if there is no better observation.",
             "type": "DateTime"
+        },
+        "output": {
```
pharook: Forgot to note - also added.

intelmq/bots/experts/idea/expert.py
wagner-certat: No, it's not. That's done in `Bot.__init__`. `Bot.init` is "empty".

pharook: Ahha, I got misdirected by markdown reformatting of `__init__` and thought we're talking the same thing. Now I see your point, the Bot has the `init` method, which gets called during `__init__`. Ok, makes sense.

intelmq/bots/experts/idea/expert.py
pharook: Unit test added. Also I had to remove the "test" category - my translation list was from taxonomy expert bot, and its classification list is different than that in harmonization.py. Is that ok? Maybe taxonomy bot could use that unittest too? :)

intelmq/bots/experts/idea/expert.py
pharook: Done.

pharook:
Added documentation to docs/Bots.md.

pharook:
So the psql error in travis was also caused by me. :) Ok, I didn't know I'm gonna open even this can of worms. Well, there is a price in attempting to make things general (as discussed with Aaron - we wanted the export bot to be independent from Idea expert and usable separately for other data, even original IntelMQ events/reports, and that is doable only by passing translated data through another key, which in turn must be added to harmonization).
Anyway - I have no big experience with postgresql (we are not using this output bot), however I have set it up and run the unit test (succesfully) over old db schema, new db schema and altered (migrated) schema  before comitting.

wagner-certat:
> So the psql error in travis was also caused by me. :) Ok, I didn't know I'm gonna open even this can of worms.

We have some checks in place to ensure harmonization changes do not break anything, the change is "intentional" and other parts of intelmq can deal with it/do not conflict.

pharook:
> We have some checks in place to ensure harmonization changes do not break anything, the change is "intentional" and other parts of intelmq can deal with it/do not conflict.

Sure, I understand and generally I like the effort going into safety nets here.

intelmq/bots/BOTS
```diff
@@ -598,6 +605,17 @@
...
         },
+        "Files": {
+            "description": "Files is the bot responsible for lockfree saving of events into separate files.",
+            "module": "intelmq.bots.outputs.files.output",
+            "parameters": {
+                "tmp": "/opt/intelmq/var/lib/bots/files-output/tmp",
+                "dir": "/opt/intelmq/var/lib/bots/files-output/incoming",
+                "suffix": ".json",
+                "hierarchical_output": false,
+                "single_key": "output"
```
wagner-certat: I'm not sure if we want to have "output" as default, as opposed to all other output bots, which send/save the intelmq event as a whole. @aaronkaplan what do you think?

wagner-certat:
@aaronkaplan are the harmonization changes ok?

intelmq/bots/BOTS
pharook: Hmm, even I think you're right, I was trying to be consistent with 'file' bot in hierarchical_output setting, but I haven't noticed this. I'll update to null, so output defaults to the whole event.

pharook: Done.

intelmq/bots/experts/idea/expert.py
```diff
@@ -0,0 +1,199 @@
+# -*- coding: utf-8 -*-
+from collections import Sequence, Mapping
+from base64 import b64decode
+from uuid import uuid4
+from urllib.parse import quote_plus
+
+from intelmq.lib.bot import Bot
+
+
+def quot(s):
```
wagner-certat: missing e

pharook:
Can I do anything more to push this forward? :) ( @aaronkaplan ?)

intelmq/bots/experts/idea/expert.py
```diff
@@ -0,0 +1,199 @@
...
+    return quote_plus(s or "")
+
+
+def addr4(s):
```
aaronkaplan: please use the python 3 ip address type functions.
https://docs.python.org/3/howto/ipaddress.html#ipaddress-howto

intelmq/bots/experts/idea/expert.py
```diff
@@ -0,0 +1,199 @@
...
+    return s if ":" not in s else None
+
+
+def addr6(s):
+    return s if ":" in s else None
```
aaronkaplan: https://docs.python.org/3/howto/ipaddress.html#ipaddress-howto

intelmq/bots/outputs/files/output.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/python
```
aaronkaplan: Can we merge the files output bot with the existing file output bot?
The similar naming might be confusing.
When to use which one of the file output bots?

intelmq/etc/harmonization.conf
```diff
@@ -197,6 +197,10 @@
             "regex": "^[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[0-9a-z]{12}$",
             "type": "LowercaseString"
         },
+        "output": {
```
aaronkaplan: "output" is a very generic name for the field.
On the input side we have a field called "raw" .
I assume what you want here is a field called "transformed" or similar?

intelmq/bots/experts/idea/expert.py
wagner-certat: Note that this code does no validation, other parts of the code are resonsible of this. These small helper function only have the task to differentiate between the two version.

aaronkaplan: okay, then it does not matter.. ignore my comment in this case pls.

docs/Bots.md
```diff
@@ -712,6 +727,27 @@ FIXME
...
+* `cache (redis db):` none
+* `description:` saving of messages as separate files
+
+#### Configuration Parameters:
+
+* `dir`: output directory (default `/opt/intelmq/var/lib/bots/files-output/incoming`)
+* `tmp`: temporary directory (must reside on the same filesystem as `dir`) (default: `/opt/intelmq/var/lib/bots/files-output/tmp`)
+* `suffix`: extension of created files (default `.json`)
+* `hierarchical_output`: if `true`, use nested dictionaries; if `false`, use flat structure with dot separated keys (default)
+* `single_key`: if `none`, the whole event is saved (default); otherwise bot saves only contents of the specified key
```
aaronkaplan: grammar : "otherwise the bot ..."

intelmq/bots/outputs/files/output.py
wagner-certat: I also thought about this, but then decided not to raise this topic because:
They behave quite differently and have 2 sets of parameters. Merging them would mean the user has a longer list of parameters but there are two groups for the two modes which does not make the configuration easier but more complex.

intelmq/bots/experts/idea/expert.py
pharook: You sure? This is also helper function for translation definition, deliberately short (because used on various places), and deliberately misnamed to not get confused with urllib.parse.quote.

wagner-certat: That's ok for me too.

intelmq/bots/outputs/files/output.py
pharook: The thought behind naming was that "file" bot writes all event into one file, whereas "files" bot writes onto separate, distinct files. Anything different sounded long or weird, however no strong opinions, if there are any ideas. :)

Concerning merging - I'm a bit afraid merged bot would get hairy and even more confusing - now they are two, but with reasonably simple and understandable/reviewable code, and straightforward set of (some same, some distinct) configuration options.

intelmq/etc/harmonization.conf
pharook: Well, I used the name we were talking about at Prague. If you feel 'transformed' is more fitting, I can refactor. However I took this as a temporary solution until next version of IntelMQ anyway - as I understand it, there are plans of rearchitecting configuration/contracts, adding transformation class of bots, extending harmonization (which would allow describing foreign formats, like Idea).

PR #896, opened 2017-03-01
Strip API keys from events in parsers

### LINKED ISSUE #889: Strip API Keys from feed URLs in parsers
Follow up of #796 

There are some feeds which use API keys in URLs:

 - [x] ~~Alienvault OTX~~ (does not save the key at all)
 - [x] AnubisNetworks Cyberfeed
 - [x] MalwarePatrol Dans Guardian
 - [x] PhishTank

According to my colleagues, they should not end up in the events itself. Do you agree?

wagner-certat:
fixes #889 

As I also found a sample data of Malwarepatrol I added tests for this feed and revised the parser a bit.


intelmq/bots/parsers/anubisnetworks/parser.py
```diff
@@ -62,6 +62,7 @@ def process(self):
         raw_report = json.loads(utils.base64_decode(report.get('raw')))
         extra = {}
         event = self.new_event(report)
+        event.change("feed.url", event["feed.url"].split("?key=")[0])
```
aaronkaplan: hmmm... hacky. But I have no better solution

aaronkaplan:
looks ok.

PR #898, opened 2017-03-01
Enhance log handling in ctl

wagner-certat:
* The Stream-Logging (which is used by the intelmqctl tool too!) now uses stdout and sterr based on the log level instead of writing everything to stderr
  This allows now easier handling of the outputs (grepping etc)
* `intelmqctl log` can read from syslog


intelmq/lib/utils.py
```diff
@@ -193,6 +198,21 @@ def handleError(self, record):
             raise
 
 
+class StreamHandler(logging.StreamHandler):
+    def emit(self, record):
+        try:
+            msg = self.format(record)
+            if record.levelno < 30:  # Lower than warning (debug, info)
```
aaronkaplan: seems a bit hard-coded

PR #903, opened 2017-03-03
expose the '_origin' field from the feed so that we ...

aaronkaplan:
can filter out certain "_origins" from anubis later downstream in the pipeline.


intelmq/bots/parsers/anubisnetworks/parser.py
```diff
@@ -83,7 +82,7 @@ def process(self):
                 if "server_name" in value:
                     event.add('destination.fqdn', value["server_name"],
                               raise_failure=False)
-                for k in ["request_method", "cookies", "path_info", "http_referer"]:
+                for k in ["request_method", "cookies", "path_info", "http_referer", "_origin", "_provider", "pattern_verified"]:
```
wagner-certat: Why not `event_description.text`?

PR #906, opened 2017-03-06
BUG: anubis parser: Fix incomplete extra fields

wagner-certat:
Fixes the incomplete PR #903 #904 

`_origin`, `_provider` and `pattern_verified` are not in `env`.

For pattern_verified I had no sample to verify it.

And added tests for it.


aaronkaplan:
thx

PR #909, opened 2017-03-07
Shadowserver Accessible VNC parser

wagner-certat:
As of today shadowserver delivers a new feed with accessible VNC servers

### DIFF
--- a/intelmq/bots/parsers/shadowserver/config.py
+++ b/intelmq/bots/parsers/shadowserver/config.py
@@ -80,6 +80,7 @@ def get_feed(feedname):
         "Blacklisted-IP": blacklisted_ip,
         "Accessible-Telnet": accessible_telnet,
         "Accessible-CWMP": accessible_cwmp,
+        "Accessible-VNC": accessible_vnc,
     }
 
     return feed_idx.get(feedname)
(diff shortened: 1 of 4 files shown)


intelmq/bots/parsers/shadowserver/config.py
```diff
@@ -1280,3 +1281,28 @@ def validate_fqdn(value):
...
+        ('source.geolocation.city', 'city'),
+        ('extra.', 'naics', invalidate_zero),
+        ('extra.', 'sic', invalidate_zero),
+        ('extra.', 'banner', validate_to_none),
+        ('extra.', 'product', validate_to_none),
+    ],
+    'constant_fields': {
+        'protocol.transport': 'tcp',
+        'protocol.application': 'vnc',
+        'classification.type': 'vulnerable service',
```
aaronkaplan: actually this is a problem with all "openly accessible services":
  * for some we know that this is not a good idea / best practice and it should be not allowed (e.g. open recursive DNS) - since you can argue that it harms others.
  * some others are open  but "meh" not so pretty. But it's hard to argue that they are vulnerable. Do we know it? This might be the very latest and most current patched VNC server with password and a good crypto authentication library. We don't know. 

In short: we might have come up with a new category . Let's call it "accessible services" or  "potentially  vulnerable services" . This would also require new texts.

-> I think I want to discuss this with the team and community.

dmth: > In short: we might have come up with a new category . Let's call it "accessible services" or "potentially vulnerable services" . This would also require new texts.
>
> -> I think I want to discuss this with the team and community.

I agree on discussing this.

wagner-certat: This has been discussed, how do you want to proceed? Also, is that really blocking the merge of the code? IMHO the discussion is about "Do you send this out" and not "should we merge this parser".

Issue #793, opened 2016-11-24
Make the usage of extra field easier


sebix:
I just noticed that I haven't created an issue for this yet, had it in mind since the creation of this field...

It's easier to write e.g.:

```
event.add('extra.foo') = 'bar'
```

dmth:
Would be of great use in the Generic CSV parser in case a column cannot be mapped to the dho, but should be processed afterwards.

dmth:
> Would be of great use in the Generic CSV parser in case a column cannot be mapped to the dho, but should be processed afterwards.

Since the merge of https://github.com/certtools/intelmq/pull/935 this function is available for the generic csv-parser.

PR #913, opened 2017-03-09
Elasticsearch output bot

navtej:
An output bot for sending messges to elasticsearch instance. Since Elasticsearch does not allow "." in incoming data since 2.X, a configurable replace char can be supplied as parameter, default is _.


intelmq/bots/outputs/elastic/output.py
```diff
@@ -0,0 +1,78 @@
...
+            self.es.indices.create(index=self.elastic_index, ignore=400)
+
+    def process(self):
+        event = self.receive_message()
+        event_dict = event.to_dict(hierarchical=False)
+
+        for field in self.flatten_fields:
+            if field in event_dict:
+                val = event_dict[field]
+                # if it string try to convert to json
```
wagner-certat: *Only* do this for `extra` and nothing else.

See also #793

intelmq/bots/outputs/elastic/output.py
```diff
@@ -0,0 +1,78 @@
+# -*- coding: utf-8 -*-
+
+import sys
```
wagner-certat: unused import

intelmq/bots/outputs/elastic/output.py
```diff
@@ -0,0 +1,78 @@
...
+from json import loads
+from collections.abc import Mapping
+
+from elasticsearch import Elasticsearch
+
+from intelmq.lib.bot import Bot
+
+#
+# TODO: recursion depth check
```
wagner-certat: Do you mean checking for loops? That will never happen (and if, it should be prohibited by the core).

intelmq/bots/outputs/elastic/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+Bot parameters:
+
+* elastic_host       : Name/IP for the elasticsearch server, defaults to 127.0.0.1
+* elastic_port       : Port for the elasticsearch server, defaults to 9200
+* elastic_index      : Index for the Elasticsearch output, defaults to intelmq
+* elastic_doctype    : docname to put the event data, defaults to events
+* sanitize_keys      : true/false value to indicate if key names should be changed to remove '.', which is forbidden character in ES, defaults to true
+* replacement_char   : if sanitize_keys is true, which character should be used as replacement for '.', defaults to '_'
+* flatten_fields     : In ES, some query and aggrigations work better if the fields are flat and not nested. Here you can provide a list of fields to flatten out.
```
wagner-certat: Does it make sense to flatten all fields? (by default or with the special value `True`)

intelmq/bots/outputs/elastic/README.md
```diff
@@ -0,0 +1,15 @@
...
+### Output Bot that sends events to a Elasticsearch
+
+
+Bot parameters:
+
+* elastic_host       : Name/IP for the elasticsearch server, defaults to 127.0.0.1
+* elastic_port       : Port for the elasticsearch server, defaults to 9200
+* elastic_index      : Index for the Elasticsearch output, defaults to intelmq
+* elastic_doctype    : docname to put the event data, defaults to events
+* sanitize_keys      : true/false value to indicate if key names should be changed to remove '.', which is forbidden character in ES, defaults to true
```
wagner-certat: If it's forbidden anyway, why does the user have to option to run into problems?

intelmq/bots/outputs/elastic/README.md
```diff
@@ -0,0 +1,15 @@
+# Elasticsearch Output Bot
+
+### Output Bot that sends events to a Elasticsearch
+
+
+Bot parameters:
+
+* elastic_host       : Name/IP for the elasticsearch server, defaults to 127.0.0.1
+* elastic_port       : Port for the elasticsearch server, defaults to 9200
+* elastic_index      : Index for the Elasticsearch output, defaults to intelmq
```
wagner-certat: And it will be created if it does not exist yet.

intelmq/bots/outputs/elastic/output.py
```diff
@@ -0,0 +1,78 @@
...
+        for field in self.flatten_fields:
+            if field in event_dict:
+                val = event_dict[field]
+                # if it string try to convert to json
+                if isinstance(val, str):
+                    try:
+                        val = loads(val)
+                    except:
+                        pass
+                if isinstance(val, Mapping):
```
wagner-certat: The current harmonization does not allow Mappings like dictionaries, what do you intend to do here?

intelmq/bots/outputs/elastic/README.md
navtej: If this value is not provided, then the code doesnt do anything it gets the default value using getattr and if it doesnt exists it gets [], thus flattening field is skipped by looping over empty lis. However the BOT file has set the default field to extra. What do you suggest, remove the value from BOT file?

intelmq/bots/outputs/elastic/README.md
navtej: Old ES versions allowed it. I am not sure if someone even runes those versions, but you never know. Default is yes anyways.

intelmq/bots/outputs/elastic/README.md
navtej: Yes index will be created if it does not exists.

intelmq/bots/outputs/elastic/output.py
navtej: This is happening when flattening extra field. As if now the extra field comes as string and not dict. The flow here is :
0. get the event as dict
1. extract extra field
2. check if it is string
    2.1 try to load it as json
3. else check if it is mapping/dict
   3.1 if so being all embedded keys values to root.

We want to bring only embedded dict in extra to root level, because it will be a problem to handle other fields like strings and arrays, with a key. We cannot assume a key name and place the value there.

intelmq/bots/outputs/elastic/output.py
navtej: left it as configuration to end user, extra field is by default however if he wants it can do it for others as well. though i dont see the need as of now, however leaving a provision is not a bad idea imo.

intelmq/bots/outputs/elastic/output.py
navtej: removed the todo

intelmq/bots/outputs/elastic/output.py
navtej: removed

navtej:
For Auth ie username/password, ES itself doesnot provides any mechanism for auth, there are two ways to achieve this,
1/ use plugins, eg x-pack(commercial from elastic.co) and there is equivalent oss plugin.
2/ put es behind nginx reverse proxy.

Because ES doesnt provides a way, i have not put it there.
How do you suggest we proceed here?

intelmq/bots/outputs/elastic/README.md
wagner-certat: If the user needs to flatten everything, he needs to give *all* field names?

intelmq/bots/outputs/elastic/README.md
wagner-certat: ES < 2.0 had the EOL 2016-02-02. If upstream does not support it, wo don't need to, too.

intelmq/bots/outputs/elastic/output.py
aaronkaplan: @navtej let me ask, when you mean "flatten field"  do you mean to also "flatten" the "." (dot) away ?
I remember @SYNchroACK having done that in early days. Back then we used logstash to still re-write things.

Anyway, I'd like to know (maybe we can get some feedback from multiple users of ES on the mailing list) what format per se would be the most suitable for ES?

wagner-certat:
> ES itself doesnot provides any mechanism for auth

Dafuq? Ok, then leave it as is.

intelmq/bots/outputs/elastic/output.py
```diff
@@ -0,0 +1,73 @@
+# -*- coding: utf-8 -*-
+
+from json import loads
+from collections.abc import Mapping
+
+from elasticsearch import Elasticsearch
+
+from intelmq.lib.bot import Bot
+
+def replace_keys(obj, key_char='.', replacement='_'):
```
wagner-certat: minor pep8 issue: missing one line (should be two empty lines)

intelmq/bots/outputs/elastic/output.py
navtej: When i say flatten, in this context, it means to flatten the structure ie un-nest the nested fields.
Example, turn this
```
{
"a": "1",
"b": "2",
"c": {
      "d":"4",
      "e":"5"
      }
}
```
to 
```

{
"a": "1",
"b": "2",
"d":"4",
 "e":"5"
}
```

intelmq/bots/outputs/elastic/README.md
navtej: unfortunately yes, the assumption here is for the end user to put some effort and understand if he needs it. for some fields if he does not need to run aggregation type queries he does not need to flatten, however if he does need those for some stats etc, he has to be explicitly tell that these fields have value beyond search. Your thinking?

intelmq/bots/outputs/elastic/output.py
wagner-certat: What if the key already exists in the lower level? E.g. we have `source.ip` and `destination.ip`

navtej: in current setup those will be overwritten, since this is working on dict.

intelmq/bots/outputs/elastic/README.md
navtej: removed the optional sanitize_keys param

intelmq/bots/outputs/elastic/output.py
navtej: fixed. pep8 didnt complained on my setup though.

intelmq/bots/outputs/elastic/output.py
wagner-certat: You can use non-hierarchical messages ("c.d" etc), but this approach is a no-go.

wagner-certat:
I finally tested the code locally and fixed some minor things. Please have a look if you think those are ok.

However I failed to write a test case. Could you help here?
I started here: 2784743037a02e62b89b635a8554cc00ade0629a
I could not figure out how I could convert this query:
```bash
curl -XGET 'http://localhost:9200/intelmq/events/_search?q=source_asn:64496&pretty'
```
to the syntax required by `es.search()`. I want t check if the event is there and then delete it (Could be done in one step).

intelmq/bots/outputs/elastic/output.py
```diff
@@ -54,15 +54,15 @@ def process(self):
...
                 if isinstance(val, str):
                     try:
                         val = loads(val)
-                    except:
+                    except ValueError:
                         pass
                 if isinstance(val, Mapping):
                     for key, value in val.items():
-                        event_dict[key] = value
+                        event_dict[field + '_' + key] = value
```
navtej: I was thinking of taking a different approach viz allowing the option to overwrite keys or not.  If `overwrite_keys` is `True` then skip the key while flattening else overwrite. It is a matter of choice though, however the reason I favour this is purely aesthetic i.e a field named `time_observation` vs `extra_time_observation`.

Overwriting, at first thought, sounds like a bad idea however, considering most of the non-extra fields will anyways be filled with proper values, only extra data which does not fits into existing harmanization will be flattened, as such the probability of overwriting actually being hit will be very less, correct me if i am missing something.

wagner-certat: It's not only about overwriting, it's also the lost hierarchy, no valid harmonization, broken compatibility.

Once #793 is solved, this workaround can be removed anyway.

wagner-certat:
If you do not have any objections/comments, I will merge this soon.

navtej:
Test looks fine.  Wish I was able spend some time on it earlier.

PR #916, opened 2017-03-10
Shadowserver Compromised-Website: handle missing / in URL

antoinet:
Recently, some occurences of compromised-website reports were not sound: the 'url' value did not start with a /. As a consequence, the resulting 'source.url' field was erroneous, since it is the concatenation of 'hostname' + 'url'.

Example:
...,"hostname",...,"url",...
...,"example.com",...,"blip.php",...

results in: "http://example.comblip.php" instead of "http://example.com/blip.php"

### DIFF
--- a/intelmq/tests/bots/parsers/shadowserver/test_compromised_website.py
+++ b/intelmq/tests/bots/parsers/shadowserver/test_compromised_website.py
@@ -2,6 +2,7 @@
 
 import os
 import unittest
+import copy
 
 import intelmq.lib.test as test
 import intelmq.lib.utils as utils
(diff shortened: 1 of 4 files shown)


intelmq/tests/bots/parsers/shadowserver/test_compromised_website.py
```diff
@@ -44,6 +45,9 @@
        'time.observation': '2015-01-01T00:00:00+00:00',
        'time.source': '2017-01-16T00:43:48+00:00'}
 
+EVENT01 = copy.copy(EVENT00)
```
wagner-certat: Just use `EVENT00.copy()` and remove the unnecessary import.

SYNchroACK: @sebix EVENT00 is not an Event object. Should we convert it to Event obj before copy?

wagner-certat: @SYNchroACK That's not necessary. It will be converted to an Event object in the tests anyway.

wagner-certat: Fixed it myself now.

PR #918, opened 2017-03-13
Reduce error messages: Exiting is not an error

wagner-certat:
In case of an error, it will be logged as error anyway.
But a regular kill is not an error.


aaronkaplan:
works for me, pls. merge

Issue #919, opened 2017-03-13
Tor expert bot: handle existing field


wagner-certat:
If the field `tor_node` exists, the bot will throw an error.

PR #920, opened 2017-03-13
tor_nodes expert bot handle existing fields

wagner-certat:
fixes certtools/intelmq#919


aaronkaplan:
looks good, can merge. Thx.

PR #922, opened 2017-03-14
Blueliv Crimeserver add some additional data in extra

navtej:
Adds,
status, updated time, last seen time, confidence in extra field


navtej:
@SYNchroACK can we extend this discussion related to time. Should we expand Harmonization to include these fields:

time.observation : when intelmq processed the data
time.source : time assigned by source/ created time by source.
time.updated : when the information was last updated
time.first_seen: first seen as provided by source
time.last_seen: last seen as provided by source

navtej:
@wagner-certat I have updated the test however CI still fails, here is the output from my local setup
`root~/intelmq/intelmq/intelmq (blueliv-parser-enh)# python3 -m unittest tests/bots/parsers/blueliv/test_parser_crimeserver.py
..
Ran 2 tests in 0.224s

OK
`

wagner-certat:
> @wagner-certat I have updated the test however CI still fails, here is the output from my local setup

a temporary lookup failure

> I would say that time.source should be the value of item[lastSeenAt]. Is this make any sense for you guys?

yes, fixed in cfbb2e6

> time.observation : when intelmq processed the data

yes
> time.source : time assigned by source/ created time by source.

latests time of occurrence as reported by the source
> time.updated : when the information was last updated

not relevant, probably in extra
> time.first_seen: first seen as provided by source

could be done in extra, not really relevant
> time.last_seen: last seen as provided by source

same as time.source

PR #925, opened 2017-03-14
Better handle misconfigurations by stopping the bot

wagner-certat:
Two bugs are fixed:

* Handle non existing src or dest queues
  Currently, a bot without destination queue is retrying forever, now it stops immediately with a ConfigurationError
* bots handle invalid keys in incoming messages
  If an incoming message does contain an key/value which is not valid, we are retrying too, now it stops immediately with a ConfigurationError

### DIFF
--- a/intelmq/bin/intelmqctl.py
+++ b/intelmq/bin/intelmqctl.py
@@ -653,7 +653,7 @@ Outputs are additionally logged to /opt/intelmq/var/log/intelmqctl'''
     def list_queues(self):
         source_queues, destination_queues, internal_queues, all_queues = self.get_queues()
         pipeline = PipelineFactory.create(self.parameters)
-        pipeline.set_queues(source_queues, "source")
+        pipeline.set_queues(None, "source")
         pipeline.connect()
 
         counters = pipeline.count_queued_messages(*all_queues)
--- a/intelmq/lib/bot.py
+++ b/intelmq/lib/bot.py
@@ -352,8 +352,13 @@ class Bot(object):
         # handle a sighup which happened during blocking read
         self.__handle_sighup()
 
-        self.__current_message = libmessage.MessageFactory.unserialize(message,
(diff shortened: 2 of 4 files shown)


wagner-certat:
continuous-integration/travis-ci/push fails because of a lookup error

intelmq/bin/intelmqctl.py
```diff
@@ -653,7 +653,7 @@ def get_queues(self):
     def list_queues(self):
         source_queues, destination_queues, internal_queues, all_queues = self.get_queues()
         pipeline = PipelineFactory.create(self.parameters)
-        pipeline.set_queues(source_queues, "source")
+        pipeline.set_queues(None, "source")
```
SYNchroACK: did not understood this line

wagner-certat: We need to set at least a dummy value here, otherwise we would fail later.

intelmq/lib/bot.py
```diff
@@ -343,8 +352,13 @@ def receive_message(self):
...
 
-        self.__current_message = libmessage.MessageFactory.unserialize(message,
-                                                                       harmonization=self.harmonization)
+        try:
+            self.__current_message = libmessage.MessageFactory.unserialize(message,
+                                                                           harmonization=self.harmonization)
+        except exceptions.InvalidKey as exc:
+            # In case a incoming message is malformed an does not conform with the currently
+            # loaded harmonization, stop now as this will happen repeatedly without any change
+            raise exceptions.ConfigurationError('harmonization', exc.args[0])
```
aaronkaplan: Which bots breaks now because of this ?

wagner-certat: This is a better error handling (clear message for the user) in case of misconfigurations.

wagner-certat:
No feedback - merged

PR #926, opened 2017-03-14
BUG+TST: complete types list in taxonomy expert

wagner-certat:
New category `test` to be consistent and add test to ensure this consistence.

thanks to @pharook for the hint


aaronkaplan:
looks good, let's merge?

PR #931, opened 2017-03-27
RT collector: error handling and time filtering

wagner-certat:
* If a resource needs to be fetched via download and the resource is not available, we now skip the creation of a report. (bugfix) Also, the ticket status will be changed according to the configuration
* Allows optional time-based searches, both absolute and relative, see the added docs.

### DIFF
--- a/intelmq/bots/collectors/rt/collector_rt.py
+++ b/intelmq/bots/collectors/rt/collector_rt.py
@@ -25,15 +25,15 @@ class RTCollectorBot(CollectorBot):
 
         self.set_request_parameters()
 
-        if getattr(self.parameters, 'search_newer_than', None):
+        if getattr(self.parameters, 'search_not_older_than', None):
             try:
-                self.newer_than = parser.parse(self.parameters.search_newer_than)
-                self.newer_than_type = 'absolute'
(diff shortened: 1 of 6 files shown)


aaronkaplan:
looks good.

intelmq/bots/collectors/rt/collector_rt.py
```diff
@@ -22,17 +25,35 @@ def init(self):
 
         self.set_request_parameters()
 
+        if getattr(self.parameters, 'search_newer_than', None):
+            try:
+                self.newer_than = parser.parse(self.parameters.search_newer_than)
+                self.newer_than_type = 'absolute'
+            except ValueError:
+                self.newer_than_relative = timedelta(minutes=FilterExpertBot.parse_relative(self.parameters.search_newer_than))
```
aaronkaplan: This is a dependency on another bot in the code.
Proposal: let's pull out the relevant code from the FilterExpertBot and move it to ``lib/``
This way , we only have dependencies to ``lib/``

aaronkaplan: IMHO we should have an error message for the admin/user if neither the search_newer_than field is relative nor absolute.

PR #932, opened 2017-03-27
Strip API keys for spamhaus cert feed

wagner-certat:
fixes certtools/intelmq#889


intelmq/bots/parsers/spamhaus/parser_cert.py
```diff
@@ -44,6 +44,7 @@ def process(self):
 
             row_splitted = [field.strip() for field in row.split(',')]
             event = self.new_event(report)
+            event.change("feed.url", event["feed.url"].split("key=")[0])
```
aaronkaplan: How do we notice if spamhaus decides to change that?

aaronkaplan: we can't

PR #935, opened 2017-03-28
Generic csv parser: use extra fields and regular expressions

wagner-certat:
Although we don't need that now in our setup, I'm sure it is useful for other purposes.

* You can now use `extra.some_field_name` as columns names
* and regular expressions to filter data inside fields

See also the newly added docs for examples


wagner-certat:
I added support for the unix timestamp format and windows NT time format.

dmth:
nice!

intelmq/bots/parsers/generic/parser_csv.py
```diff
@@ -52,25 +65,38 @@ def parse_line(self, row, report):
         event = self.new_event(report)
 
         for key, value in zip(self.columns, row):
+            extra = {}
```
wagner-certat: Thanks for finding this boner!

PR #937, opened 2017-03-29
Microsoft azure collector

wagner-certat:
Needed for MS DCU feed (feed description and configuration will follow)

### DIFF
--- a/intelmq/bots/collectors/microsoft/collector_azure.py
+++ b/intelmq/bots/collectors/microsoft/collector_azure.py
@@ -43,7 +43,6 @@ class MicrosoftAzureCollectorBot(CollectorBot):
                 report = self.new_report()
                 report.add('raw', unzipped)
                 self.send_message(report)
-                return  # TODO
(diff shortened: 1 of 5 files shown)


intelmq/bots/collectors/microsoft/collector_azure.py
```diff
@@ -0,0 +1,54 @@
...
+    azure = None  # noqa
+
+
+class MicrosoftAzureCollectorBot(CollectorBot):
+    def init(self):
+        if azure is None:
+            raise ValueError('Could not import azure.storage. Please install it.')
+
+    def process(self):
+        storage_client = azure.storage.CloudStorageAccount(self.parameters.account_name,
```
aaronkaplan: We should also support proxies settings.

See: def set_proxy(self, host, port, user=None, password=None):
https://azure-storage.readthedocs.io/en/latest/_modules/azure/storage/storageclient.html?highlight=proxy

PR #938, opened 2017-03-31
ENH: Extra Fields for Generic CSV

navtej:
This patch adds the capability to have extra fields as dict (similar to otx and blueliv parsers). 
If a column name starts with `extra.` , the property of extra will be added as dict to "extra" field in event.

### DIFF
--- a/intelmq/bots/parsers/generic/parser_csv.py
+++ b/intelmq/bots/parsers/generic/parser_csv.py
@@ -50,6 +50,7 @@ class GenericCsvParserBot(ParserBot):
 
     def parse_line(self, row, report):
         event = self.new_event(report)
+        additional = {}
 
         for key, value in zip(self.columns, row):
 
@@ -65,11 +66,17 @@ class GenericCsvParserBot(ParserBot):
                     value = self.type_translation[value]
                 elif not hasattr(self.parameters, 'type'):
                     continue
+            elif key.startswith("extra."):
+                _, key_ = key.split('.', 1)
+                additional[key_] = value
+                continue
             event.add(key, value)
 
         if hasattr(self.parameters, 'type')\
                 and "classification.type" not in event:
             event.add('classification.type', self.parameters.type)
+        if len(additional) > 0:
+            event.add('extra', additional)
         event.add("raw", self.recover_line(row))
         yield event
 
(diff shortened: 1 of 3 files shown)


wagner-certat:
This is also part of #935

intelmq/bots/parsers/generic/parser_csv.py
```diff
@@ -65,11 +66,17 @@ def parse_line(self, row, report):
                     value = self.type_translation[value]
                 elif not hasattr(self.parameters, 'type'):
                     continue
+            elif key.startswith("extra."):
+                _, key_ = key.split('.', 1)
```
wagner-certat: If `key` is `extra.foo.bar` then `key_` is `foo`

intelmq/bots/parsers/generic/parser_csv.py
```diff
@@ -65,11 +66,17 @@ def parse_line(self, row, report):
...
+                additional[key_] = value
+                continue
             event.add(key, value)
 
         if hasattr(self.parameters, 'type')\
                 and "classification.type" not in event:
             event.add('classification.type', self.parameters.type)
+        if len(additional) > 0:
```
wagner-certat: `if additional` is sufficient.

intelmq/bots/parsers/generic/parser_csv.py
navtej: No, then key is `foo.bar`

```
>>> "extra.foo.bar".split(".",1)
['extra', 'foo.bar']

```

intelmq/bots/parsers/generic/parser_csv.py
navtej: Not really
```
>>> additional = {}
>>> if additional: print("True")
...
>>>
```

navtej:
I started a review for #935 . With one minor change it does offer the same functionality.

intelmq/bots/parsers/generic/parser_csv.py
wagner-certat: Ah, right. Sorry.

intelmq/bots/parsers/generic/parser_csv.py
navtej: ideed, TIL.

wagner-certat: Which proves that it is sufficient, isn' it?

`bool({}) == False`

wagner-certat:
> With one minor change it does offer the same functionality.

So is there any functionality in this PR which is not in #935?

navtej:
Yes more like a bug, details are already on the opened review @ #935 . However I noticed i started a review but didnt submit it so it was just sitting there probably and you never saw it.

PR #939, opened 2017-04-03
Added authentication to mongoDB output

vince-nayal:
Hello,

With this PR, i'd like to add the authentication option to mongoDB ouput bot.

It should not have any impact on config not needing auth.

Regards,

### DIFF
--- a/docs/Bots.md
+++ b/docs/Bots.md
@@ -725,6 +725,8 @@ FIXME
 
 * `collection`: MongoDB collection
 * `database`: MongoDB database
+* `db_user` : Database user that should be used if you enabled auth
+* `db_pass` : Password associated to `db_user`
 * `host`: MongoDB host (FQDN or IP)
 * `port`: MongoDB port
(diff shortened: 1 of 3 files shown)


wagner-certat:
I read the documentation concerning authentication in mongodb. There's always a separate database for users, called `admin`. To authenticate one must first use this DB, authenticate there, and then switch to the actual database.


Did I undertand this correctly?

vince-nayal:
From my understanding and my recent use of auth in mongoDB, i don't think so.
You have both **Authentication** and **Authorization**

- **Authentication** is per db and simply forbids anonymous access to the db
- **Authorization** grants roles to the db users (R, RW, userAdmin). it can be per db or mongo wide
The admin db is where you define roles for all db (eg: dbAdminAnyDatabase)

ref : [mongo doc (2.4)](https://docs.mongodb.com/v2.4/core/access-control/)

wagner-certat:
Thanks, pushed (slightly rebased) to master, added default values and even [tests](f4c8297efe2a503076194cfab7c80ebf48993ef4).

vince-nayal:
Thanks !

PR #944, opened 2017-04-09
Netlab 360 enh

navtej:
Two enhancements for netlab-360 parser
1. Add Mira Scanner Feed Parser logic
2. Change to DGA feed parser,  the feed columns are `DGA family, Domian, Start and end of valid time(UTC)`, original parser picked `End Time` as time.source. However End Time is unlike last updated time, and some values can be too far in future. Some of the items in the feed has end times for 2030 year. It is more sensible to use Start time as `time.source`


intelmq/bots/parsers/netlab_360/parser.py
```diff
@@ -21,18 +22,28 @@ def parse_line(self, line, report):
 
             if report['feed.url'] in Netlab360ParserBot.DGA_FEED:
                 event.add('source.fqdn', value[1])
-                event.add('time.source', value[3] + ' UTC')
+                # DGA Feed format is
+                # DGA family, Domian, Start and end of valid time(UTC)
+
+                event.add('time.source', value[2] + ' UTC')
```
wagner-certat: The last time this domain has been seen is in the fourth column, right?

I think this should be the latest time we know this event is true. @aaronkaplan?

navtej:
added the documenation

intelmq/bots/parsers/netlab_360/parser.py
aaronkaplan: yes. If we can tell ISPs that we according to the feed this is a valid point in time where their logs (proxy logs, netflow ,...) would find this IP address behaving in this particular way.

docs/Feeds.md
```diff
@@ -1633,6 +1633,41 @@ id: netlab360-magnitude-ek-parser
...
+**Bot Name:** Generic URL Fetcher
+
+**Bot Module:** intelmq.bots.collectors.http.collector_http
+
+**Configuration Parameters:**
+```
+id: netlab360-mirai-scanner-collector
+provider: Netlab 360
+feed: Netlab 360 Mirai Scanner List
+rate_limit: FIXME
```
aaronkaplan: We should not have a FIXME here.

intelmq/bots/parsers/netlab_360/parser.py
```diff
@@ -21,18 +22,28 @@ def parse_line(self, line, report):
...
                 event.add('classification.type', 'c&c')
                 event.add('event_description.url', 'http://data.netlab.360.com/dga')
 
             elif report['feed.url'] in Netlab360ParserBot.MAGNITUDE_FEED:
                 event.add('time.source', DateTime.from_timestamp(int(value[1])))
                 event.add('source.ip', value[2])
-                event.add('source.fqdn', value[3])
+                # ignore ips as fqdns
```
aaronkaplan: What happens when a non-fqdn and non IP address strings comes in here? 
Like for example the string "|||" . That's neither an IP nor a domain.
IMHO a parser should carefully ensure that the input is correct. If not, it must not pass on invalid input.

intelmq/bots/parsers/netlab_360/parser.py
wagner-certat: So please leave it as `value[3]`, as this is the most recent time we can get and this is most useful for the processing afterwards.

docs/Feeds.md
navtej: There are multiple places in documentation having FIXME. https://github.com/certtools/intelmq/blob/master/docs/Feeds.md . I assumed where the feed did not provided update frequency, docs put FIXME for the user to choose an interval.

wagner-certat: I think a daily update is ok. Their last update was in February tough: http://data.netlab.360.com/mirai-scanner/

intelmq/bots/parsers/netlab_360/parser.py
navtej: I dont see how year 2030 is a valid date for any practical use? So can you please rationalize that choice?

wagner-certat:
A time in the future is indeed ridiculous.
97979 of 1034451 domains have 203* as end date.

Probably we should ask the mailinglist how to deal with time frames. I
think this needs some changes...

docs/Feeds.md
navtej: Changed.

wagner-certat:
I merged all commits except ee4fd25 to master. concerning the time frames I opened a thread at the mailing list: https://lists.cert.at/pipermail/intelmq-dev/2017-April/000191.html

navtej:
We havent heard anything from the list do we want to take a call on how to handle this?

wagner-certat:
Actually I have no idea - except my proposal to use the time.observation (as proposed on the ML). I'll push the thread on the list.

navtej:
Still nothing on the mailing list. Can we solve this by couple of `time.XXX` harmonization fields. I was thinking perhaps, `time.last_updated` and `time.valid_till`. Some feeds do provide the updated timestamps along with timestamp when the initial data came in.  We don't need to make assumptions as of which field to use, if we have `time.last_updated`. End users can make their own decision regarding this. Similarly `time.valid_till` can possibly solve the time range issue, without making assumptions on end use cases.

wagner-certat:
I don't want to make such decisions on my own. So I fear this is open until others comment on it.

SYNchroACK:
Hi @navtej . Can you adapt the code as @wagner-certat suggested (and I also agreed) on mailing-list in order to proceed with the merge. Thank you for your contribution and sorry for the long delay.

wagner-certat:
> Hi @navtej . Can you adapt the code as @wagner-certat suggested (and I also agreed) on mailing-list in order to proceed with the merge. Thank you for your contribution and sorry for the long delay.

We could to that in the core too.
And probably optional if some users (@navtej?) want to deal with future dates?

navtej:
@wagner-certat I am ok with changes in core as well, that can take care of any date anamoly. Otherwise I will make required changes, in couple of days.

navtej:
have anyone taken a call if the changes should be to core or this bot?

PR #951, opened 2017-04-19
Typosquatting

wagner-certat:
used aspell for all files except the tests


wagner-certat:
Pushed a fix after the last merge from master broke the 'docs/Harmonization-fields.md' file

Issue #959, opened 2017-04-25
codestyle: avoid 'bare excepts'


bernhardreiter:
Python's PEP8 style guide recommends to avoid 'bare excepts', see the quote in https://github.com/PyCQA/pycodestyle/issues/579 .

Pycodestyle's E722 warns against bare excepts.
Here is a run with 9ae778cff9363e85b24ae4e1928fd4fd222c220e
and pycodestyle v2.3.1:
```
pycodestyle --select=E722
./intelmq/bots/parsers/shadowserver/parser.py:137:21: E722 do not use bare except'
./intelmq/bots/outputs/postgresql/output.py:46:9: E722 do not use bare except'
./intelmq/bots/outputs/redis/output.py:26:9: E722 do not use bare except'
./intelmq/bots/outputs/udp/output.py:48:9: E722 do not use bare except'
./intelmq/bots/outputs/tcp/output.py:35:9: E722 do not use bare except'
./intelmq/bots/experts/cymru_whois/lib.py:131:13: E722 do not use bare except'
./intelmq/bots/experts/ripencc_abuse_contact/expert.py:49:9: E722 do not use bare except'
./intelmq/bots/experts/generic_db_lookup/expert.py:40:9: E722 do not use bare except'
./intelmq/bots/experts/abusix/lib.py:30:9: E722 do not use bare except'
./intelmq/lib/bot.py:65:9: E722 do not use bare except'
./intelmq/lib/bot.py:259:9: E722 do not use bare except'
./intelmq/lib/bot.py:393:9: E722 do not use bare except'
```

The code should be more clear and indicate if the bare except is on purpose or not
by using the explicit Exception or BaseException classes as indicated in pep8 and some comments
why this is the wanted behaviour I guess. :)

wagner-certat:
Already on my agenda too :)

bernhardreiter:
@wagner-certat good! Then we can remove the ignore-E722-setting for pycodestyle.  
(I had pondered if I should directly proposes changes while preparing https://github.com/certtools/intelmq/pull/960 but felt that I do not understand the structure well enough to come up with the right meaning.)

PR #960, opened 2017-04-25
Codestyle: move to pycodestyle style checker

bernhardreiter:
http://pep8.readthedocs.io/en/release-1.7.x/ states:

> This package used to be called pep8 but was renamed to pycodestyle to reduce confusion. 

And [pycodestyle](https://pypi.python.org/pypi/pycodestyle) is further maintained and improved since the last pep8 release. I propose to move to pycodestyle with this pull-request.

Because of https://github.com/certtools/intelmq/issues/959 I have added E722 to the warning that are ignored until the issue is resolved.


setup.cfg
```diff
@@ -12,8 +12,8 @@ all_files  = 1
 [upload_sphinx]
 upload-dir = docs/build/html
 
-[pep8]
-ignore = E221
+[pycodestyle]
```
wagner-certat: At least pycodestyle 2.0.0 still reads from the [pep8] section, for compatibility we could just duplicate this section.

bernhardreiter: I'd rather go fully pycodestyle. Thanks for the review!

wagner-certat: pycodestyle 2.0 (2016-05-31) is also pycodestyle, but it does not recognize the `[pycodestyle]` section in `setup.cfg`. This has been changed in [2.1.0 (2016-11-04)](https://pypi.python.org/pypi/pycodestyle#id4). So for distros not newer than some months, the settings would not apply. I duplicated the section to keep support for these versions. See 59246fbd5f1636797697b1676e08eceb3475ac16

bernhardreiter: For development I believe it is fine to demand an uptodate version, but its your call. Thanks for merging the commit.

PR #971, opened 2017-05-11
Add note to explain absent /opt/intelmq dir with pypi install

chorsley:


wagner-certat:
Hi, thanks for addressing this. However, this does not solve the problem the users are facing. So please add short instructions how to get this done (ie. copy the files from site-packages)

Issue #743, opened 2016-10-17
Logging: Reduce amount of log-messages


dmth:
Whilst running our Test-System with approx 19 collectors + parsers and ~5 experts, we've noticed that the log-file directory will get very huge in a rather short amount of time.  Since beginning of July we've collected approx. 11 GB Logfiles. Logrotation and zipping of logfiles is active.

The Loglevel in our setup is quite verbose (info), and should be set to warning. We suspect that this will create much less log-messages. In addition it will filter out the "Processed i+500 messages" messages. (IMHO This message should also be a debug one)

Nevertheless, we've noticed that the logfiles can be flooded with messages like "empty message received" For instance when redis has not been started. This message is added as a warning to all bots in this case an will be added to the log every X milliseconds.

Possible ToDos:
- [x] The implications of a verbose log-level could be added to the user-documentation.
- [ ] Decide if the "Processed i+500 messages" should be info or debug. There might also a chance for optimization in ll. 269 of lib/bot.py for instance by checking the configured log-level prior to the modulo operation. (For each message the modulo is calculated even if the code after the operation does not do anything)
- [x] Find other cases where logging is too verbose / or not sufficient
- [x] We should agree on a "sane" default for logging. This might be warning in case of a productive environment.

sebix:
Also see #614

> In addition it will filter out the "Processed i+500 messages" messages. (IMHO This message should also be a debug one)

Or make it configurable? But it's obsolete as soon as we have #361

sebix:
> The implications of a verbose log-level could be added to the user-documentation.

9c0c55aa8044e54efafe7b83cf58e67dc654e607

>  Find other cases where logging is too verbose / or not sufficient

#641
>  We should agree on a "sane" default for logging. This might be warning in case of a productive environment.

Should warning be default for packaged releases? I think INFO is still a good default, but for production environments this should be warning tough.

dmth:
> >  We should agree on a "sane" default for logging. This might be warning in case of a productive environment.
>
> Should warning be default for packaged releases? I think INFO is still a good default, but for production environments this should be warning tough.

I agree on "Warning" as a reasonable default for productive settings.

wagner-certat:
> Decide if the "Processed i+500 messages" should be info or debug. There might also a chance for optimization in ll. 269 of lib/bot.py for instance by checking the configured log-level prior to the modulo operation. (For each message the modulo is calculated even if the code after the operation does not do anything)

[...]

PR #974, opened 2017-05-15
Log count of processed messages configurable

wagner-certat:
It's now configurable how often the "Processed .. messages" is logged. You can give a count and seconds.

See #743


wagner-certat:
> Have you implemented your idea from #743 (comment), also?

Thanks for the reminder. Did that including tests and also merged master.

PR #979, opened 2017-05-22
error in the bot shouldnt affect whole botnet

e3rd:
We had an error – our proprietary bot wasn't listed in BOTS file. Thus, no launcher in bin was created. When launching whole botnet only bots that preceded this bot in the alphabet were launched, the rest stayed stopped.  
With this change, when an error occurs, the rest of the botnet will stay unaffected → only erroneous bots won't get launched → it will be easier to detect the bug for the administrator.


e3rd:
Would you add the error message like this?

PR #980, opened 2017-05-22
[ENH]AbuseCH Ransomware Parser Fixes

navtej:
Sometimes in the feed we see an ip address with a leading ','
this enhancement fixes that case

### DIFF
--- a/intelmq/bots/parsers/abusech/parser_ransomware.py
+++ b/intelmq/bots/parsers/abusech/parser_ransomware.py
@@ -40,10 +40,12 @@ class AbuseCHRansomwaretrackerParserBot(Bot):
                         ev.add('classification.type', 'c&c')
                         ev.add('time.source', nrow[0] + ' UTC', overwrite=True)
                         ev.add('status', nrow[5])
-                        # in the feed sometimes a , is seen before srcip
+                        # in the feed sometimes a ',' is seen before srcip
                         # perhaps it is an error on feeds part
                         # however we have to deal with it
-                        src_ip = nrow[7][1:] if nrow[7][0] == ',' else nrow[7]
+                        src_ip = nrow[7]
+                        if src_ip[0] == ',' and len(src_ip) > 8:
+                            src_ip = src_ip[1:]
                         ev.add('source.ip', src_ip)
                         #ev.add('source.ip', nrow[7])
                         ev.add('raw', ','.join(nrow))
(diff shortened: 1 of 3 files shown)


intelmq/bots/parsers/abusech/parser_ransomware.py
```diff
@@ -40,7 +40,15 @@ def process(self):
                         ev.add('classification.type', 'c&c')
                         ev.add('time.source', nrow[0] + ' UTC', overwrite=True)
                         ev.add('status', nrow[5])
-                        ev.add('source.ip', nrow[7])
+                        # in the feed sometimes a ',' is seen before srcip
+                        # perhaps it is an error on feeds part
+                        # however we have to deal with it
+                        src_ip = nrow[7]
+                        if src_ip and len(src_ip) > 8 and src_ip[0] == ',':
```
wagner-certat: Why so complicated? With a simple `src_ip.lstrip(',')` (or event the more generic strip) you don't need these 3 checks.

navtej: This takes care of an additional case where the feed can contain values like `,104.` as IP address. Generally these are fix in the next fetch but in case it comes, the current processing will not stop. This is not the best way to take care of it though.

wagner-certat: So if the ip address is too short, you want to ignore them? That's not what this code does. IPv6 adresses can be shorter.

I reported these issues (comma, cut ip addresses) upstream at abuse.ch and those have been fixed already!

So the only thing we need to take care of now are the 0.0.0.0 IPs.

navtej: sliced ip addresses are generally fixed for the next fetch.

wagner-certat: They are fixed upstream, so we don't need to workaround anymore

intelmq/bots/parsers/abusech/parser_ransomware.py
```diff
@@ -43,11 +43,9 @@ def process(self):
                         # in the feed sometimes a ',' is seen before srcip
                         # perhaps it is an error on feeds part
                         # however we have to deal with it
-                        src_ip = nrow[7]
-                        if src_ip and len(src_ip) > 8 and src_ip[0] == ',':
-                            src_ip = src_ip[1:]
+                        src_ip = nrow[7].strip(',')
                         if src_ip != '0.0.0.0':
-                            ev.add('source.ip', src_ip)
+                            ev.add('source.ip', src_ip, raise_failure=False)
```
wagner-certat: Why do you want to use it then?

wagner-certat:
I cherry-picked commit cba9ee8cf0132ad01db14ba1de2d821bd80526fc for master: 7bbbe7193cf4037e3de34f76bdc0eef875f4301d to fix the 0.0.0.0 issue

PR #984, opened 2017-05-30
[ENH] extract malware name from fraunhofer dga feed

navtej:
Adds the malware name to the dga feed domain.


wagner-certat:
I asked upstream if this is a good solution and it's guaranteed to work.

wagner-certat:
> I asked upstream if this is a good solution and it's guaranteed to work.

According to upstream it is guaranteed that everything before the first underscore can be treated as malware name.

navtej:
I have sqashed the commits but i cant push something is wrong with upstream, could you please do it? Thanks.

wagner-certat:
Done in 0eb770cc1ca7f146974afd45a7fa8653c15171bd

Added an additional note in 48760afcca5e486f22a7304d79898b1401d6ec2e

PR #993, opened 2017-06-05
Alienvault OTX: cycling due to an invalid message failure

e3rd:
The invalid lines should be ignored.  
Now, the bot is dying due to ```intelmq.lib.exceptions.InvalidValue: invalid value 'http:///ls5/forum.php' (<class 'str'>) for key 'source.url'```  
In the moment of exception, the rest of lines won't get parsed and the message is acknowledged (so that the error repeats in few minutes again and again).


intelmq/bots/parsers/alienvault/parser_otx.py
```diff
@@ -33,50 +35,54 @@ def process(self):
...
-                else:
+
+                    # CVE
+                    elif indicator["type"] in ['CVE']:
+                        additional_indicator['CVE'] = indicator["indicator"]
+                        # TODO: Process these IoCs: FilePath, Mutex
+                    else:
+                        continue
+                except InvalidValue:
+                    self.logger.exception('Failed to parse line: {}\n{}'.format(indicator, format_exc()))
```
wagner-certat: When using `logger.exception` the traceback is appended automatically: https://docs.python.org/3/library/logging.html#logging.Logger.exception

But AFAIS the message is not dumped at all then, it's only logged. Which means you need to get the message from the logs, add it to the dump file and then recover it. That's neither intuitive nor in line with all the other bots.

e3rd: Thanks, now I understand the dump functionality pretty better!

I extended the functionality of parse_line. In the docs, there is written `parse_line: Parses elements, returns an Event.` But till now, it handled list of Events only. Now, it can get list, the single event or None (no event in the row).
* Please comment if you find this extension useful!
* I've added methods for parsing JSON. What about to create JsonParserBot, CsvParserBot and CsvDictParserBot as children of ParserBot. Now, we have to redefine `parse` and `recover_line` if we want to use csv or csvDict. Maybe it would be easier if adopting CsvDictParserBot automatically used parse_csv_dict instead of parse_csv and recover_line_csv_dict instead of recover_line.

e3rd:
It is strange to me that Travis failed because I didnt change the file intelmqctl.py .

```bash
intelmq/bin/intelmqctl.py:136:12: E111 indentation is not a multiple of four
````

It seems to me there is no error in intelmqctl.py in this PR, it IS multiple of four:
```python
        bot = getattr(module, 'BOT')
        try:
            instance = bot(bot_id)
            instance.start()```

**Comment by wagner-certat** [comment]
Strange, so just ignore it for now.

**Comment by wagner-certat** [comment]
Fine if the tests would be working :D 
Could you squash these commits please? At lease the two "travis" ones.

**Comment by e3rd** [comment]
Squatched. Wow, rebasing is cool :o

**Comment by e3rd** [comment]
I tried to squash it, at localhost it seemed alright, but here at GH a merge of master branch was made in a new commit. However, it seems alright...

**Comment by e3rd** [comment]
There is an error now in Travis but I'm not able to see the same on localhost. I have to investigate this when I get back in 14 days :(

**Comment by dmth** [comment]

**Comment by wagner-certat** [comment]
@dmth at least currently the test fail is not due to the certat-contact expert, it's a bug in this alienvault bot.

**Comment by dmth** [comment]
> @dmth at least currently the test fail is not due to the certat-contact expert, it's a bug in this alienvault bot.

Yes, that's true. After yesterdays changes in the certat-contact-database, the test `TestCERTatContactExpertBot` works again, but `tests/bots/parsers/alienvault/test_parser_otx.py", line 116,` fails instead:

```
  File "/home/travis/build/certtools/intelmq/intelmq/tests/bots/parsers/alienvault/test_parser_otx.py", line 116, in test_event
    self.assertMessageEqual(11, EXAMPLE_EVENT_2)
  File "/home/travis/build/certtools/intelmq/intelmq/lib/test.py", line 429, in assertMessageEqual
    event = self.get_output_queue()[queue_pos]
nose.proxy.IndexError: list index out of range
```

e3rd:
I fixed the test and merged carefully the changes from #1009 and #1016 .

It seems that the errors in Travis are bound rather to `ripencc_abuse_contact` bot. Could you check it please?

wagner-certat:
The fixed ripencc test is in develop now.

I fixed the (again) wrong indent in 4270930e050e71bd2cab6b99e5487916cebecd08

Thanks!

PR #997, opened 2017-06-07
Shadowserver config - feed names adjusted

CERT-Bund:
This PR changes the feed names in the config.py for the Shadowserver parsers to match the currently used feed names. Additionally, missing taxonomy definitions have been added and the spelling was corrected (lower-/upper case).


dmth:
Thank you for reviewing and editing the configuration.

I've some comments which might need to be checked / discussed:
1. The keys of the feed-names have been chosen in a way to be identical to the wiki of shadowserver and not the subjects in the emails. By changing these keys, the PR is breaking compatibility of already existing installations. (DNS-open-resolvers instead of DNS-Open-Resolvers). That's how Shadowserver chose to document their feed-names. Keeping the Keys that way simplifies configuration for a new user, as we can simply refer to the shadowserver documentation. The shadowserver wiki tends to contain a lot of deprecated information. But I still think it is a good baseline for development and naming-conventions.
If the key shall be changed to the E-Mail subjects the Readme of the parser (https://github.com/certtools/intelmq/blob/master/intelmq/bots/parsers/shadowserver/README.md) needs to be updated.
**Question:** Are the subjects the same for all receivers?
2. Renaming the identifiers to be on par with the feed itself is ok, it might break other instances of IntelMQ, thus needs to be communicated to the community.

The other changes are fine.

CERT-Bund:
The names of the reports in the subject lines should be the same for all recipients.

Here is a list of currently used report names:
Accessible CWMP Report
Accessible/Open MongoDB Service Report
Accessible RDP Services Report
Accessible SMB Service Report
Accessible Telnet Service Report
Accessible VNC Service Report
Accessible XDMCP Service Report
Blacklisted IP Addresses Report
Chargen Report
Command and Control Report
Compromised Website Report
DNS Open Resolvers Report
Drone Report
IPMI Report
IPv6 Sinkhole HTTP Drone Report
ISAKMP Vulnerability Scan Report
Microsoft Sinkhole Report
Netbios Report
NTP Monitor Report
NTP Version Report
Open DB2 Discovery Service Report
Open Elasticsearch Server Report
Open LDAP Services Report
Open mDNS Servers Report
Open Memcached Server Report
Open Portmapper Scan Report
Open Redis Server Report
Open TFTP Servers Report
QOTD Report
Sandbox URL Report
Sinkhole HTTP Drone Report
SNMP Report
Spam URL Report
SSDP Report
SSL/Freak Vulnerable Servers Report
SSLv3/Poodle Vulnerable Servers Report
Vulnerable NAT-PMP Systems Report

Some names (like SSDP) are missing the 'Open' prefix which we have added.

dmth:
> Some names (like SSDP) are missing the 'Open' prefix which we have added.

So the assumption "keys are like the e-mail subjects" will not fit either.

wagner-certat:
This breaks existing configurations, so some kind of backwards-compatibility would be appropriate.
This modifies a lot of identifiers, NEWS.md should contain queries for databases to apply these changes to existing data.

The purpose of the identifier is to be identical for the same kind of event.

Have `Microsoft Sinkhole` and `Sinkhole HTTP Drone` the same data? Both have the identifier `infected system`.

CERT-Bund:
Yes, 'Microsoft Sinkhole' and 'Sinkhole HTTP Drone' both include data on infected systems. The first one only includes data from Microsoft's sinkholes shared with Shadowserver for remediation. Currently, there is only data on B68/ZeroAccess.

wagner-certat:
> Yes, 'Microsoft Sinkhole' and 'Sinkhole HTTP Drone' both include data on infected systems. The first one only includes data from Microsoft's sinkholes shared with Shadowserver for remediation. Currently, there is only data on B68/ZeroAccess.

Then the identifier would be the malware familiy name. I'll work on that in the next weeks.

dmth:
> This breaks existing configurations, so some kind of backwards-compatibility would be appropriate.

One could use the feed-index https://github.com/CERT-Bund/intelmq/blob/e30fcde471fa767832f6f7b5076a1e5947d93b9a/intelmq/bots/parsers/shadowserver/config.py#L49 to provide the configurations under the old feednames, too. Nevertheless, the changed identifiers will still break existing configurations.

dmth:
Currently one test fails, because the feed-name was altered, this patches the test:

```patch
git diff /opt/src/intelmq/intelmq/tests/bots/parsers/shadowserver/test_drone_hadoop.py
diff --git a/intelmq/tests/bots/parsers/shadowserver/test_drone_hadoop.py b/intelmq/tests/bots/parsers/shadowserver/test_drone_hadoop.py
index a9c6997..b7875fd 100644
--- a/intelmq/tests/bots/parsers/shadowserver/test_drone_hadoop.py
+++ b/intelmq/tests/bots/parsers/shadowserver/test_drone_hadoop.py
@@ -285,7 +285,7 @@ def test_event(self):
             self.assertMessageEqual(i, EVENT)
 
 
-TESTING_OVERWRITE_FEEDNAME = 'My-Botnet-Drone-Hadoop'
+TESTING_OVERWRITE_FEEDNAME = 'Botnet-Drone-Hadoop'
 
 
 class TestOverwriteShadowserverParserBot(test.BotTestCase, unittest.TestCase):
@@ -295,7 +295,7 @@ def set_bot(cls):
         cls.bot_reference = ShadowserverParserBot
         cls.default_input_message = EXAMPLE_REPORT.copy()
         cls.default_input_message['feed.name'] = TESTING_OVERWRITE_FEEDNAME
-        cls.sysconfig = {'feedname': 'Botnet-Drone-Hadoop',
+        cls.sysconfig = {'feedname': 'Drone',
                          'overwrite': True}
 
     def test_bot_name(self):
@@ -307,7 +307,7 @@ def test_overwrite(self):
         self.run_bot()
         for i, EVENT in enumerate(EVENTS):
             event = EVENT.copy()
-            event['feed.name'] = 'Botnet-Drone-Hadoop'
+            event['feed.name'] = 'Drone'
             self.assertMessageEqual(i, event)
```

intelmq/bots/parsers/shadowserver/config.py
```diff
@@ -422,9 +422,9 @@ def validate_fqdn(value):
     ],
     'constant_fields': {
         'classification.type': 'vulnerable service',
-        'classification.taxonomy': 'Vulnerable',
-        'classification.identifier': 'openportmapper',
-        'protocol.application': 'portmapper',
+        'classification.taxonomy': 'vulnerable',
+        'classification.identifier': 'open-portmapper',
+        'protocol.application': 'portmap',
```
wagner-certat: Why is the application `portmap` and the identifier `open-portmapper`?

wagner-certat:
I addresses the issues mentioned here in my PR cert-bund/intelmq#1 to @CERT-Bund:

* documented the new names in the readme, changelog and news
* added backwards compatibility for the old feed names and added it to the readme
* added upgrade statements for postgres to the news file
* merged with master

For https://github.com/certtools/intelmq/pull/997#issuecomment-308138577 see #1034

intelmq/bots/parsers/shadowserver/config.py
dmth: The application is called portmap: https://en.wikipedia.org/wiki/Portmap

wagner-certat:
reminder @aaronkaplan

wagner-certat:
reminder @aaronkaplan

wagner-certat:
I fixed the merge conflicts

wagner-certat:
Moved the changelog and news sections to 1.1

wagner-certat:

PR #999, opened 2017-06-09
[ENH] Generic csv enh

navtej:
Apologies for a consolidated change set
Here are all the changes 

1.  Multiple values for columns.
    Frequently we come across feeds which can be easily handled with generic csv parser, however the parser often fails because of mixed content in some column eg FQDN/IP, IP/CIDR. This change allows one column to specify multiple possible values using '|' char eg `source.ip|source.network`
2.  Fix for weird timestamp values of length 12 and 13
    I came across some prop device feeds where the timestamp had addition two or three 000.  An attempted fix for such cases is implemented.
3.  data type for columns
    A possibility for data type conversion, similar to timestamp conversion. Currently only json data type is implemented.
eg `"data_type": "{\"extra.tags\":\"json\"}"` alongwith `{"columns": [ "source.ip", "source.url", "extra.tags"]`
4.  fix for empty url
         if the source omitted the value for .url column, existing code added http:// to it and proceeded to add the value which would fail. A check is added to handle this case.
5.  fix for regex test case
 although the test is focused on asn regex match but the testcase example event check never included the asn value. Fixes the omission. 
6.  .strip() values before adding to accommodate  white space between csv values
     feeds sometimes contain a space char between values separated with some delimiter. The IP values with leading/trailing space fail the harmonization checks, unless stripped.

**Updated:  23rd Aug**
Add ability for csv parser to skip/process line based on if the line contains or not contains some value.


intelmq/bots/parsers/generic/parser_csv.py
```diff
@@ -66,30 +71,61 @@ def parse_line(self, row, report):
...
-                    extra[key[6:]] = value
-            else:
-                event.add(key, value)
+
+                if key in self.data_type:
+                    value = DATA_CONVERSIONS[self.data_type[key]](value)
+
+                if key in ["time.source", "time.destination"]:
+                    if self.time_format == 'timestamp':
+                        if len(value) == 12:
```
wagner-certat: I'd prefer a new time conversion 'epoch' for this kind of format. And tests.

intelmq/bots/parsers/generic/parser_csv.py
```diff
@@ -66,30 +71,61 @@ def parse_line(self, row, report):
...
+                regex = self.column_regex_search.get(key, None)
+                if regex:
+                    search = re.search(regex, value)
+                    if search:
+                        value = search.group(0)
+                    else:
+                        value = None
+
+                if key in ["__IGNORE__", ""]:
+                    stop_processing = True
```
wagner-certat: See navtej#1

navtej:
Sorry for the delay on this, i will work on it this week.

wagner-certat:
@navtej Any updates in sight? I'd like to include these additions!

docs/Bots.md
```diff
@@ -364,14 +364,30 @@ Lines starting with `'#'` will be ignored. Headers won't be interpreted.
...
+    Some use cases - 
+        - mixed data set, e.g. URL/FQDN/IP/NETMASK  `"columns": "source.url|source.fqdn|source.ip|source.network"`
+        - parse a value and ignore if it fails  `"columns": "source.url|__IGNORE__"`
  * `"column_regex_search"`: Optional. A dictionary mapping field names (as given per the columns parameter) to regular expression. The field is evaulated using `re.search`. Eg. to get the ASN out of `AS1234` use: `{"source.asn": "[0-9]*"}`.
  * `"default_url_protocol"`: For URLs you can give a defaut protocol which will be pretended to the data.
  * `"delimiter"`: separation character of the CSV, e.g. `","`
  * `"skip_header"`: Boolean, skip the first line of the file, optional. Lines starting with `#` will be skipped additionally, make sure you do not skip more lines than needed!
- * `time_format`: Optional. If `"timestamp"` or `"windows_nt"` the time will be converted first. With the default `null` fuzzy time parsing will be used.
+ * `time_format`: Optional. If `"timestamp"`, `"windows_nt"` or `"epoc"` the time will be converted first. With the default `null` fuzzy time parsing will be used.
```
wagner-certat: Did you mean `epoch` instead of `epoc`?

navtej: yes. I will fix it. I need fix the formatting as well.

intelmq/lib/harmonization.py
```diff
@@ -230,6 +230,21 @@ def __parse_utc_isoformat(value):
         return value
 
     @staticmethod
+    def from_epoch_millis(tstamp, tzone='UTC'):
+        """
+        Returns ISO formatted datetime from given epoch timestamp with milliseconds.
+        It ignores the milliseconds, converts it into normal timestamp and processes it.
+        """
+        bytecount = len(str(tstamp))
+        int_tstamp = tstamp if isinstance(tstamp, int) else int(tstamp)
```
wagner-certat: What is that for? If you call `int()` on a int object, the same object is returned.

wagner-certat:
merge with develop and some fixes including a rename of `text` to `filter` are in navtej/intelmq#3

wagner-certat:
Merged with my additions/changes

PR #1000, opened 2017-06-11
Update for new CleanMX parser behaviour

chorsley:
The cleanmx parser appears to be a single class now, detecting type based on URL args as of 87d5bea69.


wagner-certat:
Thanks!

PR #1002, opened 2017-06-11
Offer complete path to bash completion script

JesseBowling:
Modify the path to give exact path from base directory of git project.


wagner-certat:
Thanks!

Issue #996, opened 2017-06-06
 Cymru-whois bot returning registry 'OTHER' for 6to4 IPv6 address


haam3r:
So I'm seeing things like this:
```
intelmq.lib.exceptions.InvalidValue: invalid value 'OTHER' (<class 'str'>) for key 'source.registry' is_valid returned False.
2017-06-06 03:43:20,871 - cymru-whois-expert - INFO - Current Message(event): {'source.network': '2002::/16', 'classification.type': 'b
lacklist', 'source.asn': 6939, 'feed.url': 'https://lists.blocklist.de/lists/all.txt', 'feed.accuracy': 100.0, 'time.observation': '201
7-06-05T18:52:38+00:00', 'source.ip': '2002:d06e:4282::d06e:4282', 'feed.provider': 'blocklist.de', 'feed.name': 'blocklist', 'raw': 'M
jAwMjpkMDZlOjQyODI6MDAwMDowMDAwOjAwMDA6ZDA2ZTo0Mjgy', 'source.as_name': 'HURRICANE - Hurricane Electric, Inc., US', 'classification.tax
onomy': 'other'}.
```
What should be the correct way to handle this?

wagner-certat:
IMHO we should (explicitly) ignore that

haam3r:
I'm currently using a very dirty workaround of adding 'OTHER' as a valid registry in lib.harmonization at:
https://github.com/certtools/intelmq/blob/master/intelmq/lib/harmonization.py#L689

wagner-certat:
Upstream said that this is intention: As it's an 6to4 address, there's no registry. The same will happen for RFC1918 and private addresses too.

For 6to4 addresses we can do the registry lookup based on the extracted IPv4 address.

wagner-certat:
Also note that the network owners for IPv6 and IPv4 are different:
```
> dig +short 2.8.2.4.e.6.0.d.0.0.0.0.2.8.2.4.e.6.0.d.2.0.0.2.origin6.asn.cymru.com. TXT
"6939 | 2002::/16 |  | other |"
> dig +short 130.66.110.208.origin.asn.cymru.com. TXT
"32097 | 208.110.64.0/19 | US | arin | 2006-03-17"
```
We could duplicate such events to have both the v4 and the v6 data.

navtej:
I ran into this today. Pushing a fix.

PR #1009, opened 2017-06-19
[Fix] Two minor fixes

navtej:
1. Ignore registry value "Other" from cymru. Provides a workaround #996 
2. Came across two types of timestamp for OTX feed
```
"modified": "2017-05-20T13:43:54"
"modified": "2017-05-20T13:43:28.563000"
```
I am not sure if this was one time system error or their lib strips sub-seconds part if it is perfect 0?
The check here fixes it only for modified time entry and other timestamps are left unchanged.  Test case for this is added as well.

### DIFF
--- a/intelmq/tests/bots/parsers/alienvault/test_parser_otx.py
+++ b/intelmq/tests/bots/parsers/alienvault/test_parser_otx.py
@@ -46,8 +46,8 @@ attacks.""".replace('\n', ' '),
 
 EXAMPLE_EVENT_2 = {
   '__type': 'Event',
-   'classification.type': 'blacklist',
-      'comment': 'Active users of mobile banking apps should be aware of a new '
+  'classification.type': 'blacklist',
+  'comment': 'Active users of mobile banking apps should be aware of a new '
                  'Android banking malware campaign targeting customers of large '
                  'banks in the United States, Germany, France, Australia, Turkey, '
                  'Poland, and Austria. This banking malware can steal login '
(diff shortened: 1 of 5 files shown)


intelmq/tests/bots/parsers/alienvault/test_parser_otx.py
```diff
@@ -47,15 +47,15 @@
...
-  'comment': 'Active users of mobile banking apps should be aware of a new '
-             'Android banking malware campaign targeting customers of large '
-             'banks in the United States, Germany, France, Australia, Turkey, '
-             'Poland, and Austria. This banking malware can steal login '
-             'credentials from 94 different mobile banking apps. Due to its '
-             'ability to intercept SMS communications, the malware is also '
-             'able to bypass SMS-based two-factor authentication. '
-             'Additionally, it also contains modules to target some popular '
-             'social media apps.',
+      'comment': 'Active users of mobile banking apps should be aware of a new '
```
wagner-certat: Why did you indent here by *3* spaces?

intelmq/tests/bots/parsers/alienvault/test_parser_otx.py
```diff
@@ -68,6 +68,37 @@
   'time.source': '2016-11-03T20:01:00+00:00'
   }
 
+EXAMPLE_EVENT_3 = {
+  '__type': 'Event',
+   'classification.type': 'blacklist',
+      'comment': 'Myanmar is a country currently engaged in an important '
```
wagner-certat: not the same indentation as the line above

navtej:
1. added 6to4 address test case
2. fixed the indentation.

Issue #982, opened 2017-05-30
Catch permission errors


wagner-certat:
Calls to psutil might raise `psutil.AccessDenied` as reported in #981 This should be handled and logged properly.

PR #1012, opened 2017-06-19
catch permission errors in psutil

wagner-certat:
fixes certtools/intelmq#982


wagner-certat:
no feedback -> merging

PR #1013, opened 2017-06-19
New harmonization type ASN

wagner-certat:
It checks for the valid range:
* 0 < asn < 4294967295 and
* asn != 65535


dmth:
Hi, 
should 4294967295 and 65535 really be excluded? From my understanding they are "should not" use internally and "must not" use globally. So when only getting reports for your internal networks, it still might happen that these ASN are used.

wagner-certat:
>  should 4294967295 and 65535 really be excluded? From my understanding they are "should not" use internally and "must not" use globally.

True. Although there are explicit ranges for internal use, it's not forbidden to use these two addresses internally.

intelmq/lib/harmonization.py
```diff
@@ -413,6 +413,39 @@ def sanitize(value):
...
+
+    Only valid are: 0 < asn <= 4294967295
+    See https://en.wikipedia.org/wiki/Autonomous_system_(Internet)
+    > The first and last ASNs of the original 16-bit integers, namely 0 and
+    > 65,535, and the last ASN of the 32-bit numbers, namely 4,294,967,295 are
+    > reserved and should not be used by operators.
+    """
+    @staticmethod
+    def check_asn(value):
+        if not (0 < value <= 4294967295):
```
aaronkaplan: can we reformulate this to a more reasonable: 
```
if (0 < value <= 42...):
   return True
else
  return False

?

PR #1014, opened 2017-06-20
Revised RIPE abuse contact expert, fix behavior of overwriting an event

wagner-certat:
* can now optionally replace existing abuse contact
* split parameter `query_ripe_stat` into `query_ripe_stat_asn` and `query_ripe_stat_ip`
* better error handling
* lib/message: overwriting something with None now has the same behavior as adding sth with None:
  if you overwrite an existing value with None/empty string, it will be deleted. Previously this did nothing.
* small fixes/additions in the tests library

and the test coverage of this bot is now 100% incl. all branches :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ development
 ### Development
 - We are now testing with and without optional libraries/lowest recommended versions and most current versions of required libraries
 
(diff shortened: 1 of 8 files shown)


wagner-certat:
no feedback -> merging

PR #1019, opened 2017-06-27
[enh] bypss mechanism for deduplicator

navtej:
Add boolean parameter `bypass`. If set the deduplicator will simply pass along the msgs bypassing deduplication mechanism. Useful in case you dont want to change the configuration but still want to bypass deduplicator temporarily.

### DIFF
--- a/intelmq/bots/experts/deduplicator/expert.py
+++ b/intelmq/bots/experts/deduplicator/expert.py
@@ -45,17 +45,17 @@ class DeduplicatorExpertBot(Bot):
 
         if self.bypass:
             self.send_message(message)
-
-        message_hash = message.hash(filter_keys=self.filter_keys,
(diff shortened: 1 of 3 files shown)


navtej:
travis-ci failed because of ```The command "sudo apt-get install dpkg-dev dh-python python-setuptools python3-setuptools python3-all debhelper quilt polipo lighttpd fakeroot" failed and exited with 100 during .

Your build has been stopped.```

intelmq/bots/experts/deduplicator/expert.py
```diff
@@ -15,6 +15,8 @@
 
     filter_type: string ["blacklist", "whitelist"]
 
+    bypass: boolean
```
wagner-certat: please add that the default is true

PR #1024, opened 2017-06-28
Add note about intelmqctl run to FAQ

chorsley:


aaronkaplan:
make sense . Thanks for the documentation patch! very very valuable.

PR #1028, opened 2017-06-29
Adds config for Shadowserver Accessible SMB-Feed

dmth:
@cert-bund and @th-certbund have created a configuration for shadowservers accessible smb feed.

 I've tested the config with yesterdays dataset for germany and it works well.


dmth:

aaronkaplan:
thanks, I'll change that back, no need to change it in the code

PR #1029, opened 2017-06-29
Dev shadowserver config codecleanness

bernhardreiter:
As  described in 
 * https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas
 * https://stackoverflow.com/questions/11597901/why-does-python-allow-a-trailing-comma-in-list#11597911

it is cool to use trailing commas in lists, dicts and some other python types.
So we shall do so consistently. (This is branched off https://github.com/Intevation/intelmq/tree/dev-shadowserver-Accessible-SMB to allow accepting the other first.)


dmth:
Hi,
this PR does also contain the code of PR #1028

dmth:

Issue #1022, opened 2017-06-27
wrong reverse dns lookups in shadowserver reports cause a crash of the FQDN-validation


bernhardreiter:
Right now, if a reverse dns returns something bogus, there is a crash and dump 
in sanitize()

Happened with shadowserver opendns resolver, and the hostname field was something like
`.example.net` 

```
 'source_queue': 'shadowserver-parser-dns-open-resolvers-queue',
 'traceback': ['Traceback (most recent call last):',
               '  File "/usr/lib/python3.4/encodings/idna.py", line 165, in '
               'encode',
               '    raise UnicodeError("label empty or too long")',
               'UnicodeError: label empty or too long',
               '',
               'The above exception was the direct cause of the following '
               'exception:',
               '',
               'Traceback (most recent call last):',
               '  File "/usr/lib/python3/dist-packages/intelmq/lib/bot.py", '
               'line 606, in process',
               '    events = list(filter(bool, self.parse_line(line, 
report)))',
               '  File '
               '"/usr/lib/python3/dist-packages/intelmq/bots/parsers/shadowserver/parser.py", '
               'line 150, in parse_line',
               '    event.add(intelmqkey, value)',
               '  File '
               '"/usr/lib/python3/dist-packages/intelmq/lib/message.py", 
line '
               '202, in add',
               '    value = self.__sanitize_value(key, value)',
               '  File '
               '"/usr/lib/python3/dist-packages/intelmq/lib/message.py", 
line '
               '298, in __sanitize_value',
               '    return class_reference().sanitize(value)',
               '  File '
               '"/usr/lib/python3/dist-packages/intelmq/lib/harmonization.py", '
               'line 374, in sanitize',
               "    return value.encode('idna').decode().lower()",
               "UnicodeError: encoding with 'idna' codec failed 
(UnicodeError: "
               'label empty or too long)']}
```

dmth:
harmonization.py does this, to validate "FQDN" -Data. As you can see, leading dots are not stripped.

``` python
    @staticmethod
    def sanitize(value):
        value = value.rstrip('.')
        if value:
            return value.encode('idna').decode().lower()
```

Minimalistic example to reproduce the error:
```python
>>> badurl = ".example.com"
>>> goodurl = "example.com"

>>> goodurl.encode("idna")
b'example.com'

>>> badurl.encode("idna")
Traceback (most recent call last):
  File "/usr/lib/python3.4/encodings/idna.py", line 165, in encode
[...]

Issue #1030, opened 2017-06-30
Failing FQDN-Validation


dmth:
FQDN-Validation fails with an exception, if the domain starts with a `.` character.

Details and ways to reproduce are mentioned in 
https://github.com/certtools/intelmq/issues/1022#issuecomment-311611767

This issue aims for a universal solution of this problem, whilst the aforementioned #1022 tries to resolve the shadowserver-parser specific problem.

PR #1036, opened 2017-07-06
Bugfix: Strip leading dots from FQDN in Shadowserver config

dmth:
Extends the validate_fqdn method by removing trailing dots from
the fqdn.
This fixes https://github.com/certtools/intelmq/issues/1022 on the
shadowserver-parser config level.

All "hostname" carrying values are now validated using this function.

Nevertheless this issue should still be discussed as it is present for all
other feeds. See https://github.com/certtools/intelmq/issues/1030 for the
general solution.

### DIFF
--- a/intelmq/bots/parsers/shadowserver/config.py
+++ b/intelmq/bots/parsers/shadowserver/config.py
@@ -171,8 +171,14 @@ def validate_ip(value):
 
 
 def validate_fqdn(value):
-    if harmonization.FQDN.is_valid(value, sanitize=True):
-        return value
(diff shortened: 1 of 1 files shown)


wagner-certat:
Isn't this this obsoleted by the much more generic solution in c24bb06b13073d6bce4be7bb829f3acd5f2d541f?

(Sorry for overseeing this PR in the first place!)

dmth:
Obosleted by 
https://github.com/certtools/intelmq/commit/c24bb06b13073d6bce4be7bb829f3acd5f2d541f

PR #1037, opened 2017-07-07
Add Harmonization Type for urlpath

navtej:
This PR 

- adds hamonization type for urlpath as `source.urlpath` and `destination.urlpath`. A use case for this is to consume IoCs for HTTP/Web accessible shells and backdoors.
- fixes issues with URI handling in OTX feed where a URI can contain both a url like `http://test.com/x/y/z` or just path eg `/x/y/z`.
- test cases for above.


wagner-certat:
I changed the destination branch to develop.

intelmq/bots/parsers/alienvault/parser_otx.py
```diff
@@ -61,12 +61,18 @@ def process(self):
...
                     resource = indicator["indicator"] \
                         if '://' in indicator["indicator"] \
                         else 'http://' + indicator["indicator"]
-                    event.add('source.url', resource)
+                    uri_added = event.add('source.url', resource, raise_failure=False)
+                    if not uri_added:
+                        if indicator["type"] == 'URI':
+                            event.add('source.urlpath', indicator["indicator"])
+                        else:
+                            raise ValueError("Invalid value <%s> for URL hamonization type" % indicator["indicator"])
```
wagner-certat: Missing dot at the end. Better use `%r` instead of `<%s>`.

wagner-certat:
Only the commits 2c52c8b and 1b84d44 can be merged into develop, all others are misplaced.

As always with the OTX parser: please @e3rd and @robcza have a short look if that's ok.

As it's a harmonization change also @aaronkaplan

navtej:
Please see #1039 . This branch was pulled from master before  develop was around. #1039 is cleaner.

PR #1039, opened 2017-07-17
add urlpath harmonization type

navtej:
Bringing this from #1037 

This PR 

- adds hamonization type for urlpath as `source.urlpath` and `destination.urlpath`. A use case for this is to consume IoCs for HTTP/Web accessible shells and backdoors.
- fixes issues with URI handling in OTX feed where a URI can contain both a url like `http://test.com/x/y/z` or just path eg `/x/y/z`.
- test cases for above.
- adds `source.urlpath` and `destination.urlpath` in initdb.sq
- add information to changelog and news
- add information to upgrade db.

### DIFF
--- a/intelmq/bots/parsers/alienvault/parser_otx.py
+++ b/intelmq/bots/parsers/alienvault/parser_otx.py
@@ -61,12 +61,17 @@ class AlienVaultOTXParserBot(Bot):
                 elif indicator["type"] == 'email':
                     event.add('source.account',
                               indicator["indicator"])
-                # URLs
+                # URLs/URIs, OTX URIs can contail both full url or only path
                 elif indicator["type"] in ['URL', 'URI']:
                     resource = indicator["indicator"] \
                         if '://' in indicator["indicator"] \
                         else 'http://' + indicator["indicator"]
-                    event.add('source.url', resource)
+                    uri_added = event.add('source.url', resource, raise_failure=False)
+                    if not uri_added:
+                        if indicator["type"] == 'URI':
+                            event.add('source.urlpath', indicator["indicator"])
+                        else:
+                            raise ValueError("Invalid value %r for URL hamonization type." % indicator["indicator"])
                 # CIDR
                 elif indicator["type"] in ['CIDR']:
                     event.add('source.network',
--- a/intelmq/bots/parsers/alienvault/parser_otx.py
+++ b/intelmq/bots/parsers/alienvault/parser_otx.py
@@ -61,7 +61,7 @@ class AlienVaultOTXParserBot(Bot):
                 elif indicator["type"] == 'email':
                     event.add('source.account',
                               indicator["indicator"])
-                # URLs/URIs, OTX URIs can contail both full url or only path
+                # URLs/URIs, OTX URIs can contain both full url or only path
                 elif indicator["type"] in ['URL', 'URI']:
                     resource = indicator["indicator"] \
                         if '://' in indicator["indicator"] \
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ CHANGELOG
 
 1.1.0
 -----
+### Harmonization
+- added destination.urlpath and source.urlpath to harmonization.
 
 1.0.0.rc2 Release candidate
 ---------------------------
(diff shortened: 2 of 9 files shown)


navtej:
@wagner-certat  bringing conversation from #1037 
For sanity check, urlpath is bare bones string. I wonder what sort of checks you have in mind, it may or may not start/end with `/`, may contain almost anything from charset.

I have added CHANGELOG and NEWS iteams alongwith upgrade statement, please cross check those

adding @aaronkaplan  @e3rd @robcza  for reviews.

navtej:
build failed because of ripe checks `ERROR: Test RIPE DB AS for errors.`

wagner-certat:
>  For sanity check, urlpath is bare bones string. I wonder what sort of checks you have in mind, it may or may not start/end with /, may contain almost anything from charset.

Is `/foo.html` the same as `foo.html`. IMHO yes, so we can always ensure the path starts with a a `/`.
Is `/bar/` the same as `/bar`? IMHO no.
Should encoded ASCII characters like `%20` be decoded? Browsers normalize these values AFAIK.

navtej:
> Is `/foo.html` the same as `foo.html`. IMHO yes, so we can always ensure the path starts with a a /.
It depends on one's perspective, to you it appears same, i would say No. `/foo.html` can be considered absolute path, whereas `foo.html` may be relative path, as in `/x/y/z/foo.html`. However if you believe we should always treat `urlpath` as possible subpath, we can take the sanitization approach. I would say lets not do that.
> Is /bar/ the same as /bar? IMHO no.
> Should encoded ASCII characters like %20 be decoded? Browsers normalize these values AFAIK.
I am not sure if we should do that. Consider the IIS urldecode bug [CVE-2001-0333,](https://marc.info/?l=bugtraq&m=98992056521300&w=2) it would be better to save the contents as they are. Same case for possible validation bypasses [CAPEC-80: Using UTF-8 Encoding to Bypass Validation Logic](https://capec.mitre.org/data/definitions/80.html)

wagner-certat:
I cherry-picked the commit which fixes the ripencc test.

> It depends on one's perspective, to you it appears same, i would say No. `/foo.html` can be considered absolute path, whereas `foo.html` may be relative path, as in `/x/y/z/foo.html`. However if you believe we should always treat `urlpath` as possible subpath, we can take the sanitization approach. I would say lets not do that.

So it's like a fragment, not always a full path.

> I am not sure if we should do that. Consider the IIS urldecode bug [CVE-2001-0333,](https://marc.info/?l=bugtraq&m=98992056521300&w=2) it would be better to save the contents as they are. Same case for possible validation bypasses [CAPEC-80: Using UTF-8 Encoding to Bypass Validation Logic](https://capec.mitre.org/data/definitions/80.html) 

Very interesting, thanks!

wagner-certat:
Talked with @aaronkaplan about this and we concluded that we cannot apply any checks/normalization because the content of this field can be anything.

Still waiting for a review from some else from the core team (probably @aaronkaplan?) as this is a harmonization change.

intelmq/bots/parsers/alienvault/parser_otx.py
```diff
@@ -61,12 +61,17 @@ def process(self):
                 elif indicator["type"] == 'email':
                     event.add('source.account',
                               indicator["indicator"])
-                # URLs
+                # URLs/URIs, OTX URIs can contail both full url or only path
```
aaronkaplan: "contain"

wagner-certat:
@e3rd or @robcza could you check the alienvault-related changes please?

intelmq/bots/parsers/alienvault/parser_otx.py
navtej: fixed. sorry for the delay

wagner-certat:
reminder @e3rd or @robcza

wagner-certat:
And @navtej this conflicts with master because of the conversion to a ParserBot

e3rd:
> reminder @e3rd or @robcza  

Sorry for not reacting so long. :) It seemed ok to me so that I "liked" your request from 19 Jul.

wagner-certat:
> It seemed ok to me so that I "liked" your request from 19 Jul.

I wasn't sure what it did mean.

Ok, will propose changes to that PR to make it mergeable with master.

wagner-certat:
merged with my changes/additions

navtej:
Thanks and apologies for not being able to chip in.

PR #1040, opened 2017-07-17
shadowserver smb parser: smb_implant as bool

wagner-certat:
Converts the field `smb_implant` to bool


th-certbund:
Sure, bool is fine here.

PR #1041, opened 2017-07-18
Mail collectors: sent from and new Parameters.get method

e3rd:
When there is a shared mailbox for multiple feeds, the collector bots gets very slow when iterating through all the messages encountered. It is far more easy to filter by sender.

I've added a method `get` to Parameters object so that it implements this cool functionality from built-in dict. Thanks to that I cat write: `sent_from=self.parameters.get("sent_from")` instead of much longer `self.parameters.sent_from if hasattr(self.parameters, "sent_from") else None`. 

Is that alright or self.parameters should be a dict from the beginning? I don't see the purpose of self.parameters being an bare object.


intelmq/lib/bot.py
```diff
@@ -715,4 +715,6 @@ def new_report(self):
 
 
 class Parameters(object):
-    pass
+
+    def get(self, parameter, default=None):
+        return getattr(self, parameter) if hasattr(self, parameter) else default
```
wagner-certat: Shorter: `return getattr(self, parameter, default)`

e3rd: Of course! Much cleaner, I'm removing this method.

intelmq/bots/collectors/mail/collector_mail_attach.py
```diff
@@ -22,7 +22,9 @@ def process(self):
                               self.parameters.mail_user,
                               self.parameters.mail_password,
                               self.parameters.mail_ssl)
-        emails = mailbox.messages(folder=self.parameters.folder, unread=True)
+        emails = mailbox.messages(folder=self.parameters.folder, unread=True,
+                                  sent_to=getattr(self.parameters, "sent_to", None),
```
wagner-certat: The [default values used in imbox](https://github.com/martinrusev/imbox/blob/master/imbox/query.py#L20) are `False` not `one`None`, but I guess the result is the same?

e3rd: Yes, it seemed to me like that. If I put None, the parameter is not used.

PR #1043, opened 2017-07-19
BUG: ctl: handle already removed PID

wagner-certat:
When running `intelmqctl run bot-id` and stopping/killing bot-id in parallel, this threw a FileNotFound exception. As we can't handle this in `bot_run` (It does not know who killed it), we need to implement this check in `bot_stop`.

### DIFF
--- a/intelmq/bin/intelmqctl.py
+++ b/intelmq/bin/intelmqctl.py
@@ -204,7 +204,10 @@ class IntelMQProcessManager:
                 if self.__status_process(pid):
                     log_bot_error('running', bot_id)
                     return 'running'
-                self.__remove_pidfile(bot_id)
+                try:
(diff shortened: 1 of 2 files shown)


intelmq/bin/intelmqctl.py
```diff
@@ -204,7 +204,10 @@ def bot_stop(self, bot_id, getstatus=True):
                 if self.__status_process(pid):
                     log_bot_error('running', bot_id)
                     return 'running'
-                self.__remove_pidfile(bot_id)
+                try:
+                    self.__remove_pidfile(bot_id)
```
aaronkaplan: Isn't the try except something we should catch *inside* the __remove_pidfile function?

Issue #981, opened 2017-05-26
Bots Started by IntelMQ Manager Won't Stop


mf370:
I managed to install IntelMQ in a CentOS 7. And I used the following commands:
```
yum install epel-release
yum install python34 python34-devel
yum install git gcc gcc-c++
yum install redis
curl "https://bootstrap.pypa.io/get-pip.py" -o "/tmp/get-pip.py"
python3.4 /tmp/get-pip.py
systemctl enable redis
systemctl start redis
pip3 install typing

git clone https://github.com/certtools/intelmq.git /tmp/intelmq
cd /tmp/intelmq
sudo -s

pip3 install -r REQUIREMENTS
pip3 install .

useradd -d /opt/intelmq -U -s /bin/bash intelmq
chmod -R 0770 /opt/intelmq
chown -R intelmq.intelmq /opt/intelmq
cd /opt/intelmq/etc
cp -a examples/* .
```
Then I had to copy the files `.bashrc` and ` .bash_profile` from the directory `/root` to the directory `/opt/intelmq/` because everytime I tried to change to the intelmq user, instead of appearing `[intelmq@localhost]$` it would appear `bash-4.2$`.

After this I installed IntelMQ Manager with the following commands:
```
yum install git httpd php
systemctl enable httpd
systemctl start httpd

git clone https://github.com/certtools/intelmq-manager.git /tmp/intelmq-manager
cp -R /tmp/intelmq-manager/intelmq-manager/* /var/www/
chown -R apache.apache /var/www/html/
```
```
sed -i 's|Defaults    requiretty|Defaults    !requiretty|g' /etc/sudoers

usermod -a -G intelmq apache
```

I added this line to the visudo file:
`apache ALL=(intelmq) NOPASSWD: /usr/bin/intelmqctl`

```
sed -i 's|#!/usr/bin/env python|#!/usr/bin/env python3.4|g' /usr/bin/intelmqctl
sed -i 's|SELINUX=enforcing|SELINUX=disabled|g' /etc/selinux/config

firewall-cmd --add-service http
firewall-cmd --add-service http --permanent

reboot
```

And the problem is: if I start/stop the bots using` intelmqctl ` it works perfectly, but if I start the bots using IntelMQ Manager they will start normally, but if I try to stop them it won't do anything. I also tried to stop them with `intelmqctl` after they were started by IntelMQ Manager and this is what happens:

```
[intelmq@localhost intelmq]$ intelmqctl stop ip-fileinput-collector
[...]

PR #1044, opened 2017-07-19
Use SIGTERM to stop bots

wagner-certat:
Sigint can be captured by parent processes

fixes certtools/intelmq#981
fixes certtools/intelmq-manager#103

See also https://unix.stackexchange.com/questions/149741/why-is-sigint-not-propagated-to-child-process-when-sent-to-its-parent-process

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ CHANGELOG
 
 1.0.0.rc2 Release candidate
 ---------------------------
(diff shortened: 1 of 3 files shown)


aaronkaplan:
For a difference of the signals, see: https://www.quora.com/What-is-the-difference-between-the-SIGINT-and-SIGTERM-signals-in-Linux

Issue #791, opened 2016-11-24
Expand the extra field in output bots


sykaeh:
Currently the extra field is a JSON object encoded as as string. This leads to some (for me) unexpected behaviour when you output the event in JSON format (like in the file output) because you then have a JSON object with one field that is actually another JSON object, but instead of displaying it as such it is encoded as a string. I understand that in some cases you want it as a string, but especially if you are using hierarchical output it does not make much sense to keep it as a string.

I would suggest to add another option (`expand_extra`) to the `to_dict` and `to_json` methods of the `Message` which would expand the field (using `json.loads`) instead of leaving at as string. 

We would then also have to add another parameter to relevant output bots. Making it false by default would ensure that we do not break any existing installations.

This is also somewhat related to #671 . If the field is properly unserialized into a Python dictionary when retrieving it, then it also makes it easier to modify the extra field.

sebix:
Also related: #793

wagner-certat:
As the `hierarchical_output` parameter is not honored this is actually a bug. Thus I don't think it's clever to fix that together with the proper de/serialization #671 as this would be a bigger change.

I'm working on #793 and this bug now.

PR #1046, opened 2017-07-25
Access extra.* like normal fields

wagner-certat:
You can now set and get extra-fields just as all others #793 :

```python
event['extra.foo'] = 'bla'
event['extra.foo'] # gives 'bar'
```

This is not limited to `extra.*`, it applies to anything field with type `JSONDict`

It now behaves correctly with the hierarchical keyword see #791

A new harminization type is introduced, the `JSONDict` (like the former `JSON` type, can only hold dicts as JSON) and `JSON` itself is not more relaxed (allows not only dicts, but any valid JSON). The type `JSON` is used for the field `output`.

For backwards compatibility, you can still set extra to a string holding a JSON, however I don't recommend it.

I did not yet deprecate anything (setting and getting extra as whole) because I think we need to discuss this better, also for other nested fields.


wagner-certat:

PR #1050, opened 2017-08-02
DOCS: fixes toc link to System Configuration

bernhardreiter:


wagner-certat:
Thanks.

The link to the defaults-section is broken too

wagner-certat:
Will also be in 1.0

PR #1070, opened 2017-08-10
Added modify rule for ramnit

th-certbund:
Added modify expert rule for different "ramnit" malware names (like "ramnit-0xe371bf19") in Shadowserver Botnet Drone feed.

### DIFF
--- a/intelmq/bots/experts/modify/examples/morefeeds.conf
+++ b/intelmq/bots/experts/modify/examples/morefeeds.conf
@@ -226,6 +226,15 @@
         }
     },
     {
+        "rulename": "ramnit",
+        "if": {
+            "malware.name": "^ramnit-0x.*$"
+        },
+        "then": {
+            "classification.identifier": "ramnit"
+        }
+    },
+    {
         "rulename": "default",
         "if": {
             "malware.name": ".*",


intelmq/bots/experts/modify/examples/morefeeds.conf
```diff
@@ -226,6 +226,15 @@
         }
     },
     {
+        "rulename": "ramnit",
+        "if": {
+            "malware.name": "^ramnit-0x.*$"
```
wagner-certat: Does the pattern end with a hex number?

th-certbund: Yes. Sample malware names from recent botnet drone feeds:
ramnit-0x4ec815e2
ramnit-0x606d35bf

So a more strict rule would be "^ramnit-0x[0-9a-f]{8}$" but I don't know if this is really necessary and it will probably fail to catch malware names if the length of the hex value changes.

wagner-certat: > So a more strict rule would be "^ramnit-0x[0-9a-f]{8}$" but I don't know if this is really necessary and it will probably fail to catch malware names if the length of the hex value changes.

So what about `"^ramnit-0x[0-9a-f]*$"`? This would not match malfomed values like `"ramnit-0x00 or something else"` or `"ramnit-0x123;foobar"`

th-certbund: I'm fine with "^ramnit-0x[0-9a-f]*$"

wagner-certat:
Thanks!

Issue #1067, opened 2017-08-10
JSON outpot for ctl check


wagner-certat:
needed for certtools/intelmq-manager#118

Issue #1068, opened 2017-08-10
suppress log output for clear queue


wagner-certat:
Current output:

> intelmqctl: Clearing queue abusech-domain-parser-queue
> intelmqctl: Successfully cleared queue abusech-domain-parser-queue.
> "success"

See also certtools/intelmq-manager#54

PR #1076, opened 2017-08-17
JSON output fixes

wagner-certat:
fixes bugs #1067 and #1068


wagner-certat:

Issue #1073, opened 2017-08-16
ctl does not check if a running process is really the bot


wagner-certat:
when intelmqctl checks if a bot is running it looks up the pid and then looks if a processes with this pid is running. ctl does not check if the process is really the bot but something else.

To reproduce create a pid file of any configured bot and write some PID of another running process into it. intelmqctl will say the bot runs and can't kill it because of missing permissions. The latter is correct, the former not.

The solution would be to remove the pid if the running process executable does not match the expected executable.

PR #1080, opened 2017-08-17
check commandline of running process for status

wagner-certat:
fixes certtools/intelmq#1073


wagner-certat:

Issue #1079, opened 2017-08-17
File output does not close file


wagner-certat:
A ResourceWarning is thrown, but these are ignored by default. Set a environment variable to see them:

```
> export PYTHONWARNINGS=default
> intelmqctl run file-output 
/usr/lib64/python3.4/importlib/_bootstrap.py:579: ImportWarning: Not importing directory /usr/lib64/python3.4/site-packages/zope: missing __init__
  _warnings.warn(msg.format(portions[0]), ImportWarning)
/usr/lib64/python3.4/importlib/_bootstrap.py:579: ImportWarning: Not importing directory /usr/lib64/python3.4/site-packages/mpl_toolkits: missing __init__
  _warnings.warn(msg.format(portions[0]), ImportWarning)
/usr/lib64/python3.4/importlib/_bootstrap.py:579: ImportWarning: Not importing directory /usr/lib/python3.4/site-packages/logilab: missing __init__
  _warnings.warn(msg.format(portions[0]), ImportWarning)
intelmqctl: Starting file-output...
file-output: FileOutputBot initialized with id file-output and intelmq 1.1.0.alpha1 and python 3.4.6 (default, Mar 22 2017, 12:26:13) [GCC] as process 19704.
file-output: Bot is starting.
file-output: File '/opt/intelmq/var/lib/bots/file-output/events.txt' is open.
file-output: Loading source pipeline and queue 'file-output-queue'.
file-output: Connected to source queue.
file-output: No destination queues to load.
file-output: Pipeline ready.
file-output: Waiting for incoming message.
^Cfile-output: Received KeyboardInterrupt.
file-output: Disconnected from source pipeline.
file-output: Bot stopped.
Bot exited with code 0.
sys:1: ResourceWarning: unclosed file <_io.TextIOWrapper name='/opt/intelmq/var/lib/bots/file-output/events.txt' mode='at' encoding='utf-8'>
```

PR #1081, opened 2017-08-17
BUG: file output: close file on shutdown

wagner-certat:
fixes certtools/intelmq#1079


wagner-certat:

PR #1083, opened 2017-08-25
Sieve Filter expert

digihash:
This is for the Sievebot which was written in collaboration with @antoinet and @helderfernandes1279.

The sieve bot is used to filter and/or modify events based on a set of rules. The rules are specified in an external configuration file and with a syntax similar to the Sieve language used for mail filtering.

Each rule defines a set of matching conditions on received events. Events can be matched based on keys and values in the event. If the processed event matches a rule's conditions, the corresponding actions are performed. Actions can specify whether the event should be kept or dropped in the pipeline (filtering actions) or if keys and values should be changed (modification actions).

For more information, see README.md under _"intelmq/bots/experts/sieve/README.md"_


antoinet:
I fixed the codestyle checks in https://github.com/antoinet/intelmq/commit/17c0d4d1d70903ff08f00a33ab6cb1a3276dc220
@digihash: can you add this commit to the PR?

wagner-certat:
@antoinet Thanks. If you push to antoinet:develop, they are part of this PR

antoinet:
I think I got all tests fixed. CI is still failing, because coincidentally the cymru-whois bot is failing at the moment (asn.cymru.com is not resolvable).

intelmq/tests/bots/experts/sieve/test_sieve_files/test_add.sieve
```diff
@@ -0,0 +1,3 @@
+if comment == 'add field' {
+    add destination.ip="150.50.50.10"
+}
```
wagner-certat: missing newlines everywhere

intelmq/bots/experts/sieve/expert.py
```diff
@@ -0,0 +1,229 @@
...
+        if value.__class__.__name__ == 'SingleNumericValue':
+            return self.process_numeric_operator(event[key], op, value.value)
+        elif value.__class__.__name__ == 'NumericValueList':
+            for val in value.values:
+                if self.process_numeric_operator(event[key], op, val.value):
+                    return True
+            return False
+
+    def process_numeric_operator(self, lhs, op, rhs):
+        return eval(str(lhs) + op + str(rhs))  # TODO graceful error handling
```
wagner-certat: Is it possible without eval? At least a strict checking of the arguments should be done.

intelmq/bots/experts/sieve/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.dot
```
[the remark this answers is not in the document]
aaronkaplan: because they might have used graphviz :) Anyway, yes. Not needed here.

intelmq/bots/experts/sieve/README.md
```diff
@@ -0,0 +1,166 @@
...
+The sieve bot is used to filter and/or modify events based on a set of rules. The
+rules are specified in an external configuration file and with a syntax similar
+to the [Sieve language](http://sieve.info/) used for mail filtering.
+
+Each rule defines a set of matching conditions on received events. Events can be
+matched based on keys and values in the event. If the processed event matches a
+rule's conditions, the corresponding actions are performed. Actions can specify
+whether the event should be kept or dropped in the pipeline (filtering actions)
+or if keys and values should be changed (modification actions).
+
```
aaronkaplan: Would it be possible to specify a specific outgoing pipeline as an action?

Example: 
```
  if $condition
  then
    send-to "pipeline-A";
  fi
```

intelmq/bots/experts/sieve/README.md
```diff
@@ -0,0 +1,166 @@
...
+} else {
+  remove comment
+}
+```
+
+
+## Parameters
+
+The sieve bot takes only one parameter:
+ * `file` - filesystem path the the sieve file
```
aaronkaplan: grammar: file system path of the sieve file

intelmq/bots/experts/sieve/README.md
```diff
@@ -0,0 +1,166 @@
...
+if :exists source.fqdn {
+  keep  // aborts processing of subsequent rules and forwards the event.
+}
+
+
+if :notexists source.abuse_contact || source.abuse_contact =~ '.*@example.com' {
+  drop  // aborts processing of subsequent rules and drops the event.
+}
+
+if source.ip << 192.0.0.0/24 {
```
aaronkaplan: according to the docs below, IP addresses must be quoted.

intelmq/bots/experts/sieve/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+ * `:exists` and `:notexists` match if a given key exists, for example:
+
+    ```if :exists source.fqdn { ... }```
+
+ * `==` and `!=` match for equality of strings and numbers, for example:
+
+   ```if feed.name != 'acme-security' || feed.accuracy == 100 { ... }```
+
+ * `:contains` matches on substrings.
```
aaronkaplan: wish list: if we have contains, out of reasons of symmetry , may we have ``:notcontains`` ?

intelmq/bots/experts/sieve/README.md
```diff
@@ -0,0 +1,166 @@
...
+ * Numerical comparisons are evaluated with `<`, `<=`, `>`, `>=`.
+
+ * `<<` matches if an IP address is contained in the specified network range:
+
+   ```if source.ip << '10.0.0.0/8' { ... }```
+
+ * Values to match against can also be specified as list, in which case any one
+ of the values will result in a match:
+
+   ```if source.ip == ['8.8.8.8', '8.8.4.4'] { ... }```
```
aaronkaplan: may we change the grammar here to ``:in $list`` ? Ah, same comment as sebix :)

intelmq/bots/experts/sieve/README.md
```diff
@@ -0,0 +1,166 @@
...
+  In this case, the event will match if it contains a key `source.ip` with
+  either value `8.8.8.8` or `8.8.4.4`.
+
+
+### Actions
+
+If part of a rule matches the given conditions, the actions enclosed in `{` and
+`}` are applied. By default, all events that are matched or not matched by rules
+in the sieve file will be forwarded to the next bot in the pipeline, unless the
+`drop` action is applied.
```
aaronkaplan: Is there a possibility for a default action ?

intelmq/bots/experts/sieve/README.md
```diff
@@ -0,0 +1,166 @@
...
+ (same as the default behaviour), but in addition the sieve file processing is
+ interrupted upon reaching this action.
+
+ * `drop` marks the event to be dropped. The event will not be forwarded to the
+ next bot in the pipeline. The sieve file processing is interrupted upon
+ reaching this action. No other actions may be specified besides the `drop`
+ action within `{` and `}`.
+
+
+### Comments
```
aaronkaplan: Could you explain (maybe I missed it and you did) if 
  *  the order is relevant and 
  *  does the checking stop or continue after a match?

intelmq/bots/experts/sieve/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+ * `add` adds a key value pair to the event. This action only applies if the key
+ is not yet defined in the event. If the key is already defined, the action is
+ ignored. Example:
+
+   ```add comment = 'hello, world'```
+
+ * `add!` same as above, but will force overwrite the key in the event.
+
+ * `modify` modifies an existing value for a key. Only applies if the key is
```
aaronkaplan: nit-picking here: may we rename `modify` to `update` ?

intelmq/bots/experts/sieve/README.md
```diff
@@ -0,0 +1,166 @@
...
+   ```modify feed.accuary = 50```
+
+ * `remove` removes a key/value from the event. Action is ignored if the key is
+ not defined in the event. Example:
+
+    ```remove extra.comments```
+
+ * `keep` marks the event to be forwarded to the next bot in the pipeline
+ (same as the default behaviour), but in addition the sieve file processing is
+ interrupted upon reaching this action.
```
aaronkaplan: So implicitly here you are saying that the default behaviour without `keep` is to contineue with the matching ?

intelmq/bots/experts/sieve/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+ * `drop` marks the event to be dropped. The event will not be forwarded to the
+ next bot in the pipeline. The sieve file processing is interrupted upon
+ reaching this action. No other actions may be specified besides the `drop`
+ action within `{` and `}`.
+
+
+### Comments
+
+Comments may be used in the sieve file: all characters after `//` and until the end of the line will be ignored.
```
aaronkaplan: wish-list: ``//`` or ``#`` as a comment character.

intelmq/bots/experts/sieve/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+
+## Validating a sieve file
+
+Use the following command to validate your sieve files:
+```
+$ intelmq.bots.experts.sieve.validator
```
aaronkaplan: this should be called implicitly when the ``intelmqctl configtest`` gets called.

intelmq/bots/experts/sieve/expert.py
aaronkaplan: +1

intelmq/tests/bots/experts/sieve/test_sieve_files/test_ip_range_list_match.sieve
```diff
@@ -0,0 +1,3 @@
+if source.ip << ['192.0.0.0/24', '169.254.0.0/16'] {
```
aaronkaplan: this does not seem to be documented in the docu. I mean that the ``<<`` operator also works on lists.

aaronkaplan:
I move that we merge this soon!

intelmq/bots/experts/sieve/README.md
wagner-certat: It is `intelmqctl check`

intelmq/bots/experts/sieve/README.md
wagner-certat: This is not supported by the core.

intelmq/bots/experts/sieve/README.md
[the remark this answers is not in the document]
antoinet: The `*` is a marker that refers to the position in the expression where the error occurred.

intelmq/bots/experts/sieve/README.md
antoinet: In our instance, we are already (ab)using a newly introduced key called `keep` (boolean), which can be set anywhere in the pipeline. Downstream, events with `keep==False` are dropped.

This can be generalized by introducing a new key, e.g. `route`, with values that describe a specific destination. But this requires to put filter bots upstream of every destination such that only the corresponding events are kept.

If a routing model was provided in the core, we would certainly have use-cases for it.

intelmq/bots/experts/sieve/README.md
antoinet: Yes, the `keep` and `drop` actions explicitly interrupt the processing of the rules.

Without a mention of `keep` or `drop`, processing of subsequent rules continues. Afaik this is the same behavior as mail sieve.

intelmq/tests/bots/experts/sieve/test_sieve_files/test_ip_range_list_match.sieve
antoinet: Actually, every operator works on lists. I will point that out.

intelmq/bots/experts/sieve/expert.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def process_ip_range_match(self, key, ip_range, event):
+        if key not in event:
+            return False
+
+        try:
+            addr = ipaddress.ip_address(event[key])
+        except ValueError:
```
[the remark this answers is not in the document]
antoinet: Since the event type is not enforced (any key can be specified as the left hand side of the `<<` operator), the value is validated here. We did discuss whether or not to restrict this operator to keys with a type corresponding to an IP address (https://github.com/antoinet/intelmq/issues/11), and I just realized I didn't implement it the suggested way.

intelmq/bots/experts/sieve/README.md
```diff
@@ -0,0 +1,166 @@
...
+ * Numerical comparisons are evaluated with `<`, `<=`, `>`, `>=`.
+
+ * `<<` matches if an IP address is contained in the specified network range:
+
+   ```if source.ip << '10.0.0.0/8' { ... }```
+
+ * Values to match against can also be specified as list, in which case any one
+ of the values will result in a match:
+
+   ```if source.ip == ['8.8.8.8', '8.8.4.4'] { ... }```
```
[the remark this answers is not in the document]
antoinet: Problem is: every operator, not just `==`, supports a list as a right hand side. If `==` is renamed to `:in`, what should we do with the other operators?

intelmq/bots/experts/sieve/README.md
wagner-certat: I wonder why no issue existed for this which and created #1088 for it

antoinet:
@wagner-certat and @aaronkaplan: I think I addressed all review points so far up to the `:in` operator, used to check if an ip address matches a list of values, e.g.:
```source.ip :in ['127.0.0.1', '192.168.1.1', '10.0.0.1']```
As mentioned above, most of the operators support a list in the right-hand-side by default, e.g.:
```source.ip != ['127.0.0.1', '192.168.1.1', ...]```
So it is a design question:
 * we can introduce the `:in` operator (and by symmetry `:notin`) and drop list-support for all other operators (makes it simpler)
 * `:in` is just an alias for `==` (and `:notin` an alias for `!=`), and we keep list-support for all other operators.
What do you all think?

wagner-certat:
Concerning the `:in` and `==` and lists...

If we one day support multiple source IP addresses or similar (with the IDEA format or whatever) everything blows up anyway... So IMHO I am in favor of just using the currently implemented behavior with `==`.

@aaronkaplan What do you think?
And @dmth maybe you have the leisure to comment too? :D

SYNchroACK:
> Concerning the :in and == and lists...
> 
> If we one day support multiple source IP addresses or similar (with the IDEA format or whatever) everything blows up anyway... So I am in favour of just using the currently implemented behaviour with ==.

I understand, its well pointed @wagner-certat . However, I think we are far way from that day and also, the day we change to multi source IP addresses we need to change multiple experts bots. IMHO, we should keep with our principle of using one key = value, therefore, the presented code is fine.
Makes sense?

PR #1091, opened 2017-09-11
Call bot check

wagner-certat:
Bots can specify a static method `check(parameters)` which can perform individual checks specific to the bot. These functions will be called by `intelmqctl check` if the bot is configured with the given parameters.

Needed for antoinet/intelmq#20 / https://github.com/certtools/intelmq/pull/1083#discussion-diff-136086094R146


wagner-certat:

PR #1093, opened 2017-09-20
lib/Message: overwrite parameter enhancement

wagner-certat:
Description from the changelog:
> - intelmq.lib.message.Message.add: The parameter overwrite accepts now three different values: True, False and None (new).
>   True: An existing value will be overwritten
>  False: An existing value will not be overwritten (previously and exception has been raised when the value was raised).
>   None (default): If the value exists an KeyExists Exception is thrown (previously the same as False).
>   This allows shorter code in the bots, as an 'overwrite' configuration parameter can be directly passed to the function.

You can then pass a given configuration parameter directly to the `add()` function, for example here shown for the maxmind expert.


wagner-certat:

PR #1094, opened 2017-09-20
Packaging and LSB

wagner-certat:
I used these modifications/files for all the recent release builds, so it's known to work.

The build packages use LSB paths only.

This branch contains cherry-picked changes from the dev-debian branch

The current 1.0.1-2 release in unstable repo is based on this branch


wagner-certat:
Also applied to develop in 5a13675

PR #1095, opened 2017-09-20
University of Toulouse blacklist and .tar.gz support in HTTP collector

Narzhan:
This collector bot is written to get the blacklist info posted at https://dsi.ut-capitole.fr/blacklists/. The bot takes two input parameters which specify which collection should be downloaded and files from this collection that should be processed. The processing of this bot's output can be done using generic csv parser. So far the "malware" and "phishing" collections appear to yeild quite good results as they provide some interesting data with very little false positives.


wagner-certat:
Hi,

first thanks for your contribution and taking the time!

We recently removed the FTP(s) collectors in version 1.0 because no one used them, and in case of the FTPS collection because of licensing issues too.

I found out that they support HTTP(S) too which I'd favor to use: https://dsi.ut-capitole.fr/blacklists/download/

With HTTPS we could get rid of the FTP-code and we "only" need some support for extracting tar.gz files. And for this I'd favor some code in the generic parsers, and in this case the HTTP collector. In the long term I'd like to have some kind of intermediate processing bot between collector and parser for tasks like this, but that's out of scope here. ;)

So I propose to:
 * Add a parameter `extract_files` to the http collector with default None.
    * True means: extract every file
    * list of file names mean: extract these files
  * move the extract code to some library function, either utils or the CollectorBot and call it at the end of the collector bots

I guess that can be easily extended to support zip and similar.

Would that fit your needs?

Narzhan:
I guess that seems reasonable. However I would like to clarify some aspects. What happens with the files after they are extracted. Are they processed by some specific parser which afterwards processes the values or by the Http collector which than sends the file(s) to generic csv parser (as it is happening now).

wagner-certat:
> I guess that seems reasonable.

I'd like to keep the possible bugs and the maintenance burden (for you and me) low.

> What happens with the files after they are extracted. Are they processed by some specific parser which afterwards processes the values or by the Http collector which than sends the file(s) to generic csv parser (as it is happening now).

I am unsure if I understood the question.
The data is passed on the next bot as configured. So if you use the generic csv parser, it gets the extracted data as message. That part does not change and is independent of the collection method. For that parser, the data looks same, it does not know of the collector and how it works.

Narzhan:
I've changed the behaviour according to your proposition so the default collector bot now supports tar.gz files extraction. So far I've added the ability to extract only tar.gz files since http collector handles zip files by default, but I can add a function for zip file extraction. I've been thinking about creating more general function so that it could be more general and support more compressions. However I run into some problems since I don't know a lot about this particular field. I found that pyunpack and patool support almost all compression types imaginable, but I was unable to find a way to make it suit the proposed behaviour since I, so far, found that both can extract all files from the archive on disk to some other disk location, which would create a lot of temporary files for clean-up afterwards. I wanted to consult whether this is ok behaviour or not, if so I will edit the utils function to make it more universal.

wagner-certat:
I just found out that Python's [mimetype](https://docs.python.org/3/library/mimetypes.html) lib can't determine the type from file like objects, only from URLs and files :/ We can still guess it from the filename.

We have automatic unzipping for zip, but not for tar.gz. While the user can give a list of filenames for the latter but not the first. Not very consistent...
It is for sure necessary to set to be extracted files for zip too.
Concerning the automatic unzipping: I prefer the explicit approach, but we do it implicitly for zip.
So we could do automatic extracting if it is guessable by the file extension (parameter `extract_files` is `None`, default), do unzipping or un-tar.gz-ing if a list of files or `True` is given, or don't do it at all for `False`.
I'm inclined to ask some users on this.

I also wrote a test case for the HTTP collector including file extraction: https://github.com/whalebone/intelmq/compare/univ-toulouse...wagner-certat:whalebone-univ-toulouse?expand=1

A small todo for you though:
 * Please rebase your changes on the develop branch (or cherry-pick them to it), the master branch is always the latest stable release.
 * Please add the new parameters to the BOTS file.
 * Please add a description of the feed to docs/Feeds.md (including needed settings for collector and parser)

PR #1097, opened 2017-09-29
deduplication bypass default

e3rd:
I think, the default value should be False (as seen in the code at https://github.com/certtools/intelmq/blob/master/intelmq/bots/experts/deduplicator/expert.py line 41)


wagner-certat:
Thanks, also fixed in develop branch and in the [release notes of 1.0.1 RC1](https://github.com/certtools/intelmq/releases/tag/1.0.0.rc1).

Issue #1098, opened 2017-10-03
RT collector: Search for requestor


wagner-certat:
Add parameter `search_requestor` (default: none) adds a search for the requestors' email address.

wagner-certat:
The parameter is called `Requestor`

PR #1099, opened 2017-10-09
RT collector enhancements

wagner-certat:
fixes certtools/intelmq#1098


wagner-certat:
Abusix test fails because it seems to block/rate-limit travis

wagner-certat:

Issue #852, opened 2017-01-20
Collector-Mail-Url cannot mark E-Mails as sent, if the download is very slow


dmth:
When a download is very slow the E-Mail within the "Collector-Mail-Url" can not be marked as read.
In addition, the data is retrieved a second (third,...) time when the bot failed.

The collector needs to refresh the IMAP-Connection instead of runnning into a timeout.

```
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/intelmq/lib/bot.py", line 122, in start
    self.process()
  File
"/usr/lib/python3/dist-packages/intelmq/bots/collectors/mail/collector_mail_url.py",
line 93, in process
    mailbox.mark_seen(uid)
  File "/usr/lib/python3/dist-packages/imbox/__init__.py", line 54, in mark_seen
    self.connection.uid('STORE', uid, '+FLAGS', '(\\Seen)')
  File "/usr/lib/python3.4/imaplib.py", line 816, in uid
    typ, dat = self._simple_command(name, command, *args)
  File "/usr/lib/python3.4/imaplib.py", line 1133, in _simple_command
    return self._command_complete(name, self._command(name, *args))
  File "/usr/lib/python3.4/imaplib.py", line 958, in _command_complete
    raise self.abort('command: %s => %s' % (name, val))
imaplib.abort: command: UID => Disconnected for inactivity.
2017-01-02 17:50:25,545 - Shadowserver-Open-MDNS - INFO - Current
Message(event): None.
2017-01-02 17:50:25,546 - Shadowserver-Open-MDNS - INFO - Bot will restart
in 15 seconds.
2017-01-02 17:50:40,558 - Shadowserver-Open-MDNS - INFO - Bot woke up
2017-01-02 17:50:40,559 - Shadowserver-Open-MDNS - INFO - Trying to start
processing again.
2017-01-02 17:50:40,630 - Shadowserver-Open-MDNS - INFO - Reading email report
2017-01-02 17:50:40,630 - Shadowserver-Open-MDNS - INFO - Downloading report
from https://dl.shadowserver.org/REDACTED
2017-01-02 17:52:41,955 - Shadowserver-Open-MDNS - INFO - Report downloaded.
```

dmth:
BTW: This behaviour does **most likely** not happen once #854 is solved.

Nevertheless it **will** happen if the downloaded dataset is huge and does take a long time.

So we've got at least two options here:

[...]

PR #1100, opened 2017-10-09
Mail URL collector: reconnect for marking as seen if necessary

wagner-certat:
fixes certtools/intelmq#852


wagner-certat:

Issue #1051, opened 2017-08-03
code-beauty: intelmqctl __read_pidfile(self, bot_id) function can be simplified


aaronkaplan:
```python
    def __read_pidfile(self, bot_id):
        filename = self.PIDFILE.format(bot_id)
        if self.__check_pidfile(bot_id):
            with open(filename, 'r') as fp:
                pid = fp.read()
            return pid.strip()
        return None

    def __check_pidfile(self, bot_id):
        filename = self.PIDFILE.format(bot_id)
        if os.path.isfile(filename):
            try:
                with open(filename, 'r') as fp:
                    pid = fp.read()
                return int(pid.strip())
            except ValueError:
                return None
        return None
```

Can be simplified to something like:

```python
    def __read_pidfile(self, bot_id):
        filename = self.PIDFILE.format(bot_id)
        return self.__check_pidfile(bot_id)
```

PR #1102, opened 2017-10-09
ENH: bin/ctl: simpler pid file check

wagner-certat:
fixes certtools/intelmq#1051


wagner-certat:

PR #1105, opened 2017-10-11
Minor changes from certat repo

wagner-certat:
These changes haven't found their way to this repo.

I extended some of them to be actual bug fixes, like the regex-fix and the change in the shadowserver parser


wagner-certat:
I'd like to have a review for this PR, thanks.

wagner-certat:
So I'll merge it if nobody cares

PR #1108, opened 2017-10-17
BUG: fix saving extra.* as string for postgres

wagner-certat:
* fixes certools/intelmq#1107
* introduced a new parameter jsondict_as_string for the postgresql output bot with a useful default
* new parameter jsondict_as_string for `Message.to_dict`


wagner-certat:

Issue #1111, opened 2017-11-06
Spamhaus cert: lines with AS?


wagner-certat:
Lines with unknown AS look like this:
```203.0.113.04,AS?,AT,1509955710,locky,example.invalid,198.18.0.19,80,36288,tcp```
This is not handled currently:
```
intelmq.lib.exceptions.InvalidValue: invalid value '?' (<class 'str'>) for key 'source.asn'
```

Issue #1110, opened 2017-10-31
Reference GeoLite2 database disclaimer


wagner-certat:
https://dev.maxmind.com/geoip/geoip2/geolite2/#License

This needs to be placed wherever we reference to this database:

```
This product includes GeoLite2 data created by MaxMind, available from
<a href="http://www.maxmind.com">http://www.maxmind.com</a>.
```

Which is, the cron-job and probably the docs about the maxmind bot.

Issue #470, opened 2016-04-05
ENH: please let's make intelmq LSB compatible


aaronkaplan:
While /opt etc is all nice, once we move to debian packages (.deb) or RPMs, we should really be [LSB compatible](https://en.wikipedia.org/wiki/Linux_Standard_Base). 
This includes startup scripts (/etc/init.d/*... ).

sebix:
Does this mean we abandon `/opt/intelmq` entirely? AFAIR the decision of using this directory was to separate it from the rest of the system,

This also includes rejigging the logging entirely

aaronkaplan:
On Mon, Apr 11, 2016 at 02:59:20AM -0700, Sebastian wrote:
> Does this mean we abandon `/opt/intelmq` entirely? AFAIR the decision of using this directory was to separate it from the rest of the system,
> 
> This also includes rejigging the logging entirely

To be discussed. I think this was a discussion point.

aaronkaplan:
should be solved by the Debian packaging PR: #595

PR #1112, opened 2017-11-06
Release 1.0.2

wagner-certat:
### Bots
- `collectors.mail.collector_mail_url`: Fix bug which prevented marking emails seen due to disconnects from server (#852).
- `parsers.spamhaus.parser_cert`: Handle/ignore 'AS?' in feed (#1111)

### Packaging
- Support building for more distributions
- Use LSB-paths (/etc/intelmq/, /var/lib/intelmq/, /run/intelmq/) (#470)
- Use quilt instead of native
- Fix problems in postint and postrm scripts
- Use systemd-tmpfile for creation of /run/intelmq/

### Documentation
- Add disclaimer on maxmind database in bot documentation and code and the cron-job (#1110)


intelmq/bots/collectors/mail/collector_mail_url.py
```diff
@@ -90,7 +95,12 @@ def process(self):
                         # Only mark read if message relevant to this instance,
                         # so other instances watching this mailbox will still
                         # check it.
-                        mailbox.mark_seen(uid)
+                        try:
+                            mailbox.mark_seen(uid)
+                        except imaplib.abort:
+                            # Disconnect, see https://github.com/certtools/intelmq/issues/852
+                            mailbox = self.connect_mailbox()
```
aaronkaplan: what happens in an exception inside the except block?

intelmq/bots/parsers/spamhaus/parser_cert.py
```diff
@@ -47,7 +48,9 @@ def process(self):
             event.change("feed.url", event["feed.url"].split("key=")[0])
 
             event.add('source.ip', row_splitted[0])
-            event.add('source.asn', row_splitted[1].replace('AS', ''))
+            source_asn = row_splitted[1].replace('AS', '')
+            if source_asn != '?':
```
aaronkaplan: is '?' the only option which the field may have instead of a numeric value?
What about if shadowserver renames the value to '???'  ?

Lot's of question marks ;-)

contrib/cron-jobs/maxmind
```diff
@@ -17,6 +17,9 @@
 # 02  01 *   *   1     ( cd /tmp; /my/path/to/this/script )
 # 
 
+# This product includes GeoLite2 data created by MaxMind, available from
+# <a href="http://www.maxmind.com">http://www.maxmind.com</a>.
+
```
aaronkaplan: thx 👍

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ CHANGELOG
 
 ### Bots
 - `collectors.mail.collector_mail_url`: Fix bug which prevented marking emails seen due to disconnects from server (#852).
+- `parsers.spamhaus.parser_cert`: Handle/ignore 'AS?' in feed (#1111)
+
+### Packaging
+- Support building for more distributions
```
aaronkaplan: which ones? please name them.

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ CHANGELOG
...
+- Use LSB-paths (/etc/intelmq/, /var/lib/intelmq/, /run/intelmq/) (#470)
```
aaronkaplan: applies to all systems or just Debian? (It's LSB, so I would assume it's for all)

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ CHANGELOG
...
+- Use quilt instead of native
```
aaronkaplan: why?

NEWS.md
```diff
@@ -3,6 +3,10 @@ NEWS
 
 See the changelog for a full list of changes.
 
+1.0.2 Bugfix release
+--------------------
+No changes needed.
```
aaronkaplan: force= parameter in message lib?

CHANGELOG.md
wagner-certat: Because intelmq is not a native debian program. native does mean in the debian packagers context that it's only developed for debian. Will change it to "changed debian package format from native to quilt"

intelmq/bots/collectors/mail/collector_mail_url.py
wagner-certat: The generic error handling of the Bot class handles this, like before and everywhere else too.

intelmq/bots/parsers/spamhaus/parser_cert.py
wagner-certat: It's not shadowserver, it's spamhaus.

We received a report with 'AS?', so I handle that now. I don't know what they sent tomorrow, probably they use an empty field? So I adapted to code to what we know they sent in the past. I try not to guess too much about the format.

CHANGELOG.md
wagner-certat: For all who use the packages. For others, they use their methods (like pip etc), it does not apply.

probably like this?
```
### Packaging
- dep, rpm:
  - use LSB-paths...
```

NEWS.md
wagner-certat: This file is for users, not for developers.

intelmq/bots/collectors/mail/collector_mail_url.py
aaronkaplan: OK

intelmq/bots/parsers/spamhaus/parser_cert.py
aaronkaplan: we always will have to adapt.

Issue #1117, opened 2017-11-12
performance degradation for extra fields


navtej:
I have noticed severe performance degradation in bots caused by commit 50eb548f57444e87063b456903b38eb9bbb23ed8 to file `intelmq/lib/message.py`

Here are two runs with commit 50eb548f57444e87063b456903b38eb9bbb23ed8 and prior commit 6ccaf3b82b051749164b0884607b5c161b142056

Logging is set as follows
```
    "log_processed_messages_count": 10000,
    "log_processed_messages_seconds": 300,
```

First run is with commit 6ccaf3b82b051749164b0884607b5c161b142056 and the bot is run for _**5 minutes**_
```
ubuntu@ubuntu:~/src/intelmq$ git checkout 6ccaf3b82b051749164b0884607b5c161b142056 -- intelmq/lib/message.py
ubuntu@ubuntu:~/src/intelmq$ sudo -H pip3 install -e .
--pip output--
ubuntu@ubuntu:~/src/intelmq$ intelmqctl start taxonomy-expert
ubuntu@ubuntu:~/src/intelmq$ tail -f /opt/intelmq/var/log/taxonomy-expert.log
2017-11-12 13:29:32,022 - taxonomy-expert - INFO - TaxonomyExpertBot initialized with id taxonomy-expert and intelmq 1.1.0.alpha1 and python 3.4.2 (default, Oct  8 2014, 10:45:20) as process 19841.
2017-11-12 13:29:32,022 - taxonomy-expert - INFO - Bot is starting.
2017-11-12 13:29:32,023 - taxonomy-expert - INFO - Pipeline ready.
2017-11-12 13:30:01,141 - taxonomy-expert - INFO - Processed 10000 messages since last logging.
2017-11-12 13:30:30,353 - taxonomy-expert - INFO - Processed 10000 messages since last logging.
2017-11-12 13:30:59,151 - taxonomy-expert - INFO - Processed 10000 messages since last logging.
2017-11-12 13:31:28,417 - taxonomy-expert - INFO - Processed 10000 messages since last logging.
2017-11-12 13:31:56,735 - taxonomy-expert - INFO - Processed 10000 messages since last logging.
2017-11-12 13:32:25,360 - taxonomy-expert - INFO - Processed 10000 messages since last logging.
2017-11-12 13:32:54,706 - taxonomy-expert - INFO - Processed 10000 messages since last logging.
[...]

PR #1119, opened 2017-11-16
Performance fix

navtej:
Fixes the performance issues listed in bug #1117 . 

* Comment out `functools.lru_cache` decorator
* ~reorder the call to `_get_type_config` in `add`, and pass the result to `__is_valid_key` and `__is_valid_value` in params, so that these functions dont have to call `_get_type_config`~
This change will need lot of changes in tests, it has been undone.

### DIFF
--- a/intelmq/lib/message.py
+++ b/intelmq/lib/message.py
@@ -204,7 +204,11 @@ class Message(dict):
                 del self[key]
             return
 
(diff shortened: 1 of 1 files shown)


wagner-certat:
Thanks, that's a short quick fix until I have time to rearrange/rewrite the validation methods.

PR #1126, opened 2017-12-04
Debug logging of response text for restapi bot

Narzhan:
This PR ads the option to log response texts while the bot runs in debug mode. This option might be helpful while debugging the interaction between intelmq and intelmq "consumer". 
Description: If the logging level of bot/intelmq is set to debug and response is not success (HTTP 200), the bot tries to log the response text.

### DIFF
--- a/intelmq/bots/outputs/restapi/output.py
+++ b/intelmq/bots/outputs/restapi/output.py
@@ -34,7 +34,10 @@ class RestAPIOutputBot(Bot):
                               cert=self.ssl_client_cert,
                               timeout=self.http_timeout_sec,
                               **kwargs)
-        r.raise_for_status()
+        if self.parameters.logging_level in ["DEBUG", 10] and r.status_code != 200:
+            self.logger.debug("Something failed during message sending with response body {}".format(r.text))
+        else:
+            r.raise_for_status()
         self.logger.debug('Sent message.')
         self.acknowledge_message()
 
--- a/intelmq/bots/outputs/restapi/output.py
+++ b/intelmq/bots/outputs/restapi/output.py
@@ -34,13 +34,9 @@ class RestAPIOutputBot(Bot):
                               cert=self.ssl_client_cert,
                               timeout=self.http_timeout_sec,
                               **kwargs)
-        if self.parameters.logging_level in ["DEBUG", 10] and r.status_code != 200:
-            try:
-                self.logger.debug("Something failed during message sending with response body {}".format(r.text))
-            except Exception as e:
-                self.logger.debug("There was an error in your message, but there was no response text {}".format(e))
-        else:
-            r.raise_for_status()
+        if not r.ok:
+            self.logger.debug("Error during message sending with response body: %r.", r.text)
+        r.raise_for_status()
         self.logger.debug('Sent message.')
         self.acknowledge_message()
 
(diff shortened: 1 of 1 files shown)


intelmq/bots/outputs/restapi/output.py
```diff
@@ -34,7 +34,13 @@ def process(self):
                               cert=self.ssl_client_cert,
                               timeout=self.http_timeout_sec,
                               **kwargs)
-        r.raise_for_status()
+        if self.parameters.logging_level in ["DEBUG", 10] and r.status_code != 200:
```
wagner-certat: Logging level should be never 10, if it's not that's a bug.
And when using `logger.debug` the message is only logged if the level is debug or lower, so we don't need to check that.
And to check the response, better use `r.ok` (returns False if status code is between 400 and 600).

intelmq/bots/outputs/restapi/output.py
```diff
@@ -34,7 +34,13 @@ def process(self):
...
+            try:
+                self.logger.debug("Something failed during message sending with response body {}".format(r.text))
+            except Exception as e:
+                self.logger.debug("There was an error in your message, but there was no response text {}".format(e))
```
wagner-certat: Isn't that covered by `r.raise_for_status` already?

intelmq/bots/outputs/restapi/output.py
```diff
@@ -34,7 +34,13 @@ def process(self):
                               cert=self.ssl_client_cert,
                               timeout=self.http_timeout_sec,
                               **kwargs)
-        r.raise_for_status()
+        if self.parameters.logging_level in ["DEBUG", 10] and r.status_code != 200:
+            try:
+                self.logger.debug("Something failed during message sending with response body {}".format(r.text))
```
wagner-certat: Please let the logging library do the formatting (avoids bad error message when formatting fails), e.g.:
```python
self.logger.debug("Received response body: %r.", r.text)
```

intelmq/bots/outputs/restapi/output.py
Narzhan: I tried to access the variable logging_level and when not set directly for the bot it's content is the integer representation of given level (10 in this case). I will get rid of the level check and keep only the logging statement and r.ok check.

intelmq/bots/outputs/restapi/output.py
Narzhan: I've added this block as a precaution in case there is no text in response just to be sure. I haven't managed to find any info whether some text will always be present or not.

intelmq/bots/outputs/restapi/output.py
wagner-certat: > I tried to access the variable logging_level and when not set directly for the bot it's content is the integer representation of given level (10 in this case).

That's interesting, I will check that.

intelmq/bots/outputs/restapi/output.py
wagner-certat: > I've added this block as a precaution in case there is no text in response just to be sure.

Sure, but isn't that what's `raise_for_status` does? And the method also includes more information (error code and message).

> I haven't managed to find any info whether some text will always be present or not.

Is `r.text` then `None` or does it not exist at all ( = AttributeError)?

Narzhan:
I've added following changes:

- logging is handled by logging library

- r.ok is used as the only test of response validity

- Got rid of try catch, my Apiary test showed that even with no body in response there is no exception and the output is not None

wagner-certat:
Thanks! Much shorter now with the same effect :)

PR #1132, opened 2017-12-11
FIX: adapt cymru whois expert tests

SYNchroACK:
```
# whois -h whois.cymru.com '-v 2002:3ee0:3972:0001::1'
AS      | IP                                       | BGP Prefix          | CC | Registry | Allocated  | AS Name
6939    | 2002:3ee0:3972:0001::1                   | 2002::/16           |    | other    |            | HURRICANE - Hurricane Electric, Inc., US
```


wagner-certat:
It changed again - back :/
So I reverted it for maintenance and develop

Issue #1088, opened 2017-09-08
Named and addressable destination pipelines


wagner-certat:
with multiple destinations / destination groups it would be possible for the filter and sieve expert to push messages to one of the queues based on the message content

navtej:
+1 for this. One use could be an expert which can do complex arithmatic based load balancing, an upgrade to load_balance param.

wagner-certat:
can be used for the filter-expert and the sieve-expert

e3rd:
I've just finished second draft in branch.  
When we finish this [comment](https://github.com/certtools/intelmq/commit/b65a3154bf80aef74cae1524e498aa887089b2a1#r27223975) and make intelmq-manager ready, I may pull merge this into develop branch.

wagner-certat:
> One use could be an expert which can do complex arithmatic based load balancing, an upgrade to load_balance param.

For that I'd rather use better messaging queues such as RabbitMQ. I'm currently working on that to see how they name things.

wagner-certat:
> I may pull merge this into develop branch.

Please open a PR.

e3rd:
I've opened a [PR](https://github.com/certtools/intelmq/pull/1190). When merged, we should still have that issue open at least till:
* [ ] `intelmqctl run ... send` add path parameter.
* [ ] intelmq-manager is able to handle this

wagner-certat:
fixed by #1190

e3rd:
Are you sure to close this ticket? #1190 solved the most important part but I still had two tasks right [above](https://github.com/certtools/intelmq/issues/1088#issuecomment-366021269)

wagner-certat:
Hm, the actual problem is solved, concerning the manager -> issue there: https://github.com/certtools/intelmq-manager/issues/154 and for the ctl tool here: #1210

PR #1133, opened 2017-12-11
Bot option: Error Forward Message

SYNchroACK:
`error_forward_message` is an option for Expert bots. Bots like gethostbyname sometimes raise some errors which are only due some service problems like DNS nameserver for a specific domain is replying SERVFAIL. Therefore, with this new option, intelmq administrator can configure an expert bot to forward the original message to the next bot automatically after X retries configured as usual.


wagner-certat:
I changed the target branch to develop because this is a new feature. Have a look at https://github.com/certtools/intelmq/blob/master/docs/Developers-Guide.md#branching-model maintenance only get's bugfixes and master is always the latest release (nothing between stable versions)
You may want to rebase to develop now.

Please add this to the changelog (in the 1.1 core section).

wagner-certat:
@aaronkaplan I added you because I'd like if the name of the parameter is intuitive. I am not sure if it is.

The relevant change is in 89574ff

SYNchroACK:
> You may want to rebase to develop now.

Thank you @wagner-certat for the clarification. I tried to rebase but with no luck. Could you provide the commands to perform that operation successfully? Thank you.

wagner-certat:
As the branch you branched off was very old too (at the stage of the 1.0.0 RC), cherry-picking is faster:

First update your local branch:
```
git remote update
```
If you have a develop branch locally, assuming that upstream is the remote pointing to certtools (depends on your local configuration):
```
git checkout develop
git pull upstream develop
```
If you don't have develop locally yet:
```
git checkout -b develop upstream/develop
```

Copy the old branch, delete the existing and make it a copy of develop:
```
git branch backup error_forward_message
git branch -D error_forward_message
git checkout -b error_forward_message develop
```
And now pick the commit to the new branch:
```
git cherry-pick 89574ff
```
You will get a conflict in `intelmq/etc/defaults.conf`. Fix that and do
```
git add intelmq/etc/defaults.conf
git cherry-pick --continue
```
Now upload that branch:
```
git push -f origin error_forward_message
```
If everything went fine, you can delete the backup branch.

SYNchroACK:
> If everything went fine, you can delete the backup branch.

@wagner-certat Thank you! It works.

wagner-certat:
> @wagner-certat Thank you! It works.

Great, looks good.

SYNchroACK:
@wagner-certat the current travis error is regarding the `self.__group` which are not present in `tests.py`. Do you agree with the parameter `self.group`? Makes sense to have it loaded? If yes, can you help me fix the `tests.py`, I'm not sure the best way to adapt the tests.

Thank you in advance.

wagner-certat:
> @wagner-certat the current travis error is regarding the `self.__group` which are not present in `tests.py`. Do you agree with the parameter `self.__group`? Makes sense to have it loaded?

Currently it is a "parameter", i.e. user-defined. That does not make sense for the POV from intelmq itself, because it is defined by the bot type.

> If yes, can you help me fix the `tests.py`, I'm not sure the best way to adapt the tests.

`mocked_config` needs to take an additional parameter for the group. The value is in `self.bot_type`.

SYNchroACK:
@sebix check if everything is ok please. Thank you! :)

docs/User-Guide.md
```diff
@@ -109,6 +109,9 @@ You can set these parameters per bot as well. The settings will take effect afte
 * **`error_dump_message`** - specifies if the bot will write queued up messages to its dump file (use intelmqdump to re-insert the message).
     * **`true/false`** - write or not write message to the dump file
 
+* **`error_forward_message`** - specifies if the bot will forward messages to output queue which couldn't be processed by the bot. This option MUST be only used by Expert Bots.
```
aaronkaplan: non intuitive name.

how about: 
``event_always_passthrough`` .... pass through the event in any situation, even when errors exist in the expert.

intelmq/etc/defaults.conf
```diff
@@ -6,6 +6,7 @@
     "destination_pipeline_password": null,
     "destination_pipeline_port": 6379,
     "error_dump_message": true,
+    "error_forward_message": false,
```
aaronkaplan: see above.
We definitely need a comment here! This can have pretty bad effects if turned on.

intelmq/lib/bot.py
```diff
@@ -211,6 +217,10 @@ def start(self, starting: bool=True, error_on_pipeline: bool=True,
 
                         if error_on_message:
 
+                            if self.parameters.error_forward_message and self.group == "Expert":
+                                self.logger.info("Forwarding message to output queue.")
```
aaronkaplan: ``self.logger.info ("Forcing forwarding of message even in case of errors")``

wagner-certat:
This could be a use case for https://github.com/certtools/intelmq/issues/1088
The name for this special destination queue could be e.g. `_on_error`

SYNchroACK:
> The name for this special destination queue could be e.g. _on_error

My suggestion would be:

```
"cymru-whois-expert": {
	"source-queue": "cymru-whois-expert-queue",
	"destination-queue-on-error": "error-queue",
	"destination-queues": [
		"file-output-queue"
	]
}
```

Issue #1063, opened 2017-08-09
pymongo 3.0 deprecates used insert method


wagner-certat:
Excerpt from [changelog](https://github.com/mongodb/mongo-python-driver/blob/0b0f045d56176807edea9e547ab6a1288740a5d8/doc/changelog.rst#classpymongocollectioncollection-changes) of pymongo version 3.0:

> The following methods are deprecated:
    ...
    :meth:`~pymongo.collection.Collection.insert`

We are using exactly this method to insert data into mongodb databases. `insert_one` should be used now (introduced in version 3.0). There is no date/version given for the method removal, however we should be prepared and be future-compatible before things break somewhere.

FYI @vince-nayal and @CNCS-PT (authors of the bot)

Issue #1062, opened 2017-08-09
pymongo >= 3.5: authentication changes


wagner-certat:
Excerpt from [changelog](https://github.com/mongodb/mongo-python-driver/blob/0b0f045d56176807edea9e547ab6a1288740a5d8/doc/changelog.rst#changes-in-version-35) of newly released pymongo version 3.5:

> - Username and password can be passed to
>   :class:`~pymongo.mongo_client.MongoClient` as keyword arguments. Before, the
>   only way to pass them was in the URI. 

And: 

> Deprecated :meth:`~pymongo.database.Database.authenticate`. Authenticating multiple users conflicts with support for logical sessions in MongoDB 3.6. To authenticate as multiple users, create multiple instances of :class:`~pymongo.mongo_client.MongoClient`.

We are currently using this method for authentication. I suggest to use the new authentication for newer versions of pymongo and use the newer for anything >= 3.5. with pymongo 3.6 this needs to be done, otherwise the bot will be broken.

FYI @vince-nayal and @CNCS-PT (authors of the bot)

wagner-certat:
And further in [version 3.6](https://github.com/mongodb/mongo-python-driver/blob/master/doc/changelog.rst#changes-in-version-360):

> Unavoidable breaking changes:

>    Starting in MongoDB 3.6, the deprecated methods :meth:`~pymongo.database.Database.authenticate` and :meth:`~pymongo.database.Database.logout` now invalidate all cursors created prior. Instead of using these methods to change credentials, pass credentials for one user to the :class:`~pymongo.mongo_client.MongoClient` at construction time, and either grant access to several databases to one user account, or use a distinct client object for each user.

I am not sure if that has an impact on how we use these methods.

vince-nayal:
Hi,

Sorry, i missed the first notification.

Since I never had to deal with such situation, I am not sure how to do this, do we have to break backward compatibility or is there a way to preserve both methods ?

wagner-certat:
@vince-nayal 
Using `pymongo.version` you can differentiate between the versions. I.e. do `pymongo.version.split('.') < ['3', '5']`

wagner-certat:
I fixed this now based on the documentation. If you have time, maybe you can test it?

PR #1134, opened 2017-12-11
mail attach update

e3rd:
dependency library imbox changed
https://github.com/martinrusev/imbox/commit/7c6cc2fb5f7e39c1496d68f3d432eec19517bf8e#diff-1ae09572064c2e7c225de54ad5b49154
it propageted to pypi last week


e3rd:
Merged ✓ :)

You thing you overlooked it? How could you notice such a change in an external library? :o I've discovered it be mere accident.

wagner-certat:
> You thing you overlooked it? How could you notice such a change in an external library?

Because I usually read the changelogs of our dependencies. So I did find for example #1062 and #1063

PR #1136, opened 2017-12-13
CleanMX Parser bot (XML format)

SYNchroACK:
Current CleanMX Parser is not dealing with bad data from the CleanMX feed on CSV format. Since there is a XML format available with the same data, I suggest to use it instead of csv format. During the development of this parser I found out that http collector bot sometimes hangs on this [line](https://github.com/certtools/intelmq/blob/develop/intelmq/bots/collectors/http/collector_http.py#L66) because `.text` tries to discover the best encoding (when is not provided) and in case the report is big (like CleanMX reports) the bot stops on that line and never come back.

from http://docs.python-requests.org/en/master/api/#encodings

> _If Response.encoding is None, encoding will be guessed using chardet._
>
> _The encoding of the response content is determined based solely on HTTP headers, following RFC 2616 to the letter. If you can take advantage of non-HTTP knowledge to make a better guess at the encoding, you should set r.encoding appropriately before accessing this property._


intelmq/bots/parsers/cleanmx/parser.py
```diff
@@ -50,85 +54,101 @@
...
+                if key == "source.fqdn" and event.is_valid('source.ip', value):
+                    continue
+
+                if key == "time.source":
+                    value = value + " UTC"
+
+                if key == "source.asn":
+                    if value.upper().startswith("ASNA"):
+                        continue
+                    for asn in value.upper().split(','):
```
wagner-certat: Can there be multiple ASNs?

SYNchroACK:
Sebastian suggestion:
> for #1136 cleanmx: can you issue a warning in the parser if an old URL in feed.url is detected? That would help the users, point them to the NEWS-file where you explain the issue.

wagner-certat:
from the IRC synchroack:

> I doubt that someone has CleanMX bot running without errors... the CSV file format from the feed is really broken... I tried to fix using the current approach but its not worth it... the best approach is to use XML file format... so I consider this as a Fix.

-> a bug fix for maintenance

intelmq/bots/parsers/cleanmx/parser.py
SYNchroACK: yes, thats why Im picking one.

Issue #1082, opened 2017-08-23
check for orphaned queues


wagner-certat:
`intelmqctl check` could show "orphaned" existing queues which do not belong to any existing (configured) bot.

wagner-certat:
that also requires #1059

Issue #1059, opened 2017-08-07
intelmqctl check reachable redis


wagner-certat:
intelmqctl can check if the redis configuration is valid and working by connecting to it

PR #1139, opened 2017-12-18
ENH: ctl: check redis and orphan queues

wagner-certat:
fixes certtools/intelmq#1059
fixes certtools/intelmq#1082

checks if redis connection succeeds
and if there are more queues in redis than in the pipeline configuration


wagner-certat:

PR #1140, opened 2017-12-18
small run_mode fixes and checks

wagner-certat:
the default run_mode is called continuous according to docs, in lib/bot it was stream
and add a check for this parameter in ctl check


wagner-certat:

Issue #1104, opened 2017-10-10
Rules for harmonization keys not enforced and not sufficient


wagner-certat:
The current [data harmonization docs say](https://github.com/certtools/intelmq/blob/9fb52b38d6b56a099c1b6de0a0b4d663d386a243/docs/Data-Harmonization.md#rules-for-keys) that any key must match the regular expression `[a-z_.]`. Actually this is not enough, as `extra.` shouldn't be allowed just as `extra..foo`.

Currently both keys are allowed too, as it only checks if they exists in the harmonization configuration and, if the type is a JSONDict anything behind the dot is not checked.

wagner-certat:
The regular expression `^[a-z_](.[a-z_]+)*$` could work:
```python
>>> re.match('^[a-z_](.[a-z_]+)*$', 'extra.foo.asd')
<_sre.SRE_Match object; span=(0, 13), match='extra.foo.asd'>
>>> re.match('^[a-z_](.[a-z_]+)*$', 'extra.foo.')
>>> re.match('^[a-z_](.[a-z_]+)*$', 'extra..')
```

wagner-certat:
Any thoughts @aaronkaplan?

aaronkaplan:
I'd even be willing to relax the RE to  ``^[a-z_](.[a-z_0-9]+)*$``

wagner-certat:
Sure

PR #1141, opened 2017-12-19
BUG: enforce rule on harmonization key names

wagner-certat:
fixes certtools/intelmq#1104


wagner-certat:

Issue #1056, opened 2017-08-07
duplicate cron job scripts


wagner-certat:
There is [intelmq/bots/experts/maxmind_geoip/update-geoip-data](https://github.com/certtools/intelmq/blob/develop/intelmq/bots/experts/maxmind_geoip/update-geoip-data) by @swilde and [contrib/cron-jobs/maxmind](https://github.com/certtools/intelmq/blob/develop/contrib/cron-jobs/maxmind) written by @aaronkaplan.

I'd like to maintain just one of them.

wagner-certat:
The same applies to tor nodes and asn lookup bots / scripts

PR #1142, opened 2017-12-20
README and reload hint for cron-jobs, cleanup

wagner-certat:
fixes certtools/intelmq#1056
removes duplicate code
nothing changes for packages


wagner-certat:

PR #1143, opened 2017-12-20
BUG: ctl: proper exit codes

### LINKED ISSUE #977: Proper exit codes for intelmqctl
intelmqctl should return non zero status when an action did not succeed (start/stop).
The same for status, if (one of) the bot(s) is stopped. That's analogous to other tools and useful for monitoring.

wagner-certat:
intelmqctl now exits with exit codes > 0 when errors happened or the operation was not successful. Also, the status operation exits with 1, if bots are stopped, but enabled. (fixes #977)


wagner-certat:

Issue #1078, opened 2017-08-17
Mail collectors do not have debug output


wagner-certat:
In DEBUG they do not log very much what they are doing (connecting, searching, iterating etc).

e3rd:
I tried to add some lines in #1151 . Is that alright or should I add / remove some debug lines or do you prefer to manage up yourself (in that case just decline the PR please)?

wagner-certat:
Ideas from #1621:

Log the the mail's subject and date.

In case of errors, log information also as ERROR.
Also in case of URL fetching errors, see #1621.

PR #1151, opened 2018-01-10
mail collectors: added debug information #1078

e3rd:
It seems to me these log lines will suit to most test cases.

### DIFF
--- a/intelmq/bots/collectors/mail/collector_mail_attach.py
+++ b/intelmq/bots/collectors/mail/collector_mail_attach.py
@@ -38,7 +38,7 @@ class MailAttachCollectorBot(CollectorBot):
                 if (self.parameters.subject_regex and
                         not re.search(self.parameters.subject_regex,
                                       re.sub("\r\n\s", " ", message.subject))):
-                    self.logger.debug("Message %s skipped.", message.subject)
+                    self.logger.debug("Message with date %s skipped because subject %r does not match.", message.date, message.subject)
                     continue
 
                 for attach in message.attachments:
--- a/intelmq/bots/collectors/mail/collector_mail_attach.py
+++ b/intelmq/bots/collectors/mail/collector_mail_attach.py
@@ -23,6 +23,7 @@ class MailAttachCollectorBot(CollectorBot):
             self.stop()
 
     def process(self):
+        self.logger.debug("Connecting to %s", self.parameters.mail_host)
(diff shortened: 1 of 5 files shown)


e3rd:
Thanks for the advices! Any other thing I can do here?

> dot

I did an regex search and added puctuation to multiple places in the project.

> parameters for string formatting are better passed as argument to the log function

What does that mean please? I'm already passing the parameters as arguments. See, subject is passed as an argument. `self.logger.debug("Message %s skipped.", message.subject)` 

> Another possible debug output:

I did rather "no unred mails to check" because there may be other parameters than read/unread (also sent-to/from).

intelmq/bots/collectors/mail/collector_mail_attach.py
```diff
@@ -37,6 +38,7 @@ def process(self):
                 if (self.parameters.subject_regex and
                         not re.search(self.parameters.subject_regex,
                                       re.sub("\r\n\s", " ", message.subject))):
+                    self.logger.debug("Message %s skipped", message.subject)
```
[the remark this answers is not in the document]
e3rd: implemented

intelmq/bots/collectors/mail/collector_mail_url.py
```diff
@@ -46,9 +47,10 @@ def process(self):
                 if (self.parameters.subject_regex and
                         not re.search(self.parameters.subject_regex,
                                       re.sub("\r\n\s", " ", message.subject))):
+                    self.logger.debug("Message with date %s skipped because subject %r does not match.", message.subject)
```
wagner-certat: There's a parameter missing

wagner-certat:
> > parameters for string formatting are better passed as argument to the log function

> What does that mean please?

-.- Sorry, don't know where I looked at.

> I did an regex search and added puctuation to multiple places in the project.

Thanks!

> I did rather "no unred mails to check" because there may be other parameters than read/unread (also sent-to/from).

Apart from the one missing parameter, ready to merge. I think it's sufficient for most purposes.

intelmq/bots/collectors/mail/collector_mail_url.py
e3rd: added here and in the other file, too

e3rd:
sorry, my bad – added ✓

wagner-certat:
merged and fixed the code style too (travis/pycodestyle complained about 2 too long lines)

PR #1153, opened 2018-01-11
BUG: fix smtp output starttls

wagner-certat:
fixes certtools/intelmq#1152


wagner-certat:
Also pushed to develop

PR #1154, opened 2018-01-11
Custom Filter Expert Bot

e3rd:
We're successfully using this bot for months. Custom-filter is more complicated than standard intelmq filter, well documented in README and in test examples. If the community's willing to have it in the main repository, I'll add notion to CHANGELOG.md, BOTS.md and in bots/BOTS file.

However, if there is a reason the bot should be declined, do not hesitate.

### DIFF
--- a/intelmq/bots/experts/custom_filter/README.md
+++ b/intelmq/bots/experts/custom_filter/README.md
@@ -1,0 +1,63 @@
+# Custom filter bot
+
+A bot to filter specific events from the pipeline. See valid fields at https://github.com/certtools/intelmq/blob/master/docs/Harmonization-fields.md
+
+With the filter bot, one should be able to define filters such as:
+ - throw away all events for specific ASN or CIDR
+ - pass just events with particular CC or Taxonomy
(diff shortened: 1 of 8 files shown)


intelmq/bots/experts/custom_filter/README.md
```diff
@@ -0,0 +1,63 @@
+# Custom filter bot
+
+A bot to filter specific events from the pipeline. See valid fields at https://github.com/certtools/intelmq/blob/master/docs/Harmonization-fields.md
```
wagner-certat: Actually the valid fields are defined in etc/

wagner-certat:
How do we want to continue? You said you didn't know the sieve expert. Do you plan to use and maintain the bot, or are you switching to the sieve expert? If the latter, then I'd not merge this as there's no maintenance then.

e3rd:
You are right, sorry for letting the pull request open for so long without a valid reason. We are half-switched and as we discussed I don't think neither it is good idea to have a 3 bots with the same functionality while there could be just 2 of them.

PR #1156, opened 2018-01-12
ENH: Fix issues found by lgtm tool

wagner-certat:
Most notably: Implemented `Message.__eq__` for proper comparisons

See: https://lgtm.com/projects/g/certtools/intelmq
Many of them are false positives, but some hints are very good :)

### DIFF
--- a/intelmq/lib/message.py
+++ b/intelmq/lib/message.py
@@ -116,6 +116,8 @@ class Message(dict):
             iterable = message.items()
         elif isinstance(message, tuple):
             iterable = message
+        else:
+            raise ValueError("Type %r of message can't be handled, must be dict or tuple.", type(message))
         for key, value in iterable:
             if not self.add(key, value, sanitize=False, raise_failure=False):
                 self.add(key, value, sanitize=True)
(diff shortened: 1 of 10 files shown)


intelmq/lib/message.py
```diff
@@ -443,6 +445,25 @@ def to_json(self, hierarchical=False, with_type=False, jsondict_as_string=False)
...
+            type_eq = type(self) == type(other)
+            harm_eq = self.harmonization_config == other.harmonization_config if hasattr(other, 'harmonization_config') else False
+            if type_eq and dict_eq and harm_eq:
+                return True
+        elif dict_eq:
+            return True
+
+    def __ne__(self, other) -> bool:
+        return not self.__eq__(other)
+
```
SYNchroACK: I think its missing a final `return False`.

intelmq/lib/message.py
```diff
@@ -443,6 +445,25 @@ def to_json(self, hierarchical=False, with_type=False, jsondict_as_string=False)
...
+        """
+        dict_eq = super(Message, self).__eq__(other)
+        if issubclass(type(other), Message):
+            type_eq = type(self) == type(other)
+            harm_eq = self.harmonization_config == other.harmonization_config if hasattr(other, 'harmonization_config') else False
+            if type_eq and dict_eq and harm_eq:
+                return True
+        elif dict_eq:
+            return True
+
```
SYNchroACK: I think its missing a final `return False`.

wagner-certat: Implicitly it's None here.
According to the docs, if the return value is not boolean, bool() is called.

Pushed an additional commit

SYNchroACK: Yes, its true but I tested here and it gaves me `None`, not `False`. May be I did something wrong.

```
>>> def return_test(other) -> bool:
...   if other == "a":
...     return True
... 
>>> print( return_test("a") )
True
>>> print( return_test("b") )
None
```

PR #1161, opened 2018-01-18
Extra log line during dumping message process

SYNchroACK:
The only "small" reason for the additional line is just to help users on the same problem as I had which is:

A parser bot is processing thousands of lines and in the middle of the process, parser found one line with an error and automatically wrote a log message about it but continued. In the end, after all thousand lines processed, the log file is full of messages saying `INFO - Processed 500 messages since last logging.` and in the end has an additional log message saying `INFO - Dumping message from pipeline to dump file.`. 

In my perspective, before the log line `INFO - Dumping message from pipeline to dump file.`, bot should write another log line mentioning the issue with the specific message that is about to dump in order to alert the user that need to check all log file looking for that error. Otherwise the user may have difficulties to understand where the dump message came from after all processed lines.

### DIFF
--- a/intelmq/lib/bot.py
+++ b/intelmq/lib/bot.py
@@ -604,7 +604,10 @@ class ParserBot(Bot):
             self.acknowledge_message()
             return
 
-        for line in self.parse(report):
+        lines = self.parse(report)
+
+        for line in lines:
+
             if not line:
                 continue
             try:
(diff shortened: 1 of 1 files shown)


wagner-certat:
The bug you're trying to fix is specific to the ParserBot, so only fix it there, [in `process`](https://github.com/certtools/intelmq/blob/develop/intelmq/lib/bot.py#L667). And to make it more useful we could also give the number of errors in the message (length of `self.__failed`).

e.g.:
> ERROR ...
> Sending message ...
> ... - INFO - %d errors occurred during the processing of the report. See error messages above.
> Dumping ...

wagner-certat:
How do you know that each line represents one event? This is often not the case (think about headers).

wagner-certat:
FYI: 6c3d66a0a2f065ead0012e631e6b0dc731079b19

PR #1167, opened 2018-01-20
Feeds configurations generator tool

SYNchroACK:


SYNchroACK:
> Please add an entry to the changelog
Where should I put the entry on changelog? Which section?!

> Please add documentation on how to use this for single bots
done

PR #1168, opened 2018-01-22
DOCS: Adds travis config hint.

bernhardreiter:
Travis default clone depth is 50 and `VERSION=$(git describe --abbrev=0 --tags)` used currently will fail
if there is no tag in the last 50 commits.

https://docs.travis-ci.com/user/customizing-the-build/#Git-Clone-Depth


wagner-certat:
I could make this optional too as most forks won't have these tags.

wagner-certat:
Obsolete, got rid of the version detection with tags, see #1171

Issue #536, opened 2016-06-01
RFD remove, integrate or replace termstyle package


bernhardreiter:
Right now it seems that intelmq depends on https://pypi.python.org/pypi/termstyle/0.1.10
which is considered "dormant" from the author.

Because terminal styling may not be the most important task, I think we may ease our packaging
by going to a  more stable situation here. Either:
 * remove termstyle entirely (did so for https://github.com/Intevation/intelmq-mailgen/issues/11 )
 * integrate and ship it as part of intelmq (thus be prepared to maintain it ourselfs or occasionally sync from https://github.com/timbertson/termstyle/
 * switch to a different library. More popular and actively maintained seem to be, for example
 ** https://pypi.python.org/pypi/termcolor
 ** https://pypi.python.org/pypi/blessings

sebix:
>  which is considered "dormant" from the author.

All the module does is inserting ANSI escape codes, there's not much to maintain.

Packaging is not hard, it's just a plain python package. Did it for opensuse recently, no pain at all.

But as it's not used in your case anyway, you may want to first evaluate what you need. Then decide what to use.

bernhardreiter:
This is a small cleanup issue, of course. 
I know that it is simple and small, but it is also somehow unmaintained. There is a small extra
hurdle that may bite us later (in a few years of maintenance) and now if someone wants to get
started soon.

Yesterday I ran into this, I wanted to start intelmq-mailgen (which then still used termstyle)
on Opensuse 13.1. Now termstyle packaged there. Once I've looked at the file, I knew that packaging is not the main issue, I could have just placed there, however I would have had to think which is suboptimal, if I need to do this for many small packages it adds up fast.

sebix:
> Yesterday I ran into this, I wanted to start intelmq-mailgen (which then still used termstyle)
on Opensuse 13.1. No termstyle packaged there. Once I've looked at the file, I knew that packaging is not the main issue, I could have just placed the file there, however I would have had to think which is suboptimal. If a person that want to install intelmq needs to do this for many small packages it adds up fast.

Working on it. I plan to push it to other suse's also. It's already in tumbleweed and leap 42.2.

bernhardreiter:
@sebix: If you think that termstyle (which only works on some platforms I assume) is an improvement to the Free Software world (over all the other python packages for terminal colors out there).... it is very cool that you are stepping up at its maintainer. :)

[...]

PR #1169, opened 2018-01-22
Improves travis setup.

bernhardreiter:
* Puts popd on a line of its own, otherwise its return value will
   override the exit code of the dpkg-buildpackage command, causing a
   successful build, even with it failed.


wagner-certat:
I noticed this too a few days ago.

We probably need to deactivate the build requirement check.

wagner-certat:
Changed the branch to maintenance (next upcoming version). The failing unit test is fixed there.

bernhardreiter:
> We probably need to deactivate the build requirement check.

I'm not sure what you mean here.

(If debian packages are to be build by the CI, of course success should be checked. )

wagner-certat:
Trusty (used by travis) has all build dependencies except for redis and termstyle. (See also #536)

So we can't fulfill all the requirements without adding my own repositories.

> (If debian packages are to be build by the CI, of course success should be checked. )

But they are not distributed, so we only check for "syntactic" correctness with this build.

Will merge and push additional fixes tomorrow.

wagner-certat:
We can't run the debhelpers' tests because of travis-ci/travis-ci#2231 (see log https://travis-ci.org/wagner-certat/intelmq/jobs/332231121#L779 )

PR #1178, opened 2018-02-01
BUG: lib/bot: enhance pipeline error handling

wagner-certat:
don't initialize source pipeline if not necessary
always raise PipelineError in case of connection issues
always stop in these cases, do never pass, would only result in loops


aaronkaplan:
looks good & logical, I think we can merge.

PR #1180, opened 2018-02-01
parser for openphish commercial feed

fpokorny:


wagner-certat:
I also don't understand why this change is necessary or useful, please be more verbose!

https://www.openphish.com/feed.txt (see https://github.com/certtools/intelmq/blob/master/docs/Feeds.md#openphish) has one URL per line, no JSON there.

fpokorny:
Ye sorry, it's my bad. The free openphish feed is just a textfile, however they offer free access for CSIRTs to their commercial feed which is a file with one json per line. So for the sake of intelmq I suppose it should be a "new" feed, something like openphish-commercial..

wagner-certat:
Ok, so it's not a bugfix, but a new "feature"/feed.

Then please create a new entry in etc/feeds.yml describing the Feed, write the new parser+tests etc.

fpokorny:
Okay, now it should be better. :-)

intelmq/etc/feeds.yaml
```diff
@@ -105,6 +105,25 @@ providers:
...
+    Phishing:
+      description: OpenPhish is a fully automated self-contained platform for phishing
+        intelligence. It identifies phishing sites and performs intelligence analysis
+        in real time without human intervention and without using any external resources,
+        such as blacklists.
+      additional_information:
+      bots:
+        collector:
+          module: intelmq.bots.collectors.http.collector_http
+          parameters:
```
wagner-certat: The username/password parameters are missing here.

fpokorny:
Should be fixed now.

wagner-certat:
FYI, some small adaptions I made additionally: 91b0cca8

PR #1182, opened 2018-02-05
Malwareurl parser

dargen3:
There is new malwareurl parser, which parse data from website https://www.malwareurl.com/.


wagner-certat:
Thanks, merged!

Also added feaa132cb033aeb79835c1aea724d0813a0f24b8

PR #1186, opened 2018-02-06
threatminer parser, test, and modified CHANGELOG.md and Feeds.md

dargen3:
Threatminer parser parse data from http collector, which take data from https://www.threatminer.org/ .


wagner-certat:
Thanks again!

Issue #123, opened 2014-08-24
API


SYNchroACK:
Suggestions for API:
- Tornado

```
Basic Example Tornado API: https://github.com/jonycse/tornado_rest_api/blob/master/api.py
SSL Tornado: http://stackoverflow.com/questions/20636145/tornado-ssl-certs
```

Any other suggestions?

PR #1187, opened 2018-02-09
Bot for collecting data using API

tavipoldma:
initial version for issue #123


wagner-certat:
Any chance to get progress here? Especially *docs*.

cc @haam3r

tavipoldma:
excuses for delay / winter sleep. start and stop issue is now resolved, please check
@wagner-certat

intelmq/bots/collectors/api/REQUIREMENTS.txt
```diff
@@ -1 +1 @@
-tornado>=4.5.2
+tornado>=5.1
```
wagner-certat: Is this because you only tested it with this version or do the changes require the new version?

tavipoldma: both, changes are using the latest tornado syntax and only tested for that version

wagner-certat: Interesting, I successfully tested it with 4.5.3

PR #1189, opened 2018-02-14
Twitter collector and parser

hugeox:
Twitter collector: collects texts of tweets from given users. Further, if there is a link to pastebin, the text from it can also be collected and sent in a separate report.

Twitter parser: allows extraction of urls in a fuzzy format from text.


robcza:
@aaronkaplan I'd like to draw your attention to this PR as well. This twitter collector/parser is oriented on fqdn only at the moment, but I hope it won't be a blocker for starting the Twitter adventure in intelmq.
Could you go briefly through the code/comments and post your thoughts?

intelmq/bots/parsers/twitter/parser.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def get_domain(self, address):
+        try:
+            dom = re.search('(/|^)([a-z0-9.-]+\.[a-z0-9]+?)([/:]|$)', address).group(2)
+            if not self.in_whitelist(dom):
+                if get_tld(url_normalize(dom), fail_silently=True):
+                    return url_normalize(dom)
+                else:
+                    return None
+        except:
```
wagner-certat: do not use bare excepts, use at least `except Exception` if you do not know the exact exception you are expecting here (I think you want `TypeError`)

intelmq/bots/parsers/twitter/parser.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def process(self):
+        report = self.receive_message()
+        data = []
+        text = utils.base64_decode(report["raw"])
+        classification_type = self.get_classification_type(text)
+        data.extend(self.get_data_from_text(text))
+        for url in data:
+            event = self.new_event(report)
+            event.add('raw', url, overwrite=True)
```
wagner-certat: * there shouldn't be the need to overwrite
* raw means raw and unparsed. here parsed and unparsed is the same.

intelmq/bots/collectors/twitter/collector_twitter.py
```diff
@@ -0,0 +1,119 @@
...
+        access_token_key
+        access_token_secret
+
+To get api login data see: https://python-twitter.readthedocs.io/en/latest/getting_started.html
+
+"""
+
+
+import time
+import twitter
```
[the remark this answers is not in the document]
hugeox: Can you please explain why would one not want to raise ImportError? 
The tests failed because requirements are not installed. What can be done about that?

intelmq/bots/parsers/twitter/parser.py
hugeox: So should the raw be the thing that was sent from the collector?

intelmq/bots/collectors/twitter/collector_twitter.py
```diff
@@ -0,0 +1,119 @@
...
+                            access_token_secret=self.parameters.access_token_secret,
+                            tweet_mode="extended")
+
+    def get_text_from_url(self, url: str) -> str:
+        supported_sources = ["pastebin.com"]
+        for source in supported_sources:
+            if source in url:
+                if "raw" not in url:
+                    request = requests.get(
+                            url.replace("pastebin.com", "pastebin.com/raw"))
```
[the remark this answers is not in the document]
hugeox: In response to above. I remove that because what we care about is only the domain. However, I appreciate that it doesn't make sense to unnecessarily get rid of the additional information in the url. 
Ad foo.bar, I will change / to // in the regex.

intelmq/bots/collectors/twitter/collector_twitter.py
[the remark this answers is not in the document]
wagner-certat: The problem is that if we do that, the user is completely unaware why the bot does die without writing any logs.

intelmq/bots/parsers/twitter/parser.py
wagner-certat: Or the part of the previous raw that has been parsed. E.g. in csv's it's the header and the line.

hugeox:
I am sorry about the number of commits, but unfortunately I merged from develop in the middle and now commits of other people get mixed up, making it really hard to rebase. I have squashed the commits that I could, but I don't know how to do the ones before the merge.

wagner-certat:
> I am sorry about the number of commits, [...]

Thanks, already better :)

hugeox:
It seems like it wasn't sorted in the first place (AnubisNetwork is in the wrong place)

wagner-certat:
Fixed the remaining formatting problems in c9df3aac

Issue #252, opened 2015-07-07
Harmonization: confidence, TLP, URL/URI, regex checks


SYNchroACK:
- [x] add `confidence_level` key to Event object (Harmonization)
- [x] add 'TLP' key to Event object (Harmonization)
- [ ] rename `source.url` to `source.uri` (https://danielmiessler.com/study/url_vs_uri/)
- [x] Add regex (harmonization.conf parameters) check in harmonization.py

sebix:
- `confidence_level` done by #381
- regex checks are implemented by #324 and #304

Any comments about URL <-> URI and adding a TLP key? @aaronkaplan @robcza @mauroasilva

robcza:
- URL <-> URI seems kind of complicated to me, it is not just about changing the name, the data type would have to change a lot to comply with full URI. And I'm happy with URL. I don't see any need for URI in IntelMQ use cases I work on, that does not mean others don't see them
- TLP would be useful. I would like to have it on the Report level and from there inherited to the Event level

sebix:
Further comments by @aaronkaplan @mauroasilva to this issue? It's tagged with milestone 1.0, so it should be done soon.

aaronkaplan:
TLP: what does it refer to ? which info gets TLP protected? The whole event? Individual parts?
Does the receiver understand TLP? It's not so easy...
Please, let's keep source.url.

aaronkaplan:
See also FIRST.org's [new TLP definiton](https://www.first.org/global/sigs/tlp). It's not so trivial. I'd like to move this to version 1.1

wagner-certat:
Now we are working on 1.1 and we still need a TLP field.

aaronkaplan:
okay, so... if we define TLP for the whole event (log line), then I am fine with it. But this needs to be in the documentation

wagner-certat:
TLP is also done in the meantime

wagner-certat:
As the conses here is to keep the URL I'm closing.

PR #1194, opened 2018-02-28
TLP level

ghost:
See certtools/intelmq#252

Only valid values: WHITE, GREEN, AMBER, RED.

Accepted for sanitation are different cases and the prefix 'tlp:'.


wagner-certat:
Could you please have a look at the changes I made?

PR #1195, opened 2018-02-28
Fixes for shadowserver/config.py

th-certbund:
Add missing dash in 'opensmb'.
Rename identifier 'accessible-vnc' to 'open-vnc' like with rdp etc.
Clean up some comments.


th-certbund:
Done.

PR #1196, opened 2018-03-02
Add taxonomy to abuse.ch parsers

th-certbund:


wagner-certat:
Thanks, I fixed the newline of the test in db1ab0fd

PR #1200, opened 2018-03-05
Shadowserver config.py netbios, poodle, freak

th-certbund:
Rename identifiers:
Make SSL-POODLE/SSL-FREAK lowercase like all other identifiers.
Rename open-netbios to open-netbios-nameservice as this report only deals with 137/udp, not netbios in general.


wagner-certat:
Thanks again!

PR #1202, opened 2018-03-05
Sieve strictness (allow ip networks with hosts bits set)

e3rd:
I found it handy to have a parameter for setting strict=False when creating an IP network. If strict=True, we get ValueError: XY has host bits set.

### DIFF
--- a/intelmq/bots/experts/sieve/README.md
+++ b/intelmq/bots/experts/sieve/README.md
@@ -46,6 +46,7 @@ if classification.type == ['phishing', 'malware'] && source.fqdn =~ '.*\.(ch|li)
 
 The sieve bot takes only one parameter:
  * `file` - filesystem path of the sieve file
+ * `ip_network_strict` (default true) - throw error if the network has host bits set
 
 
 ## Reference
--- a/intelmq/bots/experts/sieve/README.md
+++ b/intelmq/bots/experts/sieve/README.md
@@ -46,7 +46,6 @@ if classification.type == ['phishing', 'malware'] && source.fqdn =~ '.*\.(ch|li)
 
 The sieve bot takes only one parameter:
  * `file` - filesystem path of the sieve file
- * `ip_network_strict` (default true) - throw error if the network has host bits set
 
 
 ## Reference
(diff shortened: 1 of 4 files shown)


wagner-certat:
I'm not sure what you want to achieve because intelmq rejects networks which have the host bit set in the harmonization layer:

```
>>> import intelmq.lib.harmonization as harm
>>> harm.IPNetwork.is_valid('10.0.0.1/8')
False
>>> harm.IPNetwork.is_valid('10.0.0.0/8')
True
```
And for sanitation the host bits are removed:
```
>>> harm.IPNetwork.is_valid('10.0.0.1/8', sanitize=True)
True
>>> harm.IPNetwork.sanitize('10.0.0.1/8')
'10.0.0.0/8'
```

e3rd:
From a partner, I've received a long list of IP networks, some of them have hosts bits set. The list will further be handled by sieve bot you showed me in Prague.

IntelMQ sanitization is great but of no use for me because the list will be stored in a separate sieve-config file.  
Since the list may get updated in the future, I didn't want to sanitize it manually.  
But if you found this a dumb idea, don't hesitate to reject the pull and I'll do a one time sanitization and in the case of a future update, I'll require a sanitized list from partner.

wagner-certat:
Ah, ok. Now I understand. The partially invalid network names are not in the data to be processed but in the sieve script itself.

I imagine the expert could be unstrict by "inaccurate" by default. Because if the network is part of the rules, it intentionally.

also cc @antoinet any comments on this change/idea?

e3rd:
@antoinet , could you please say yes/no? :) We'd like to use great sieve bot features, however we're little in a hurry.

antoinet:
Good point, and good that you implemented the parameter with `default=True` to maintain backwards compability. Are you also considering to implement corresponding unit tests?

wagner-certat:
IMO the unstrict parsing can be default because as the netmask is given, it is already explicit to the user that the value is a network, not an IP.

And I think the change does not break compatibility, does it?

e3rd:
If antoinet is not against, I'll add some unittests and make strict=False as default as Sebastian points-out. IMHO, it doesn't change back. comp.

wagner-certat:
> make strict=False as default

I wouldn't add the parameter at all then

antoinet:
on second thought, agree it doesn’t break compability... also agree that if default=False then it’s not needed. thx for the unittests 🙂

wagner-certat:
The build fail is in master, will push a fix for it.

e3rd:
Thanks a lot, guys, that you helped me merge this so quickly!

PR #1203, opened 2018-03-06
Microsoft Interflow and CTIP

wagner-certat:
Collector for Interflow API and parser for CTIP data.

Collector has some possibilities to reduce the amount of data to be processed (time- and name-based filter, cache as "memory")


wagner-certat:

PR #1205, opened 2018-03-06
Malware family names download & convert script

wagner-certat:
Downloads the current mapping from https://github.com/certtools/malware_name_mapping/ and converts it to the syntax applicable for the modify bot.

### DIFF
--- a/contrib/malware_name_mapping/README.md
+++ b/contrib/malware_name_mapping/README.md
@@ -1,0 +1,10 @@
+Malware Name Mapping
+====================
+
+This script fetches the current mapping file from the [Malware Name Mapping repository](https://github.com/certtools/malware_name_mapping/raw/master/mapping.txt) and converts it to the syntax needed by the modify expert.
+It expects a tab-separated file and uses the first column as regular expression and the second one as family name. Other URLs can be given as parameter.
+The resulting configuration sets the field `classification.identifier` to the malware family name corresponding to the given `malware.name`.
+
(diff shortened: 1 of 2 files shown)


contrib/malware_name_mapping/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+See also the built-in help of the script by calling it with `-h`.
+
+Default rule
+------------
+
+A default rule can optionally be added the the end of the rules with the `-d` flag.
+With this rule, the field `classification.identifier` is set to `malware.name` if the first is empty and the latter is not.
+As this is the last rule, it is kind of "last resort" to fill the identifier field.
+
```
aaronkaplan: proposal: let's also in the docu make an example (cron entry) how to use the .py script. Think: copy & paste

aaronkaplan: Another proposal ( I know this is a bit out of scope for this particular PR): make a check if the generated modify.conf file is actually valid (syntax check).

I am just thinking about the user-process here.

wagner-certat: > Another proposal ( I know this is a bit out of scope for this particular PR): make a check if the generated modify.conf file is actually valid (syntax check).

It's definitely valid json as we use the json lib here, so I will point the user to the `intelmq check` command.

wagner-certat: I added a section about automation

PR #1212, opened 2018-03-13
Twitter bugfix and docs

hugeox:
Changelog is updated with twitter entries.
Bots.md now contains twitter entries.
Whitelist in twitter parser now works.


wagner-certat:
Thanks!

PR #1213, opened 2018-03-13
shouldn't that be IPv6?

e3rd:


wagner-certat:
This file is generated from lib/harmonization.py, please fix it there. The program that generates the resulting docs is bin/intelmqctl_gen_harm_docs.py.

And we have a test for that^^

e3rd:
Fixed in harm .py too :)

wagner-certat:
Fixed the missing one in 457ab770

PR #1231, opened 2018-04-19
Added support for Shadowserver Accessible-Hadoop reports

antoinet:
Added mapping of Shadowserver accessible hadoop csv to intelmq in `config.py` according to https://www.shadowserver.org/wiki/pmwiki.php/Services/Accessible-Hadoop


wagner-certat:
> according to https://www.shadowserver.org/wiki/pmwiki.php/Services/Accessible-Hadoop

The documentation does not always reflect the actual data format sent, did you cross-check that?

antoinet:
Yes, I'm already running the code with live data.

wagner-certat:
Great, thanks!

Issue #1075, opened 2017-08-17
log level is info at beginning of interactive run


wagner-certat:
When running bots interactively, the log level is not debug at the beginning:
```
> intelmqctl run malc0de-parser 
intelmqctl: Starting malc0de-parser...
malc0de-parser: Malc0deParserBot initialized with id malc0de-parser and intelmq 1.1.0.alpha1 and python 3.4.6 (default, Mar 22 2017, 12:26:13) [GCC] as process 8765.
malc0de-parser: Bot is starting.
malc0de-parser: Loading source pipeline and queue 'malc0de-parser-queue'.
malc0de-parser: Connected to source queue.
malc0de-parser: Loading destination pipeline and queues ['deduplicator-expert-queue'].
malc0de-parser: Connected to destination queues.
malc0de-parser: Pipeline ready.
malc0de-parser: Waiting for incoming message.
```

In the debug mode, all parmeters should be logged, which is not the case here. Also, the configured log level *is* debug, so it should be shown in any case.

wagner-certat:
using bisecting I found this bad commit: 31ff5c0d20f05fbec16eba99f1d0583b2827cf34

cc @e3rd

e3rd:
Sorry, could you please explain further? While using `run` command, the initial logging level is set as follows:
`init_log_level = {"console": logging.DEBUG, "message": logging.WARNING, "process": logging.INFO, None: logging.INFO}`

* You mean that we should change `process` and `None` keys to `logging.DEBUG`?
* What do you mean by "Also, the configured log level is debug" – where is that configured please?

wagner-certat:
Okay, found the reason.

Before 31ff5c0 the log level DEBUG has been set for `intelmqctl run`. After that commit, the log level is INFO for `intelmqctl run` and DEBUG for `intelmqctl run botid console`.

The docs are still on the state before 31ff5c0 https://github.com/certtools/intelmq/blob/maintenance/docs/intelmqctl.md#run (DEBUG for run, not INFO). So the docs should be fixed.

Thinking more about it: As `intelmqctl run` is just the interactive version of `intelmqctl start` now, we could actually use the configured log level and not modify it at all in that case.

wagner-certat:
Hm, why should the run-command set a higher log level than the configured one? So I'd only set it to DEBUG for the console subcommand (because it's for debugging) at leave the others at the configured value. We could also add a commandline argument for that.

e3rd:
[...]

PR #1239, opened 2018-05-16
BUG/ENH: ctl run: log level handling

wagner-certat:
fixes certtools/intelmq#1075
introduces new loglevel parameter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,9 @@ CHANGELOG
 
 ### Tests
 
+### Tools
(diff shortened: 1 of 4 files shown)


wagner-certat:
Ad autocompletion: I agree. I'd like to have an automated solution, because maintaining it manually is hard.

e3rd:
Agree... but the only automated solution I imagine is to create an external library that takes `intelmqctl.py`, extracts all mentions of `argparse` and generates the `intelmq.sh`. But I don't dare to guess if that's an easy or a tough task.

wagner-certat:
-> #1240

PR #1243, opened 2018-05-28
Fix Shadowserver parser convert_http_host_and_url

th-certbund:
Both functions convert_hostname_and_url() and convert_httphost_and_url() were broken or incomplete. Replaced by new function convert_http_host_and_url().

### DIFF
--- a/intelmq/bots/parsers/shadowserver/config.py
+++ b/intelmq/bots/parsers/shadowserver/config.py
@@ -378,7 +378,7 @@ ipv6_sinkhole_http_drone = {
         ('source.reverse_dns', 'hostname'),
         ('extra.', 'sysdesc', validate_to_none),
         ('extra.', 'sysname', validate_to_none),
-        ('destination.url', 'url', convert_http_host_and_url, True),
+        ('destination.url', 'http_url', convert_http_host_and_url, True),
         ('extra.', 'http_agent', validate_to_none),
         ('destination.fqdn', 'http_host'),
         ('extra.', 'http_referer', validate_to_none),
--- a/intelmq/bots/parsers/shadowserver/config.py
+++ b/intelmq/bots/parsers/shadowserver/config.py
@@ -171,7 +171,7 @@ def convert_http_host_and_url(value, row):
 
         application = "http"
         if "application" in row:
-            if row['application']:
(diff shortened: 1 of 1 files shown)


intelmq/bots/parsers/shadowserver/config.py
```diff
@@ -138,37 +138,44 @@ def convert_float(value):
...
-            return row['application'] + '://' + row['hostname']
+    if hostname and path:
+        # remove potential leading/trailing HTTP request information
+        path = re.sub(r'^[^/]*', '', path)
+        path = re.sub(r'\s.*$', '', path)
 
-    return value
+        application = "http"
+        if "application" in row:
+            if row['application'] in ['http', 'https']:
```
wagner-certat: Otherwise it's an error, right?

th-certbund: Most events don't include a value for application at all, so we should use "http" as default to be used if host and url are set.
Some events have application set to "irc" or to strange values like "3.1002" but then there is no url and the value of application should be disregarded.
As "http" is default, this could also be removed from the list in line 174 so the value is only changed if application is set to "https".

intelmq/bots/parsers/shadowserver/config.py
```diff
@@ -138,37 +138,44 @@ def convert_float(value):
...
+    if "url" in row:
+        if row['url']:
+            path = row['url']
+    elif "http_url" in row:
+        if row['http_url']:
+            path = row['http_url']
 
-        elif row['hostname'] and not row['url']:
-            return row['application'] + '://' + row['hostname']
+    if hostname and path:
```
wagner-certat: The columns for the URL can be empty. Do you know if this means that the URL is unknown or is it equivalent with `/`?

th-certbund: I checked the Shadowserver reports for the past weeks.
If "url" is present (and does not include the full HTTP request :-/), it always starts with "/".
There are also lots of events with a url of only "/".
So I consider an empty value as the url not provided at all.

wagner-certat:
Also cherry-picked into maintenance for version 1.0.5.

Issue #1188, opened 2018-02-13
Elasticsearch output - dots in key names


cert-lv:
Hi,

Modern Elasticsearch supports dots in key names. Currently an output bot does a replacement ("." to "_" by default) without any options.

One solution could be to add a bool config switch to enable/disable the replacement. (Somewhen in the future, probably, this func could be removed at all).

Fast solution is to replace dots with dots.

Also this ticked is opened as you wanted to change the default/name of the parameter.

cert-lv:
**extra** keys are also concatenated with "_" without choice.
Result **extra.somekey** is allowed now, and is kind of more JSON-ish.

navtej:
@cert-lv last i checked ES stopped allowing "." in key name. You can use the `replacement_char` option, set its value to '.' and you should be good to go, if you ES instance supports '.''s. IIRC after 1.X they dropped support for dots. Even then the usage of dots was discouraged. However I haven't been following ES for a while now. May be that changed. Could you please let us know your ES version?

cert-lv:
Hello, @navtej ,

Yes, I remember the period when dots were not allowed. Now we use ES 6.x (with 5.x it was 
 Ok too), but it seems dots were fixed a long time ago: https://www.elastic.co/guide/en/elasticsearch/reference/2.4/dots-in-names.html

In short - Elasticsearch 2.x does not support dots-to-object transformation and so dots in field names are not allowed in versions 2.0 - 2.3. In modern ES such fields are converted into multilevel objects.

navtej:
@cert-lv does adding dot as replacement_char works for you?

cert-lv:
@navtej, yes, as I've mentioned fast solution is to replace dots with dots, it works.

I've also noticed, that putting nothing in **flatten_fields** disables concatenation of **extra** fields with "_", so the second question is solved too.

But one more question - currently you put everything into one single ES indice, **intelmq** by default. May I send a Pull Request with a small fix to split indices into names like **intelmq-MM-YYYY** ? So users can manage/delete/backup indices by month.

navtej:
@cert-lv please see this, https://github.com/navtej/intelmq/blob/latest/intelmq/bots/outputs/elasticsearch/output.py . it uses dynamic index based on date and time from time observation. I havent been able to make it generic because people may have vastly varied requirements.

navtej:
I guess we can close this issue or we still have something open on it?

wagner-certat:
[...]

PR #1251, opened 2018-06-04
WIP: Kafka/AVRO output bot.

z0r0:
This is my first output bot contribution, and should be considered a work in progress. 

This output bot goes a bit farther than simply outputting all threat intel to a kafka topic. The idea is that the intelligence can get routed to a topic that's dedicated to its source intelligence type. I'm opening this PR to solicit feedback on ironing out this bot.


z0r0:
First of my questions: in order to make avro work, i've got to define an avro schema. For IntelMQ, this is a large data type mapping that should be stored as a configuration template. An example can be seen [Here](https://pastebin.com/diVe1HH0). 

Another thing to note is that for output, especially to avro, fields need to be flattened, and all `.` chars need to be removed from field names, so these field names are statically defined in avro as `_`'s. I'd be curious about any thoughts about the liberties taken with regards to field renaming + flattening for this use case.

wagner-certat:
> First of my questions: in order to make avro work, i've got to define an avro schema. For IntelMQ, this is a large data type mapping that should be stored as a configuration template. An example can be seen Here.

Where's the question? :D 

> Another thing to note is that for output, especially to avro, fields need to be flattened, and all . chars need to be removed from field names, so these field names are statically defined in avro as _'s. I'd be curious about any thoughts about the liberties taken with regards to field renaming + flattening for this use case.

This is also the case for the elastic search bot, where a parameter has been used to specify the replacement character and `_` was the default. (Not necessary any more in ES, see #1188)

intelmq/bots/outputs/kafka/output.py
```diff
@@ -0,0 +1,126 @@
...
+try:
+    from confluent_kafka import Producer
+except ImportError:
+    Producer = None
+try:
+    from confluent_kafka import avro
+except ImportError:
+    avro = None
+
+if avro is not None:
```
wagner-certat: Can this and the two lines above be removed?

```
except ImportError:
    from confluent_kafka.avro import AvroProducer
```

intelmq/bots/outputs/kafka/output.py
```diff
@@ -0,0 +1,126 @@
...
+    from confluent_kafka.avro import AvroProducer
```
wagner-certat: Is this for a different library version or why is it necessary? Please document the reason inline.

And this line is not in a try-except block.

intelmq/bots/outputs/kafka/output.py
```diff
@@ -0,0 +1,126 @@
...
+                    for key, value in val.items():
+                        event_dict[field + '.' + key] = value
+                    event_dict.pop(field)
+
+        event_dict = replace_keys(event_dict)
+
+        if self.enable_avro is False:
+            self.producer.produce(self.kafka_topic, dumps(event_dict).encode('utf-8'), callback=self.delivery_report)
+            self.acknowledge_message()
+            self.kafka.flush()
```
wagner-certat: Where is `self.kafka` defined?

z0r0:
So the question is simply, where should i put the files in the source so they can be referenced? Should i add them in contrib? For my instance i put config files in /opt/intelmq/var/lib/bots/kafka-output/[key_file.avsc, topics.conf, value_file.avsc]

intelmq/bots/outputs/kafka/output.py
```diff
@@ -0,0 +1,126 @@
...
+                self.logger.debug('Event %s has no field %s, dropping.', format(event_dict), format(self.avro_topic_field))
+                self.acknowledge_message()
+                return
+
+            submit_key = {'indicator': event_dict[key_field]}
+            event_topic = self.avro_topic_schema[key_field]
+            self.logger.debug('Shipped %s to topic: %s', format(submit_key), format(event_topic))
+            self.producer.produce(topic=event_topic, value=event_dict, key=submit_key)
+            self.acknowledge_message()
+            self.producer.poll(0)
```
wagner-certat: Better do this check before acknowledging and raise an error if that happened so the message does not get lost.

intelmq/bots/outputs/kafka/output.py
z0r0: I'm just using this to throw better errors in init() for external dependencies. I can remove them if you'd like.

intelmq/bots/outputs/kafka/output.py
z0r0: Did this again for better error handling as i was writing this. It can probably be removed as there's multiple python packages which fit within the kafka namespace. many of them have Producer class definitions, but no avro class definitions. confluent_kafka is a more specific namespace.

intelmq/bots/outputs/kafka/output.py
z0r0: It doesn't look like the AvroProducer has much in terms of error handling, looking at it now, i don't think that poll() does anything. I think i'm just going to have to catch serialization errors on produce.

[Here's a link](https://github.com/confluentinc/confluent-kafka-python/blob/fb9d973a69e9b86afe50dd4260c4f20850bef815/confluent_kafka/avro/__init__.py#L16)

intelmq/bots/outputs/kafka/output.py
wagner-certat: > Did this again for better error handling as i was writing this.

But an ImportError is not catched here.

intelmq/bots/outputs/kafka/output.py
wagner-certat: If a message could not be successfully delivered, `self.acknowledge()` should not be called. For error handling (and logging), an Exception should be raised.

wagner-certat:
AFAIU this is still WIP, right?

Changing the Milestone then to the next version, 1.1.0 is now coming soon.

wagner-certat:
@z0r0 are you still working on this? The PR is in a pretty good shape and it would be a pity if we'd not continue the work.

z0r0:
Hello, we ended up not implementing it and went with something else at work. That being said, I'm going to finish this one up on my own time in the coming weeks, so stay tuned.

wagner-certat:
> Hello, we ended up not implementing it and went with something else at work. That being said, I'm going to finish this one up on my own time in the coming weeks, so stay tuned.

Sorry to hear that and thanks for your offer to finish it! Please let us know how we can support you.

PR #1263, opened 2018-06-27
bot debugger fix

e3rd:
collector bots have no input queue and no send_message(path) keyword

### DIFF
--- a/intelmq/lib/bot_debugger.py
+++ b/intelmq/lib/bot_debugger.py
@@ -82,7 +82,23 @@ class BotDebugger:
         module.set_trace()
 
     def _message(self, message_action_kind, msg):
-        if message_action_kind == "get":
+        if message_action_kind == "send":
+            if self.instance.group == "Output":
+                self.instance.logger.warning("Output bots can't send messages.")
+                return
+
+            if not bool(self.instance._Bot__destination_queues):
+                self.instance.logger.warning("Bot has no destination queues.")
+                return
+            if msg:
+                msg = self.arg2msg(msg)
+                self.instance.send_message(msg)
+                self.instance.logger.info("Message sent to output pipelines.")
+            else:
+                self.messageWizzard("Message missing!")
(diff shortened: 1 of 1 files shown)


wagner-certat:
> collector bots have no input queue and no send_message(path) keyword

Collectors don't have inputs, but they can send messages. Output's don't send messages.

e3rd:
Thanks, I've added the check that Output's don't send messages and rebased.

wagner-certat:
> Note that my commit doesn't prevent Collectors to send_message but only to send_message with the `path` keyword because it seems Collectors can't lead to a named queue.

Don't know a reason against it. The currently existing one's don't do it, but for custom ones it's possible.
With 30bba26208d25dd1161a12592caf44a5d05005e3 I pushed a bugfix

intelmq/lib/bot_debugger.py
```diff
@@ -12,6 +12,7 @@
  * processes single message, either injected or from default pipeline (process subcommand)
  * reads the message from input pipeline or send a message to output pipeline (message subcommand)
 """
+import inspect
```
wagner-certat: What do you use it for?

e3rd: (sorry, for nothing, removing)

e3rd:
Ah, okay.

So this PR covers the fix that Collector bot cannot perform "get" and "pop" opererations (because collectors have no input queue) and that Output bot cannot perform send operation (because outpus have no output queue). Does this make sense to you like this?

I've changed the commit message.

wagner-certat:
Thanks! Merged into all branches

78ed23ab8bcff50c20c522865f7e0f875db38f3f
19854c265cac748268abf6c9423faca74ef60dc0
3634ab45489955d79fe660d1e8568a7ad5d1a9df
fd2b67098bd9b8538290edfc49118d5079b54c9e
5de96a57831e6e3934bbebadf2584e0c2b88cadc

PR #1267, opened 2018-07-06
Recorded Future IP Risklist expert bot

olekristoffer:
A new expert bot to set risk score associated with either source.ip or destination.ip
It uses data obtained from recorded future API which is a paid and closed service.

### DIFF
--- a/intelmq/bots/experts/recordedfuture_iprisk/expert.py
+++ b/intelmq/bots/experts/recordedfuture_iprisk/expert.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+"""
+See README for database download.
+"""
+
+from intelmq.lib.bot import Bot
+import csv
+
+
+class RecordedFutureIPRiskExpertBot(Bot):
+
+    database = dict()
+
+    def init(self):
+        self.logger.info("Loading recorded future risk list.")
+
+        try:
+            with open(self.parameters.database) as fp:
+                rfreader = csv.DictReader(fp)
+                for row in rfreader:
+                    self.database[row['Name']] = row['Risk']
+
+        except IOError:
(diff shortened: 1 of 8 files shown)


intelmq/bots/experts/recordedfuture_iprisk/expert.py
```diff
@@ -0,0 +1,46 @@
...
+        self.overwrite = getattr(self.parameters, 'overwrite', False)
+
+    def process(self):
+        event = self.receive_message()
+
+        for key in ["source", "destination"]:
+            if key + '.ip' in event:
+                if "extra.rf_iprisk." + key not in event:
+                    if event.get(key + '.ip') in self.database:
+                        event.add("extra.rf_iprisk." + key, int(self.database[event.get(key + '.ip')]))
```
wagner-certat: You can use `self.database.get(event.get(key + '.ip'), 0)` here instead of the four lines. Alternatively: [defaultdict](https://docs.python.org/3/library/collections.html?highlight=defaultdict#collections.defaultdict) :)

intelmq/bots/experts/recordedfuture_iprisk/expert.py
```diff
@@ -0,0 +1,46 @@
...
+    database = dict()
+
+    def init(self):
+        self.logger.info("Loading recorded future risk list.")
+
+        try:
+            with open(self.parameters.database) as fp:
+                rfreader = csv.DictReader(fp)
+                for row in rfreader:
+                    self.database[row['Name']] = row['Risk']
```
wagner-certat: Better convert to int here instead doing in the process() call

olekristoffer:
Thank you for the review. I'll incorporate the requested changes. :)

PR #1270, opened 2018-07-12
Add Gzip decompression to http collector

kbrajneesh:


wagner-certat:
Hi,

First, sorry for the late response.

It was not initially clear to me what the parameter is exactly doing and what it is for. The docs can be improved here (also inline in the code).
This additional parameter allows to unzip files which are gzip-compressed, but not a tar. So `.gz` files. We currently *do* support `.tar.gz` files with the `extract_files` parameter. I'm not very inclined to add another third decompression parameter. I propose to merge this functionality "into" the `extract_files` parameter.

Would this work for you?

wagner-certat:
Thanks.

I cherry-picked the two relevant commits and moved some of the logic to the utils library file.

wagner-certat:
The commits are bd734cb686854189bb4c20a72c769f661edae836 dcc50a41965959adb3519eda07ae3eec3dfaeaf6 and 66ac19edde99def107344053f3c0c351ee46a994

wagner-certat:
FYI, I rewrote the utils.unzip function in 4b09fd3e79bb22d6b62eb85d53914a3e7856eeb9 a bit

PR #1272, opened 2018-07-16
named queues displayed when querying intelmqctl for queues in JSON

e3rd:
Intelmq manager will be able to work with named queues thanks to this commit.


e3rd:
I realized this is not needed for manager that loads directly pipeline.conf file.

PR #1275, opened 2018-07-19
Fix: don't display named paths amongst standard queues.

e3rd:
Queue path names were incorrectly listed amongst standard queue names. Named queue got displayed twice – as a standard queue and as a named path. Note that this information didn't serve anything because multiple queues may have the same name, ex: _default. Plus, manager got confused, having no way to distinguinsh path names from standard queue names.

> intelmqctl list queues
intelmqctl: _default - 0
intelmqctl: abuse-ch-feodo-tracker-domains-parser-queue - 0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ CHANGELOG
 
 ### Tests
 
+### Tools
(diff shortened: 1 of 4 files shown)


wagner-certat:
Thanks, also cherry-picked it for 1.1.0

Issue #1274, opened 2018-07-17
Displaying queue traffic


e3rd:
I'd like to discuss displaying queue traffic in manager. It would be cool we see what pipes had recently transferred some messages on the visjs edges in manager. There are plenty of ways this can be done. There is one solution I made up last night but I need to discuss the intelmq background first.

* `defaults.conf` value: "count messages in queues" on/off
* `pipeline.py`: For other message passed, write to the registry that there was a connection right now between the bot and a destination queue.
```python
for destination_queue in queues:            
    self.pipe.lpush(destination_queue, message)
    self.pipe.lset("counter", (bot_id, destination_queue), True, time-to-live=5 seconds)  # ← new line
```

* `intelmqctl --type json list queues-and-statuses --last-checked TIMESTAMP` would return also the list of connections between a bot and a queue that happened since TIMESTAMP. With this info provided, Manager / Configuration tab could nicely pulsate the active edges.

**Downside:** Isn't it a performance issue to perform `lset` for every parsed message (`lpush`)? What do you think?

wagner-certat:
see also #361 There was the idea to create the statistical data in/after the process() call - to differentiate between success and failure.
Another (additional) idea is to just save timestamps there to keep the processing overhead in the bot as low as possible. The the rate is then calculated "on demand" (allows rates for different timespans, graphs etc)

e3rd:
We merged first draft of queue traffic into develop. You wrote:

> However I am not sure if it is worth now to implement it in the Manager, maybe we can find some optimizations in the way we collect or store the data, which would need some changes.

When you say it seems ready to implement in the Manager, I propose to implement it this way.

Model:
```md
## Model
bot-id
  -> _default -> another-queue -> another-bot
  -> alter-path -> alternative-queue -> alternative-bot

## Queues
bot-id-queue: 103
another-bot-queue: 104
alternative-bot-queue: 105

## Piping
bot-id is piping 3 messages via alter-path, in the past it piped 12 messages that way

## Status
bot-id had 50 successes and 13 failures in message processing.
```

Current output of `intelmqctl --type json  list queues`: 
```json
# bot statuses
{"bot-id": "running"}

# bot information
{"bot-id": {"source_queue": ["bot-id-queue", 103], 
            "internal_queue": 0,
            "destination_queues": [
               ["another-queue", 104], 
               ["alternative-queue", 105]
             ]
           }
},
```

Proposal output of `intelmqctl --type json  list queues`:
```js
# bots
[...]

Issue #361, opened 2015-10-07
ENH: idea - keep event rate (events/sec) in bot and show in intelmq-manager


aaronkaplan:
While observing intelmq-manager and monitoring the throughput of different bots, I noticed that you can only get a snapshot in time of the queue length. However, the snapshot really depends on the processing speed of the bot and when the snapshot is done. This does not tell you very well, how fast a bot is processing data.
Therefore, I'd like to propose that each bot keeps an internal counter for an event rate (events/sec) and that it exposes that info to intelmq-manager.

sebix:
We had the idea to save such statistical data and status information in redis as simple keys. The bot writes his data there after one `process()`-call.

PR #1277, opened 2018-07-20
queue traffic #361 #1274

e3rd:
According to the open issues, I've drafted this PR, open to discussion.

Messages are counted in `send_message` (one liner), the rest is measured and handled in `start` method. ((Collectors don't `acknowledge`, Outputs don't `send`)). Every two seconds, current status is output to the redis.
I think this may solve our problems – it seems to me there is no big overhead.

When we agree, I may add the functionality in `intelmqctl` to `queues-and-status` from where manager'll read them.

I am not sure what queue should I write into, the destination, the source, or another?


wagner-certat:
Thanks @e3rd for your work, I think it is a good start.

Some of the changes I implemented:
 * merge with current develop and resolve conflicts
 * instead of using the pipeline database, use database 3
 * use a Cache object to handle the connection with redis
 * log errors, but only in DEBUG logging level
 * instead of using keys like `"('total', 'spamhaus-drop-collector', '_default')"` in redis, use `"deduplicator-expert.total._default"`, much easier to parse.
 * reset internal statistics after reload (like re-initialization)

I will test that in production now, including some visualizations using our monitoring system to see if it is useful. However I am not sure if it is worth now to implement it in the Manager, maybe we can find some optimizations in the way we collect or store the data, which would need some changes.

intelmq/lib/bot.py
```diff
@@ -264,14 +287,38 @@ def start(self, starting: bool = True, error_on_pipeline: bool = True,
...
+        if not (force or datetime.now() - self.__message_counter["stats_timestamp"] > self.__message_counter_delay):
+            return
+
+        try:
+            for path, n in self.__message_counter["path"].items():  # current queue traffic
+                self.__stats_cache.set(".".join((self.__bot_id, "temporary", path)), n, ttl=2)
+                self.__message_counter["path_total"][path] += n
+                self.__message_counter["path"][path] = 0
+            for path, n in self.__message_counter["path_total"].items():  # total queue traffic
+                self.__stats_cache.set(".".join((self.__bot_id, "total", path)), n)
```
e3rd: Since you're using Cache.set without ttl, the redis command will be SETNX – in other words, the initial value is never replaced. We want this value to endure even it the bot is stopped, right?

wagner-certat: Oh, thanks for finding this. It should be set of course. But also for all the caches IMO.

wagner-certat: ac66a916db36db996a8042778ccfabe4448c9b86

Issue #554, opened 2016-07-05
Support Abusehelper input


dmth:
To be capable to process data from Abushelper we need to have two working pieces of code
- [x] XMPP Support
    State: Kind of stable, Output needs MUC, though
- [ ] Second: A Abusehelper Parser
    State: Postponed to 1.1

sebix:
xmpp: done in #669 
abusehelper parser: postpone

dmth:
Already in April 16, we did a comparison of abusehelper and intelmq dho's.  I attached the results. Unfortunately github does not support .ods files but .xlsx seems to be ok.
[Data Harmonization.xlsx](https://github.com/certtools/intelmq/files/642273/Data.Harmonization.xlsx)

sebix:
`feeder` is `feed.provider` and `source cc` is `source.geolocation.cc`, then the mapping IntelMQ -> AH is complete.

wagner-certat:
The meaning of `ip`, `source ip` and `destination ip` is:
`ip` is always one of the other two.
`ip` is the one to complain about.
source and destination have the "netflow" meaning.

So: `ip` is `source.ip`, the other IP in the the AH-Data is the `destination.ip`.

wagner-certat:
AbuseHelper harmonization: https://github.com/abusesa/abusehelper/blob/master/docs/Harmonization.md

Updated mapping: [Data-Harmonization-imq-abusehelper.xlsx](https://github.com/certtools/intelmq/files/2034244/Data-Harmonization-imq-abusehelper.xlsx)

wagner-certat:
Is there still a need for this by anyone?

wagner-certat:
Closing because of inactivity. Please reopen if needed.

PR #1287, opened 2018-08-08
WIP : created parser for CERT-EU csv

gbkss:
CERT-EU provides a private feed and they will update their format on the 2018-08-20. Still WIP:

* we only have two CSV entries
* there is a mapping from their taxonomy (they use the one from abuse), I simply used a Python dict to map it to IntelMQ's taxonomy. I used #554 for the mapping.
* some fields from CERT-EU aren't explained so I don't know / am not sure to what it should be mapped.
* some fields have values which are not defined (e.g. confidence level)

I created CERT-EU CSV but they provide a JSON as-well, I am unsure how this should be handled. Should another bot for JSON be created?


intelmq/bots/parsers/cert_eu/parser_csv.py
```diff
@@ -0,0 +1,76 @@
...
+        event["source.geolocation.latitude"] = line["source latitude"]
+        event["source.geolocation.city"] = line["source city"]
+        event["source.geolocation.geoip_cc"] = line["source cc"]
+        event["source.geolocation.longitude"] = line["source longitude"]
+        # event["?"] = line["first_seen"]
+        # event["?"] = line["num_sensors"]
+        event["feed.accuracy"] = line["confidence level"]  # TBC is 0<=x<=100 ?
+        # event["?"] = line["last seen"]
+        event["event_description.target"] = line["target"]  # TBC
+        event["source.url"] = line["phishing"]  # TBC or destination.url?
```
antoinet: I think this should be mapped from a field called "url" (according to [abusehelper](https://github.com/abusesa/abusehelper/blob/master/docs/Harmonization.md)). Maybe it's an error in the example supplied by CERT-EU. I will ask them for feedback.

intelmq/bots/parsers/cert_eu/parser_csv.py
```diff
@@ -0,0 +1,76 @@
...
+
+    abuse_to_intelmq = {
+        "backdoor": "backdoor",
+        "blacklist": "blacklist",
+        "botnet drone": "botnet drone",
+        "brute-force": "brute-force",
+        "c&c": "c&c",
+        "compromised server": "compromised",
+        "ddos infrastructure": "ddos",
+        "ddos target": "--",
```
antoinet: IntelMQ will refuse the value `--`. We need to think how to map that value to another eCSIRT/IntelMQ type.

gbkss:
Any idea why it fails for certain version of Python? It seems that it only affects base64.decode()

intelmq/bots/parsers/cert_eu/parser_csv.py
g-rd: yes, this looks like a mistake in our sanitizer config for a feed, I will look into it.

intelmq/bots/parsers/cert_eu/parser_csv.py
```diff
@@ -0,0 +1,76 @@
...
+        "datasource", "source ip", "observation time", "tlp", "description",
+        "type", "count", "source time", "source country", "protocol",
+        "destination port", "source latitude", "source city", "source cc",
+        "source longitude", "first_seen", "num_sensors", "confidence level",
+        "last_seen", "target", "phishing", "asn", "domain name"]
+    ignore_lines_starting = ["#"]
+
+    def parse_line(self, line, report):
+        event = self.new_event(report)
+        # event["?"] = line["datasource"]
```
wagner-certat: Can be put into extra

intelmq/bots/parsers/cert_eu/parser_csv.py
```diff
@@ -0,0 +1,76 @@
...
+        event["source.ip"] = line["source ip"]
+        event["time.observation"] = line["observation time"]
+        event["tlp"] = line["tlp"]
+        event["event_description.text"] = line["description"]  # TBC
```
wagner-certat: yes, correct

intelmq/bots/parsers/cert_eu/parser_csv.py
```diff
@@ -0,0 +1,76 @@
...
+        event["classification.type"] = self.abuse_to_intelmq.get(
+            line["type"], "unknown"
+        )
+        # event["?"] = line["count"]
```
wagner-certat: can be put into extra if it exists

intelmq/bots/parsers/cert_eu/parser_csv.py
```diff
@@ -0,0 +1,76 @@
...
+        event["time.source"] = line["source time"]
+        event["source.geolocation.country"] = line["source country"]
+        event["protocol.application"] = line["protocol"]
+        event["destination.port"] = line["destination port"]
+        event["source.geolocation.latitude"] = line["source latitude"]
+        event["source.geolocation.city"] = line["source city"]
+        event["source.geolocation.geoip_cc"] = line["source cc"]
+        event["source.geolocation.longitude"] = line["source longitude"]
+        # event["?"] = line["first_seen"]
```
wagner-certat: can be put into extra if it exists

intelmq/bots/parsers/cert_eu/parser_csv.py
```diff
@@ -0,0 +1,76 @@
...
+        # event["?"] = line["num_sensors"]
```
wagner-certat: can be put into extra if it exists

intelmq/bots/parsers/cert_eu/parser_csv.py
```diff
@@ -0,0 +1,76 @@
...
+        event["feed.accuracy"] = line["confidence level"]  # TBC is 0<=x<=100 ?
```
wagner-certat: yes, that's the range.

intelmq/bots/parsers/cert_eu/parser_csv.py
```diff
@@ -0,0 +1,76 @@
...
+        # event["?"] = line["last seen"]
```
wagner-certat: can be put into extra if it exists

intelmq/bots/parsers/cert_eu/parser_csv.py
```diff
@@ -0,0 +1,76 @@
...
+        event["event_description.target"] = line["target"]  # TBC
```
wagner-certat: yes

intelmq/bots/parsers/cert_eu/parser_csv.py
```diff
@@ -0,0 +1,76 @@
...
+        event["source.url"] = line["phishing"]  # TBC or destination.url?
+        event["source.asn"] = line["asn"]  # TBC or destination?
```
wagner-certat: AFAIU the abuse helper taxonomy this and the line below are correct, yes. It is also matches with the IP address in the example

g-rd:
hi everyone, so I am in fault for creating this issue, we are in progress of redoing our abusehelper and this includes email that we send out.
We have invested time into trying to unify the ontology of all the feeds that we gather, in some cases it is very hard to determine what is a correct way to do that, especially to find a perfect way to fit everything to abusehelper ontology, so as of the phishing key mentioned, I think its a mistake on my part somewhere in the config .
About the email attachments we send out:

- there are two formats, csv and json in the attachment (json for automation and csv for manual investigation).
- We use https://github.com/abusesa/abusehelper/blob/master/docs/Harmonization.md ontology.
-  All values should be unicode since abusehelper forces everything to be unicode.

I would suggest to use the json format for importing to intelMQ, since csv is actually meant to go for people who needs to use excel for looking into the data (yes, there are still people doing that), also, I needed to change the csv delimiter because some uris can use the standard csv delimiters and makes the parsing break (I changed it to "|"), really sorry about that.

but the best way I think would be to set up xmpp exchange and not rely on emails if you are using intelmq or abusehelper.

wagner-certat:
@g-rd 

I prefer to use JSON as causes a lot of problems, but I'm not against a CSV parser in this PR.

Please don't use | as delimiter, use quotes. Or a csv library which does it for you.

About other methods: happy to hear that XMPP is supported. The announce mail mention only "other methods" without specifying them. We (at cert.at) will definitely use xmpp and json for this feed. So, for the intelmq part: Can I document the XMPP-bridge here? AFAIK other certs in EU would like to use this bridge with intelmq too.

g-rd:
@wagner-certat 
Hmm, I can change | to quotes if its more preferable, I just couldn't use comma or semicolon.
can you provide an example of what you are thinking of ? ('one', 'two', 'three' ?)
Why not use pipes as delimiter ?

wagner-certat:
Just use a standard csv writer part of your favorite programming language (what are you using?) and you won't have problems. You don't need to reinvent the wheel ;)

    "SR","109.125.93.10","2018-07-06 10:59:06Z","green","A URL is the most common resource with reference to malware binary distribution.","malware url",6,"2018-07-06 08:52:16Z","Germany","smbd",445,48.137428,"Munich","DE",11.57549,"2018-07-06T08:34:55.759000",1,50,"2018-07-06T08:52:16.109000",,,,

or similar, as long as strings are enclosed in quotes. And it's even possible to quote quotes. By using a library dealing with, you don't need to care :)

antoinet:
Hi all, I implemented some of the changes according to the discussion, but I don't know how to add them directly to this PR. Instead, I made a PR in @gbkss's fork here: https://github.com/gbkss/intelmq/pull/1. Is this the right way to do it?

wagner-certat:
Alternatively you can open a new PR from your branch (or I just pull in your branch)

antoinet:
If I open a new PR from my branch, will it be merged with this PR?

wagner-certat:
> If I open a new PR from my branch, will it be merged with this PR?

As this PR is part of the new PR, both are merged, once I merge your new one.

wagner-certat:
Thanks, fixed the tests and merged.

84c99386b78d104982dd84f93e91777ddad14ab2

PR #1302, opened 2018-08-20
Update Developers-Guide.md

aaronkaplan:
be consistent and if you use nosetests3, always use it.


wagner-certat:
As said in person, it depends on the installation. I'll add a very short explanation and examples.

PR #1308, opened 2018-09-02
ENH: add TLP field to elasticsearchmapper

SYNchroACK:


wagner-certat:
Cherry-picked to maintenance (for next 1.1.1) and added a changelog entry in 36e753d2f

PR #1313, opened 2018-09-04
surbl parser

dargen3:
There is new Surbl parser. It parse data from Rsync collector. It is necessarily to have opened up Sponsored Datafeed Service (SDS) access to the SURBL data via rsync to use Surbl parser.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ CHANGELOG
 - `intelmq.bots.parsers.generic.parser_csv`:
   - New parameter `columns_required` to optionally ignore parse errors for columns.
 - added `intelmq.bots.parsers.cert_eu.parser_csv` (#1287).
+- added `intelmq.bots.parsers.surbl.surbl`
 
 #### Experts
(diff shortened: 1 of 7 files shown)


dargen3:
I tried to solve error from Travis:

AssertionError: '# Av[1403 chars]\n- [Taichung](#taichung)\n- [Team Cymru](#tea[48820 chars]s:**' != '# Av[1403 chars]\n- [Surbl](#surbl)\n- [Taichung](#taichung)\n[49431 chars]s:**'

but i don't understand, what is wrong. Please, can you explain me these error?

wagner-certat:
It says:

> docs/Feeds.md does not match the output of intelmq/bin/intelmq_gen_docs.py to update the file.

The message misses some words (oops, my fault). I'll add them :)

The message should say that the feeds should be added in etc/feeds.yaml and then the markdown file should be generated with intelmq/bin/intelmq_gen_docs.py.

The codestyle check also fails: https://travis-ci.org/certtools/intelmq/jobs/424200320#L532

wagner-certat:
Thanks again!

PR #1317, opened 2018-09-06
RIPE Expert bot

kbrajneesh:


wagner-certat:
Thanks, looks good.

I think about merging the two bot and having only one "RIPE" bot. The other has the wrong name, so then we would need some code for backwards compatibility module. Some work for me :)

wagner-certat:
I changed the URL to https://stat.ripe.net/data/maxmind-geo-lite/data.json?resource= which provides current data. The old did only provide data until 2018-03

I also merged the two bots adding a compatibility mode for the old name.

See 6983bd7e45f71a0c3c29f5535c109b5472e067e9 for all changes.

PR #1318, opened 2018-09-06
MISP Parser to recognize attribtes nested in objects

tk-hendrik:
Hi,

just recognized the current parser simply ignores MISP Attributes if they are nested into an MISP Object.

Would like to see this minor code change go upstream since many misp sharing instances are using objects to group attributes together.

Regards,
Hendrik


wagner-certat:
Thanks!

Issue #1314, opened 2018-09-04
contrib: feeds-config-generator does not add feed name as parameter


SYNchroACK:
```
# python3 /opt/dev_intelmq/contrib/feeds-config-generator/intelmq_gen_feeds_conf --feeds-file=/opt/dev_intelmq/intelmq/etc/feeds.yaml

Select a feed: 57

RUNTIME CONFIGURATION:

{
    "openphish-phishing-collector": {
        "description": "placeholder",
        "group": "Collector",
        "module": "intelmq.bots.collectors.http.collector_http",
        "name": "OpenPhish Phishing",
        "parameters": {
            "http_url": "https://www.openphish.com/feed.txt",
            "rate_limit": 86400
        }
    },
    "openphish-phishing-parser": {
        "description": "placeholder",
        "group": "Parser",
        "module": "intelmq.bots.parsers.openphish.parser",
        "name": "OpenPhish Phishing",
        "parameters": {}
    }
}




PIPELINE CONFIGURATION:

{
    "openphish-phishing-collector": {
        "destination-queues": [
            "openphish-phishing-parser-queue"
        ]
    },
    "openphish-phishing-parser": {
        "source-queue": "openphish-phishing-parser-queue",
        "destination-queues": [
            "FIXME"
        ]
    }
}
```

```
# cat /opt/intelmq/etc/pipeline.conf 
{
    "openphish-phishing-collector": {
        "destination-queues": [
            "openphish-phishing-parser-queue"
        ]
    },
    "openphish-phishing-parser": {
        "source-queue": "openphish-phishing-parser-queue",
        "destination-queues": [
            "FIXME"
        ]
    }
}
```

```
# cat /opt/intelmq/etc/runtime.conf 
{
    "openphish-phishing-collector": {
        "description": "placeholder",
        "group": "Collector",
        "module": "intelmq.bots.collectors.http.collector_http",
        "name": "OpenPhish Phishing",
        "parameters": {
            "http_url": "https://www.openphish.com/feed.txt",
            "rate_limit": 86400
        }
    },
    "openphish-phishing-parser": {
        "description": "placeholder",
        "group": "Parser",
        "module": "intelmq.bots.parsers.openphish.parser",
        "name": "OpenPhish Phishing",
        "parameters": {}
    }
}
```

```
# su - intelmq

$ intelmqctl status 
intelmqctl: Bot openphish-phishing-collector is stopped.
intelmqctl: Bot openphish-phishing-parser is stopped.

[...]

PR #1321, opened 2018-09-07
Name and Provider fields cleaning

SYNchroACK:
Solution for: #1314


wagner-certat:
I fixed the bin/rewrite_config_files to do the BOTS sorting (more) correctly inad5afcb4e62f2e5bfe8ab739462fed64f895f1e8
So you can use this file to fix it. I did not apply the scrip to the file yet to avoid conflicts.

Issue #1150, opened 2018-01-09
ES output: TLS


wagner-certat:
The elasticsearch library seems to support TLS since version 6.1.1: https://github.com/elastic/elasticsearch-py/compare/6.0.0...6.1.1

navtej:
I dont have 6.XX setup, I will look into it.

wagner-certat:
Thanks!

PR #1328, opened 2018-09-18
Elasticsearch Output: Support rotating indices

stone-z:
For long term performance and data retention purposes, I need to periodically rotate the index being used to store IntelMQ events. I would love to contribute to the project, so for anyone else who would like to avoid an infinitely-growing Elasticsearch index, I offer this implementation.

Includes:
- Add 'rotate_index' boolean parameter to Output Bot (default maintains original behavior)
- Add --index-template parameter to elasticmapper to configure things on the ES side
- Doc updates for new parameters and style
- Tests for new behavior


wagner-certat:
@navtej and @SYNchroACK can you have a look at it please? Thanks.

navtej:
stone-z, can we address these alongside this.

1. rotation of indicies could be either daily, weekly or monthly. Currently you are taking into account only daily. I have a crude implementation for monthly rotation here https://github.com/navtej/intelmq/blob/latest/intelmq/bots/outputs/elasticsearch/output.py , do you think you will be able to accommodate that as well? We can probably skip weekly rotation.
2. can we address #1150 as well with this? Basically making the ES connection with TLS.
3. also lets address #1188 , disable '.' in default options.

If you are ok with these changes, then I can do a holistic review after you changes? Thx.

stone-z:
@navtej 
1. Sounds good, I'll make it a keyword argument (e.g. 'daily', 'monthly').
2. I'm willing to look into TLS, but in a separate PR. Is there a current test case implementation for any bot using TLS certs?
3. To clarify, the new desired behavior is to keep the '.' in field names by default, but allow replacement_char to be set for supporting ES 2.x? That sounds doable for this PR.

wagner-certat:
> Is there a current test case implementation for any bot using TLS certs?

Do you want Client Certs or CA Paths/files? We have some bots with these kind of parameters, have a look at the HTTP parameters: https://github.com/certtools/intelmq/blob/develop/docs/Bots.md#common-parameters (call Bot.set_request_parameters to parse them)

navtej:
> @navtej
> 
> 1. Sounds good, I'll make it a keyword argument (e.g. 'daily', 'monthly').
> 2. I'm willing to look into TLS, but in a separate PR. Is there a current test case implementation for any bot using TLS certs?
> 3. To clarify, the new desired behavior is to keep the '.' in field names by default, but allow replacement_char to be set for supporting ES 2.x? That sounds doable for this PR.

You understanding is correct.

stone-z:
Ok, PR has been updated:
- rotate_index accepts 'never', 'daily', 'weekly', 'monthly', 'yearly'. Default: null = 'never'
- replacement_char default is null. Will apply replacement chars if set
- updated tests
- updated README and Bots.md for Elasticsearch output

navtej:
@wagner-certat This looks fine.
@stone-z nice work. Thanks for this.

wagner-certat:
Thanks!

intelmq/tests/bots/outputs/elasticsearch/test_output.py
```diff
@@ -53,6 +135,131 @@ def test_event(self):
...
 
+    def test_raise_when_no_template(self):
+        """
+        Test that a bot raises a RuntimeError if 'rotate_index' is set, but a matching template doesn't exist in ES.
+        """
+        self.sysconfig = {"flatten_fields": "extra",
+                          "elastic_index": "intelmq",
+                          "elastic_doctype": "events",
+                          "rotate_index": "daily"}
+        self.assertRaises(RuntimeError, self.run_bot())
```
wagner-certat: I now noticed that this probably does not work as intended: `self.run_bot()` is run before `assertRaises` and thus the call is `self.assertRaises(RuntimeError, None)`. Because of a bug in `assertRaises` in python < 3.8, this does not error, but breaks with Python 3.8:

```
FAIL: Test that a bot raises a RuntimeError if 'rotate_index' is set, but a matching template doesn't exist in ES.
  File "/home/travis/build/wagner-certat/intelmq/intelmq/tests/bots/outputs/elasticsearch/test_output.py", line 146, in test_raise_when_no_template
    self.assertRaises(RuntimeError, self.run_bot)
AssertionError: RuntimeError not raised by run_bot
```
https://travis-ci.org/wagner-certat/intelmq/jobs/622711676#L808

@stone-z is it possible for you to look into this?

PR #1330, opened 2018-09-26
Add Shadowserver Outdated-DNSSEC-Key(-IPv6)

th-certbund:


th-certbund:
Shadowserver confirmed request to change 'cc_*' to 'dst_*' in the csv header.

wagner-certat:
Could also be classified as misconfiguration, but outage is also interesting

cc @aaronkaplan

th-certbund:
I also thought of a new type "misconfiguration" because "outage" is more physical... I'll propose to add this to the eCSIRT taxonomy.

wagner-certat:
> I also thought of a new type "misconfiguration" because "outage" is more physical... I'll propose to add this to the eCSIRT taxonomy.

If you are fast it could be discussed today

PR #1344, opened 2018-10-23
Maxmind: Use registered_country instead of country

th-certbund:
MaxMind provides different country attributes for an IP:
- country: Country where MaxMind _believes_ the user is located.
- registered_country: Country registered with the network range.

Incidents are usually reported to the network owner responsible for the affected IP or the respective national CSIRT. So in particular for CDN networks or satellite providers, we are not interested in the country a user is (potentially) located but in the location of the network owner.

Example for DE vs. US:

```
network:IP-Network:38.87.230.0/23
network:Org-Name:Interlax Telecom
network:Street-Address:50 NE 9th St.
network:City:Miami
network:State:FL
network:Country:US
```

Maxmind results:
```
country: DE
registered_country: US
```

Example for DE vs. AF:
```
inetnum:        180.222.136.0 - 180.222.143.0
netname:        Etisalat
descr:          Etisalat Afghanistan
country:        AF
```

Maxmind results:
```
country: DE
registered_country: AF
```

### DIFF
--- a/intelmq/bots/experts/maxmind_geoip/expert.py
+++ b/intelmq/bots/experts/maxmind_geoip/expert.py
@@ -43,8 +43,8 @@ class GeoIPExpertBot(Bot):
             try:
                 info = self.database.city(ip)
 
-                if info.country.iso_code:
-                    event.add(geo_key % "cc", info.country.iso_code,
+                if info.registered_country.iso_code:
+                    event.add(geo_key % "cc", info.registered_country.iso_code,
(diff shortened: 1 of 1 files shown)


wagner-certat:
cc @aaronkaplan

wagner-certat:
I extended it a bit by introducing a parameter `use_registered` to ensure backwards-compatibility and the possibility to chose. And at the chance I also fixed the docs.
26a64e03b

PR #1350, opened 2018-11-13
BUG: Update taxonomy mapping to enisa version

wagner-certat:
Updates from the September meeting of the working group

I did not rename any fields yet to be backwards-compatibly (this is for the bug fix release)

see also https://github.com/enisaeu/Reference-Security-Incident-Taxonomy-Task-Force/issues/29

ping @th-certbund

### DIFF
--- a/intelmq/bots/experts/taxonomy/expert.py
+++ b/intelmq/bots/experts/taxonomy/expert.py
@@ -1,4 +1,10 @@
 # -*- coding: utf-8 -*-
+"""
+The mapping follows
+Reference Security Incident Taxonomy Working Group – RSIT WG
+https://github.com/enisaeu/Reference-Security-Incident-Taxonomy-Task-Force/
+with extensions.
+"""
 
 from intelmq.lib.bot import Bot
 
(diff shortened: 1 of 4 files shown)


wagner-certat:
@pharook Could you please check the mappings to IDEA? Especially `malware-distribution`, `burglary`, `data-loss`.

intelmq/bots/experts/taxonomy/expert.py
```diff
@@ -1,40 +1,70 @@
...
-    "brute-force": "intrusion attempts",
-    "ids alert": "intrusion attempts",
-    "defacement": "intrusions",
-    "compromised": "intrusions",
-    "backdoor": "intrusions",
-    "vulnerable service": "vulnerable",
-    "vulnerable client": "vulnerable",
+    "infected system": "malicious code",  # ENISA eCSIRT-II taxonomy: 'infected-system'
+    "malware": "malicious code",
+    "malware configuration": "malicious code",  # ENISA eCSIRT-II taxonomy: 'malware-configuration'
```
th-certbund: Should be "malware-configuration" (dash instead of blank)

th-certbund:
In the RSIT, all machine readable values for classification and examples are written with dashes (no blanks). Here we currently have a wild mixture both in the types ("vulnerable service" vs. "vulnerable-system") and in the taxonomies ("malicious code" vs. "abusive-content").
I'd propose to use either dashes or blanks in all values.

wagner-certat:
> In the RSIT, all machine readable values for classification and examples are written with dashes (no blanks). Here we currently have a wild mixture both in the types ("vulnerable service" vs. "vulnerable-system") and in the taxonomies ("malicious code" vs. "abusive-content").
> I'd propose to use either dashes or blanks in all values.

I agree but I won't change existing allowed values in a bug fix release. That's something for a new major release, because that's a change of behavior.

th-certbund:
> I agree but I won't change existing allowed values in a bug fix release. That's something for a new major release, because that's a change of behavior.

Ok, understood. But then we should probably at least use a consistent format for newly added values similar to existing ones (new "vulnerable-system" vs. old "vulnerable service")?
Or do you prefer adding new values in the RSIT format (always dashes) and change the blanks to dashes for existing values later?

wagner-certat:
Preferably I'd like to stay as close to RSIT as possible. We can resolve the inconsistencies in the next major version to get there finally.

pharook:
> @pharook Could you please check the mappings to IDEA? Especially `malware-distribution`, `burglary`, `data-loss`.

Thanks for echo and for mapping new cats to Idea ones. I should also pull myself to get more inline with RSIT.
In Idea we allow usage of main categories only, so for now I would run with:

        "data-loss": "Information",
        "burglary": "Intrusion",
        "malware-distribution": "Malware",

and also into type_to_source_type I would add:

        "malware-distribution": "Malware"

(that goes into Idea's Source.Type, which bears different semantics)

The rest you have mapped look fine to me.

wagner-certat:
Thanks, I added your suggestions.

wagner-certat:
ping @aaronkaplan

PR #1353, opened 2018-11-15
Add support for Shadowserver Darknet

th-certbund:


intelmq/bots/parsers/shadowserver/config.py
```diff
@@ -1834,3 +1835,38 @@ def convert_date(value):
         'protocol.application': 'rsync',
     },
 }
+
+# https://www.shadowserver.org/wiki/pmwiki.php/Services/Darknet
+darknet = {
+    'required_fields': [
+        ('time.source', 'timestamp', add_UTC_to_timestamp),
+        ('source.ip', 'ip'),
+        ('source.port', 'port'),
```
wagner-certat: That's always empty in our reports. I will move it to the optional section

PR #1358, opened 2018-12-03
Fixed typo in defaults.conf key

ondj:
This value is used here without this typo:
https://github.com/certtools/intelmq/blob/develop/intelmq/bin/intelmqctl.py#L469


wagner-certat:
Thanks!

wagner-certat:
Also cherry-picked for the maintenance branch

PR #1361, opened 2018-12-13
New anyrun collector and parser

dargen3:
There is new collector and parser of data from https://any.run/report/ . It is quite non-standart data flow because collector download https://any.run/report/, parse urls, download urls and send them to parser.


wagner-certat:
Any particular reason for revoking the pull request?

PR #1366, opened 2019-01-02
collector http time-delta ability

e3rd:
Generic URL fetcher used local time only. But my feed generates always generates yesterday's data. So I needed to fetch yesterday.

`http_url_formatting` previously could be `true` to turn on time formatting. Now, it can contain a JSON that is passed to timedelta function so that one can easily do the relative date shift.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,9 @@ CHANGELOG
 - added `intelmq.bots.parsers.opendxl.collector` (#1265).
 - added `intelmq.bots.collectors.api`: collecting data using an HTTP API (#123, #1187).
 - added `intelmq.bots.collectors.rsync` (#1286).
-- `intelmq.bots.collectors.http.collector_http`: Add support for uncompressing of gzipped-files (#1270).
+- `intelmq.bots.collectors.http.collector_http`:
+  - Add support for uncompressing of gzipped-files (#1270).
+  - Add time-delta support for time formatted URLs.
 - `intelmq.collectors.blueliv.collector_crimeserver`: Allow setting the API URL by parameter (#1336).
 
 #### Parsers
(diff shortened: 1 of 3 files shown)


e3rd:
ping :) I hope this is not such a big change (no backwards incompatibility) and that would really helped :)

intelmq/bots/collectors/http/collector_http.py
```diff
@@ -30,8 +31,14 @@
 
 
 class Time(object):
+    def __init__(self, delta=None):
+        """ Delta is a datetime.timedelta JSON string, ex: '{days=-1}'. """
```
wagner-certat: Here it's not a JSON string, but a dictionary `{"days": -1}`

CHANGELOG.md
```diff
@@ -21,7 +21,9 @@ CHANGELOG
 - added `intelmq.bots.parsers.opendxl.collector` (#1265).
 - added `intelmq.bots.collectors.api`: collecting data using an HTTP API (#123, #1187).
 - added `intelmq.bots.collectors.rsync` (#1286).
-- `intelmq.bots.collectors.http.collector_http`: Add support for uncompressing of gzipped-files (#1270).
+- `intelmq.bots.collectors.http.collector_http`:
+  - Add support for uncompressing of gzipped-files (#1270).
+  - Add time-delta support for time formatted URLs.
```
wagner-certat: Now the PR# is known^^

Thanks for thinking of the changelog :)

e3rd:
Sorry, I was trying to understand your thoughts several times but I didn't succeed.
By non-JSON formatting you mean plain value for `http_url_formatting` parameters? JSON variant can be extended anyhow.
I'd be glad if you can provide me an example of a more generic variant so that I may implement it.

wagner-certat:
>  By non-JSON formatting you mean plain value for `http_url_formatting` parameters? JSON variant can be extended anyhow.

Yeah. But if we do extend it and `day` is no longer unambiguous, we can still add backwards-compatibility.

PR #1367, opened 2019-01-07
Mail url collector fix

tomas321:
The Shadowserver mail body was changed to bytes object, which could not be parsed.


wagner-certat:
Did you only want to only submit 406f7aa4de4d1ca12cfe3372710224cc9eb85346?

tomas321:
Yes, only the change in process function

wagner-certat:
I also adapted the patch to the maintenance branch

I cherry-picked the one commit for develop.

Thanks for the patch

PR #1368, opened 2019-01-07
Extend shadowserver parser

tomas321:
Add HTTP-scanners, ICS-Scanners and Amplification-DDoS-Victim parsers


tomas321:
Hello Sebastian,
is it possible to merge my pull request to intelmq/develop?

intelmq/bots/parsers/shadowserver/config.py
```diff
@@ -1970,3 +1973,113 @@ def convert_date(value):
...
+        ('extra.', 'pattern', validate_to_none),
+        ('extra.', 'url', validate_to_none),
+        ('extra.file.md5', 'file_md5', validate_to_none),
+        ('extra.file.sha256', 'file_sha256', validate_to_none),
+        ('extra.', 'request_raw', validate_to_none),
+    ],
+    'constant_fields': {
+        'classification.taxonomy': 'information gathering',
+        'classification.type': 'scanner',
+        'classification.identifier': 'http-scanners',
```
wagner-certat: I'd only use the protocol `http` here. It is already clear from the type that this is a scanner. That would be also in line with parser dealing with similar data. And we can also add `protocol.application` and `protocol.transport`

intelmq/bots/parsers/shadowserver/config.py
```diff
@@ -1970,3 +1973,113 @@ def convert_date(value):
...
+        ('extra.', 'state', validate_to_none),
+        ('extra.', 'slave_id', validate_to_none),
+        ('extra.', 'function_code', convert_int),
+        ('extra.', 'request', validate_to_none),
+        ('extra.', 'response', convert_int),
+    ],
+    'constant_fields': {
+        'classification.taxonomy': 'information gathering',
+        'classification.type': 'scanner',
+        'classification.identifier': 'ics-scanners',
```
wagner-certat: I'd actually use the protocol here again. Or is it not available for all rows?

tomas321:
Although the changes would correspond to other feeds with similar data, i think the 'classification.identifier' identifies the feed and not the protocol.

wagner-certat:
> Although the changes would correspond to other feeds with similar data, i think the 'classification.identifier' identifies the feed and not the protocol.

For the feed we already have `feed.name` and `feed.code`. The classifications are an attempt to classify the events _independent_ of the feed. So multiple feeds containing the same / similar data can and should have the same classification.

intelmq/bots/parsers/shadowserver/config.py
th-certbund: I agree 'http' should be sufficient for the identifier.

th-certbund: Btw, in the Reference Security Incident Classification Taxonomy the corresponding type is "scanning". So maybe we should change "scanner" to "scanning" for the data harmonization?

th-certbund: protocol.application should be set to 'http' and protocol.transport to 'tcp' for all events from this feed.

intelmq/bots/parsers/shadowserver/config.py
th-certbund: In the Shadowserver reports, the value of "protocol" usually specifies the _transport_ protocol (tcp or udp). However, in the "ICS Scanners" report, the value of "protocol" reflects the _application_ protocol (e.g. http or s7comm).
I'll ask Shadowserver if the format of the ICS Scanners report can be changed so that protocol specified the transport protocol there as well and an an additional column "application" can be added for the application protocol.

intelmq/bots/parsers/shadowserver/config.py
wagner-certat: > Btw, in the Reference Security Incident Classification Taxonomy the corresponding type is "scanning". So maybe we should change "scanner" to "scanning" for the data harmonization?

No, it's scanner: https://github.com/enisaeu/Reference-Security-Incident-Taxonomy-Task-Force/blob/master/working_copy/machinev1#L53
The `value` is the machine-readable specifier, expanded is for humans.

tomas321:
@wagner-certat I have committed some changes according to your recommendations. Are they sufficient enough or did I miss something?

wagner-certat:
Thanks and sorry for the delay. Too much to do and too less time :/

wagner-certat:
Cherry-picked (as there are some unreleated commits), also into the maintenance branch for the next bugfix release (if there is any) and added a changelog line for this addition (652bdb379a9bb12443ff70eba8d9cc695f81bc8d).

PR #1376, opened 2019-02-11
TCP more stable version

e3rd:
*  socket timeout added
* when used without separator, IntelMQ output waits the other side tells "Ok" so that the output know the message was acknowledged


wagner-certat:
Thanks, pushed to maintenance!

PR #1378, opened 2019-02-13
Added Abuse.ch URLhaus parser

antoinet:
Straighforward parser implementation for abuse.ch's URLhaus feed
(see: https://urlhaus.abuse.ch/feeds/)

URLhaus is a project from abuse.ch with the goal of sharing malicious URLs that are being used for malware distribution.


navtej:
If you use CSV data from URLHause located here https://urlhaus.abuse.ch/downloads/csv/, you don't really need a separate parser. You can parse this with generic CSV parser by using the following parameters

```        "parameters": {
            "skip_header": false,
            "default_url_protocol": "http://",
            "type": "malware",
            "columns": "__IGNORE__,time.source,source.url,status,extra.urlhaus.threat_type,extra.tags,extra.urlhause_link",
            "delimiter": ","
        }
```

Do you think there are some drawbacks of using CSV?

antoinet:
Oh, I wasn't aware of the generic CSV parser! We're operating a very old IntelMQ version which doesn't include it yet. I quickly wrote the URLhaus parser and thought I could share it. At first sight, I don't think there are any drawbacks in using the generic CSV parser for this task. So consider this PR as obsolete...

wagner-certat:
Thanks guys.

Maybe some can add the feed to the feeds documentation then? :)

antoinet:
I'll submit a new PR with the docs, thx.

PR #1379, opened 2019-02-14
Added documentation for abuse.ch URLhaus feed

antoinet:
Following the discussion in https://github.com/certtools/intelmq/pull/1378, here's the documentation for the Abuse.ch URLhaus feed for malicious URLs. Please note that this configuration is targeted at the CSV Feeds provided for specific ASNs, country codes or top level domains as described here: https://urlhaus.abuse.ch/feeds/

The config mentioned by @navtej in https://github.com/certtools/intelmq/pull/1378#issuecomment-463500234 corresponds to the CSV format for the entire URLhaus dump, which is probably not suitable to pull as feed (https://urlhaus.abuse.ch/downloads/csv/).

### DIFF
--- a/docs/Feeds.md
+++ b/docs/Feeds.md
@@ -120,6 +120,31 @@ To add feeds to this file add them to `intelmq/etc/feeds.yaml` and then run
 * **Configuration Parameters:**
 
 
+## URLhaus
+
+* **Status:** on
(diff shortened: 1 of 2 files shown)


antoinet:
Ok, I have totally failed at #RTFM. Will correct ASAP, sorry for the noise.

PR #1382, opened 2019-02-19
Bot.receive_message small optimisation

ondj:
Do not create `tmp_msg` variable when debug logging is not enabled


wagner-certat:
Thanks, looks good.

For all the parser tests in our suite this gives a performance boost of (only) 0.0007 seconds. That's the average of 1000000 runs (each).

PR #1386, opened 2019-03-08
Abusech parser ip fix

tomas321:
Feodotracker produces an ip blocklist feed containing firstseen time, ip, port and malware name, therefore it required adaptation changes. Other Abusech ip feeds are being produces the same way as before. Parsing is now controlled by the main dictionary defining field values. For Feodotracker the "Last updated" time is overwriting time.observation field.


intelmq/bots/parsers/abusech/parser_ip.py
```diff
@@ -6,36 +6,97 @@
...
+        feed = report['feed.url']
+
+        fields = top_line.split(',')
+        if len(fields) is not len(FEEDS[feed]['format']):
+            self.logger.warning("Feed '{}' has not the expected fields: {} != {}".format(feed,
+                                                                                         len(fields),
+                                                                                         len(FEEDS[feed]['format'])))
+            raise PipelineError("Abusech ip parser is not up to date with the format online")
+
+        for line in utils.base64_decode(report.get("raw")).splitlines():
```
wagner-certat: Why don't you use lines here?

intelmq/bots/parsers/abusech/parser_ip.py
```diff
@@ -6,36 +6,97 @@
...
+        regex = re.compile(r'^#+.*')  # filter out comments
+        top_line = list(filter(lambda i: not regex.search(i), lines))[0]
+        feed = report['feed.url']
+
+        fields = top_line.split(',')
+        if len(fields) is not len(FEEDS[feed]['format']):
+            self.logger.warning("Feed '{}' has not the expected fields: {} != {}".format(feed,
+                                                                                         len(fields),
+                                                                                         len(FEEDS[feed]['format'])))
+            raise PipelineError("Abusech ip parser is not up to date with the format online")
```
wagner-certat: Please use a generic ValueError here, PipelineError is for problems with the pipeline (infrastructure) only

intelmq/bots/parsers/abusech/parser_ip.py
```diff
@@ -6,36 +6,97 @@
...
+
+        for line in utils.base64_decode(report.get("raw")).splitlines():
+            line = line.strip()
+            if not any([line.startswith(prefix) for prefix in self.ignore_lines_starting]):
```
wagner-certat: `ignore_lines_starting` is an empty list for this bot, so I don't see a reason to check for it

intelmq/bots/parsers/abusech/parser_ip.py
```diff
@@ -6,36 +6,97 @@
...
+    }
+}
 
 
 class AbusechIPParserBot(ParserBot):
     lastgenerated = None
 
+    def parse(self, report: dict):
+        lines = utils.base64_decode(report.get("raw")).splitlines()
+        regex = re.compile(r'^#+.*')  # filter out comments
```
wagner-certat: Can be moved one level up to the class (doesn't need to be re-compiled on every run)

intelmq/bots/parsers/abusech/parser_ip.py
```diff
@@ -6,36 +6,97 @@
...
+            self.__process_fields(event, line, report['feed.url'])
             yield event
 
+    def __process_defaults(self, event, line, feed_url):
+        defaults = set([
+            ('malware.name', FEEDS[feed_url]['malware']),
+            ('raw', line),
+            ('classification.type', 'c&c'),
+            ('classification.taxonomy', 'malicious code'),
+            ('time.observation', self.lastgenerated)
```
wagner-certat: Please remove this one. The observation time is the time when the data has been fetched

intelmq/bots/parsers/abusech/parser_ip.py
Teebor-Choka: @wagner-certat Is the inability to parse the online format not a pipeline error, when it breaks the bot's intended behavior?

intelmq/bots/parsers/abusech/parser_ip.py
tomas321: It was not intended to break any of the previous functionality, so its simply a copy of the super class parse method. Since this parser was not overriding the parse method.

intelmq/bots/parsers/abusech/parser_ip.py
tomas321: Thank you, I will include this change in the next commit.

intelmq/bots/parsers/abusech/parser_ip.py
tomas321: Thank you, I will change that.

intelmq/bots/parsers/abusech/parser_ip.py
Teebor-Choka: Should the `time.observation` not reflect the changes within IntelMQ along with all supplied knowledge? When the time of last feed refresh is known, that is the last observable time in the parser's knowledge.

The same logic was applied in [this](https://github.com/certtools/intelmq/blob/fb5eba287ddd4234a3a24eba72d025dfd03079af/intelmq/bots/parsers/cert_eu/parser_csv.py#L46) bot.

If the `time.observation` field modeled the physical observation, repeated reads would produce unique messages, throwing away a piece of important information. For the sake of generic behavior, when you imagine a feed consisting of e.g. only IP addresses and the first line comment that says: `Last updated 2001-02-07`, would it be acceptable to have `time.observation` to state the (current) event read time, discarding the fact that it is an 18 years old information?

intelmq/bots/parsers/abusech/parser_ip.py
tomas321: @wagner-certat In my opinion it breaks the pipeline, since it should not process further feeds with current configuration including the feed URL. Just as @Teebor-Choka pointed out.

tomas321: Ok, I will change it.

intelmq/bots/parsers/abusech/parser_ip.py
wagner-certat: > When the time of last feed refresh is known, that is the last observable time in the parser's knowledge.

Yeah, but that's not what `time.observation` is for. I now discussed with @aaronkaplan what the intention and use case of the field is and this is the result (which I will also add to the documentation and fix in the cert-eu parser):

We save two timestamps, one is the time of the event itself. The time, when the incident happened, or the port was open, the connection has been made, etc. This is reported by the source, thus `time.source`.
The other useful information is the time, when the event was first processed by the local intelmq instance, the "observation" of the event. As an example, we at cert.at use this time often to search to events which have been processed e.g. today and to enhance the query speed on the database (when searching for certain events where you now that they are from today).
Speaking generally again: What is the advantage of knowing at what time the feed has been last updated by the feed provider? In case of abuse.ch this time is the most recent "firstseen" (time.source) value. I actually don't see a reason why I do need to attach this information to every event of the feed. The `time.source` field already has the time of the actual event (which is the more important one here).

For plain list of IPs with a single "last updated" information: That's the case for blocklists and we do have documentation for a few of them in intelmq too. In that case we know "The IP [source.ip] was on this list as of [time.source/lastgenerated]". An example is the (current) abusech ip parser:
https://github.com/certtools/intelmq/blob/0ba0a8e/intelmq/bots/parsers/abusech/parser_ip.py

aaronkaplan: yes, and if it's really necessary to add a "last_updated" for a specific feed we could add that to extra. Only when that field is so common (like it would be needed for all feeds), then we should have it standardized. 

But yes, standardizing the internal format (DHO) is non trivial.

wagner-certat: Resulting commits:
33a18883264ecc739716c494dd7917e9f4533234
7f484c3dc

wagner-certat:
I asked the feed provider how the times provided in the feed need to be interpreted

wagner-certat:
The blocklist does not contain any information about the current status of the entries (as opposed to the "Browse" page of the feodotracker).

wagner-certat:
Cherry-picked the relevant commits into develop and maintenance. I kept the last generated information in extra.

Follow-up commit: 60212d7416bdd38dcea4b45f1b23cc6fb1466018

PR #1392, opened 2019-03-27
maxmind crashed due to a typo

e3rd:


wagner-certat:
Thanks, now experienced the same bug :D

PR #1393, opened 2019-03-28
Ignore any lines starting with blank characters in netlab 360 parser

tomas321:
In our intelmq instance the parser was raising IndexErrors for DGA feed. The empty blank line separating comments from data could include multiple blank characters and therefore raising an IndexError. This was caused by checking the length of string and not stripping it from blank character prefixes and suffixes.

We couldn't target the main cause yet, but this could affect the parsing process.


wagner-certat:
Thanks, follow-up commit is 89bc32762

PR #1395, opened 2019-03-29
Remove keyword 'error' from log message associated with non ERROR state

tomas321:
Mentioning keyword 'error' in this way presents problems when greping for real error messages


wagner-certat:
Agreed, thanks
adapted the tests in 448ed3436 (see build fail)

PR #1396, opened 2019-03-29
Dataplane parser refactoring

tomas321:
Introduce feed field count checking before list elements access.

We have experienced IndexErrors with this parser, but couldn't locate the source of the problem regarding feed formats. These changes should be sufficient enough for tracing back the source and deals with IndexErrors.

### DIFF
--- a/intelmq/bots/parsers/dataplane/parser.py
+++ b/intelmq/bots/parsers/dataplane/parser.py
@@ -50,23 +50,29 @@ class DataplaneParserBot(ParserBot):
     }
 
     def parse_line(self, line, report):
+        file_format = [
+            ('source.asn', lambda x: x if x != 'NA' else None),
+            ('source.as_name', lambda x: x.split()[0] if x != 'NA' else None),
+            ('source.ip', lambda x: x),
+            ('time.source', lambda x: x + '+00:00'),
+        ]
+
         if line.startswith('#') or len(line) == 0:
             self.tempdata.append(line)
-
         else:
-            value = line.split('|')
             event = Event(report)
-            event.add('time.source', value[3].strip() + '+00:00')
-            if value[0].strip() != 'NA':
-                event.add('source.asn', value[0].strip())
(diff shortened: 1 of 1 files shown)


intelmq/bots/parsers/dataplane/parser.py
```diff
@@ -50,23 +50,29 @@ class DataplaneParserBot(ParserBot):
     }
 
     def parse_line(self, line, report):
+        file_format = [
```
Teebor-Choka: It occurs to me now that this variable could be extracted to `__init__`  in order to save the variable allocations.

intelmq/bots/parsers/dataplane/parser.py
```diff
@@ -50,23 +50,29 @@ class DataplaneParserBot(ParserBot):
...
+                raise ValueError('Incorrect format for feed {}, found line: "{}"', event.get('feed.url'), line)
+
+            if line_contents[-1].strip() in self.CATEGORY:
+                event.update(self.CATEGORY[line_contents[-1].strip()])
             else:
-                raise ValueError('Unknown data feed %r.' % value[4].strip())
+                raise ValueError('Unknown data feed {}.'.format(line_contents[-1].strip()))
+
+            for field, setter in zip(line_contents, file_format):
+                event.add(setter[0], setter[1](field.strip()))
```
Teebor-Choka: What happens, when `setter[1](field.strip())` returns `None`? Should it be taken care of or does `Event.add` handle that?

Teebor-Choka:
@wagner-certat, should you wish to add some more tests: the error messages in logs were generated by lines like these (example from the modified new bot on 2019-04-01: `'Incorrect format for feed http://dataplane.org/sshclient.txt, found line: "14080        |  Telmex Colombia S.A., CO        |"`

wagner-certat:
Thanks, I added a test case for the broken line.

bddfaed1d

PR #1398, opened 2019-04-05
Shadowserver parser: No fixed 'tag' values in extra, fix identifier for ftp.

th-certbund:
Remove storing fixed 'tag' values in extra. Fix classification.identifier for ftp.


wagner-certat:
Thanks

All fields not explicitly mapped are stored in extra anyway.

Follow up fixes: 4df6c74e7

PR #1399, opened 2019-04-05
Shadowserver DDoS victim parser: Store tag in identifier

th-certbund:
Store tag (= service abused) in classification.identifier instead of fixed value 'amplification-ddos-victim' (which can be obtained from feed.name).

### DIFF
--- a/intelmq/bots/parsers/shadowserver/config.py
+++ b/intelmq/bots/parsers/shadowserver/config.py
@@ -1991,7 +1991,7 @@ amplification_ddos_victim = {
         ('source.geolocation.cc', 'geo'),
         ('source.geolocation.region', 'region'),
         ('source.geolocation.city', 'city'),
-        ('extra.', 'tag', validate_to_none),
+        ('classification.identifier', 'tag'),  # service abused for the attack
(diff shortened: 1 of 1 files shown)


wagner-certat:
In the reports we received today, tag is always empty. The example from the tests has the tag "daytime", that would work. The samples on [shadowserver's page](https://www.shadowserver.org/wiki/pmwiki.php/Services/Amplification-DDoS-Victim) also miss the tag for some of the rows.

It may be possible to use the tag with a sane default.

bernhardreiter:
@wagner-certat what kind of feedback is needed? (Or should the code be adjusted to give a sane default to tag?)

wagner-certat:
> @wagner-certat what kind of feedback is needed?

With the data available to me (AT) I do no see a value in this change. Without further fallback, the field would be empty. The feedback I would need is: Is there any valuable data in this tag column in some other's dataset?
> (Or should the code be adjusted to give a sane default to tag?)

Would be an option, but that's currently the case.

wagner-certat:
Closing because of inactivity. Please reopen if needed.

PR #1400, opened 2019-04-08
Add new FeodoTracker IP blocklist field to parser

tomas321:
FeodoTracker has added a new field [`LastOnline`](https://feodotracker.abuse.ch/downloads/ipblocklist.csv). I've added it to the extra field.


wagner-certat:
> FeodoTracker has added a new field [`LastOnline`](https://feodotracker.abuse.ch/downloads/ipblocklist.csv). I've added it to the extra field.

I triggered this change.

The last online field is actually more the `time.source` than the first seen. And the first column should be `extra.firstseen`.

tomas321:
It could be a better choice if the `LastOnline` fields would be always present. Perhaps `FirstSeen` should should be recognized as `time.source` only if `LastOnline` is not present.

wagner-certat:
LastOnline exists now for most of the entries. I now asked if there's also a time information possible.

> It could be a better choice if the `LastOnline` fields would be always present. Perhaps `FirstSeen` should should be recognized as `time.source` only if `LastOnline` is not present.

+1

wagner-certat:
> LastOnline exists now for most of the entries. I now asked if there's also a time information possible.

I guess it is not possible to have a time information :/

wagner-certat:
I got the confirmation from abuse.ch that there's no time information, but it is UTC, so I now appended 'T00:00+00' to this field to save it in `extra.last_online` and use if as `time.source` if available.

PR #1401, opened 2019-04-08
Enrich failed to parse line error message with feed_url and line

tomas321:
The `Failed to parse line.` error message is very frequent and missing more direct information about the specific line and feed.

### DIFF
--- a/intelmq/lib/bot.py
+++ b/intelmq/lib/bot.py
@@ -813,7 +813,8 @@ class ParserBot(Bot):
                 else:
                     events = [value]
             except Exception:
-                self.logger.exception("Failed to parse line '{}' for feed '{}'".format(line, report.get('feed.url', 'Feed URL is unspecified')))
+                self.logger.exception("Failed to parse line '{}' for feed '{}'".format(
+                                      line, report.get('feed.url', 'Feed URL is unspecified')))
                 self.__failed.append((traceback.format_exc(), line))
             else:
                 events_count += len(events)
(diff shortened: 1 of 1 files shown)


intelmq/lib/bot.py
```diff
@@ -813,7 +813,7 @@ def process(self):
                 else:
                     events = [value]
             except Exception:
-                self.logger.exception('Failed to parse line.')
+                self.logger.exception("Failed to parse line '{}' for feed '{}'".format(line, report['feed.url']))
```
wagner-certat: You can't rely on the existence of `feed.url`.

tomas321: Thanks for that. I fixed the issue in the new commit

wagner-certat:
I'm not sure if I like that because then we have actual data (only for DEBUGging) in the ERROR log level.

intelmq/lib/bot.py
```diff
@@ -813,7 +813,8 @@ def process(self):
                 else:
                     events = [value]
             except Exception:
-                self.logger.exception('Failed to parse line.')
+                self.logger.exception("Failed to parse line '{}' for feed '{}'".format(
+                                      line, report.get('feed.url', 'Feed URL is unspecified')))
```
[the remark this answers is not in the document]
tomas321: Sorry for the longer wait on my reply. The more verbose log line was useful when we had many shadowserver feeds with repetitive errors of this type - it speeded up the diagnosis process. Although, I was not aware of the intelmqdump tool addressing this.

wagner-certat:
Okay, makes sense. I'll close there then

PR #1406, opened 2019-05-09
Elasticsearch SSL Support

stone-z:
Adds SSL for connection to Elasticsearch. Includes:
- new parameters for using + verifying SSL, optional path to CA cert bundle, option to suppress warnings
- docs

Client authentication is not implemented -- I left placeholders in the ES client code but don't have a setup right now using client certs to test the behavior. It may just need two more parameters added to the bot.

Related to #1150

### DIFF
--- a/intelmq/bots/outputs/elasticsearch/output.py
+++ b/intelmq/bots/outputs/elasticsearch/output.py
@@ -61,10 +61,12 @@ class ElasticsearchOutputBot(Bot):
                                      'elastic_index', 'intelmq')
         self.rotate_index = getattr(self.parameters,
                                     'rotate_index', False)
-        self.http_username = getattr(self.parameters,
-                                     'http_username', None)
-        self.http_password = getattr(self.parameters,
-                                     'http_password', None)
+        self.use_ssl = getattr(self.parameters,
+                               'use_ssl', False)
+        self.ssl_ca_certificate = getattr(self.parameters,
+                                          'ssl_ca_certificate', None)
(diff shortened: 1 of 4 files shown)


navtej:
The CI failures are because of RIPEBot failing to connect at some point.

intelmq/bots/outputs/elasticsearch/output.py
```diff
@@ -74,11 +76,17 @@ def init(self):
...
-        self.es = Elasticsearch([{'host': self.elastic_host, 'port': self.elastic_port}], **kwargs)
+        self.set_request_parameters()  # Not all parameters set here are used by the ES client
+
+        self.es = Elasticsearch([{'host': self.elastic_host, 'port': self.elastic_port}],
+                                http_auth=self.auth,
+                                use_ssl=self.use_ssl,
+                                verify_certs=self.http_verify_cert,
+                                ca_certs=self.ssl_ca_certificate,
+                                ssl_show_warn=self.ssl_show_warnings,
+                                # client_cert=self.ssl_client_cert,
```
wagner-certat: What is this comment for?

stone-z: Just suggestions. The ES client supports TLS client auth behind the scenes using those parameter names. It may be that un-commenting those lines and passing the parameters to the bot (ssl_client_cert and key) will work as-is, I just don't have a way to try myself it at the moment. If I get the chance to set up an Elasticsearch node supporting client auth I will revisit it personally, otherwise if someone has a setup already they can un-comment and try it out.

Issue #1417, opened 2019-06-12
AbuseCH IP Parser Bot: Incorrect handling of CSV header


wagner-certat:
For the bot AbusechIPParserBot, 3 feeds are handled.
'https://feodotracker.abuse.ch/downloads/ipblocklist.csv'
'https://zeustracker.abuse.ch/blocklist.php?download=ipblocklist'
'https://zeustracker.abuse.ch/blocklist.php?download=badips'

For zeustracker the header of the feed are the same.
For feodo, the header is different and you need to change this line.

fields = comments[-1].split(',')
to
fields = comments[-2].split(',')

But doing that break the other feed.
So I add a parameter call position and change the line as it:
fields = comments[-FEEDS[feed]['position']].split(',')

_Originally posted by @kalyparker in https://github.com/certtools/intelmq/issues/1415#issuecomment-501328031_

wagner-certat:
Thanks, I understand the problem.
https://github.com/certtools/intelmq/blob/develop/intelmq/bots/parsers/abusech/parser_ip.py#L58

PRs welcome :)

CC @tomas321

navtej:
Why not just use csv parser and retire these parsers?

kalyparker:
> Why not just use csv parser and retire these parsers?

I'm quite agree :)
I'll just fix this bug now because actually it is the "spirit" of intelmq, but yes I don't believe this approach is the better one.
Today I used more than 20 differents feeds, for those offering a parser bot I used it because I believe it will be easier, but I need to change fields after the parsing because it did not fit my needs (source.ip to destination.ip for example or hard-coded value which I don't want). 
Actually I work on a new bot for another feed because the generic csv does not handle properly the parsing when in the column protected by a "" you have the same separator. I probably look at it this week end, instead of creating a new one, maybe it will be smarter ^^
If I need to do something, I believe a generic parser should be enough for the first step - if there is one parser for the different possible type of entry (csv, json...) - 
In the generic parser, for column which fit directly the harmonization then fill it directly. for the other, having a kind of temporary variable (string) should be great !
After this first parsing, use a dedicated bot for processing temporary variable when needed. For example, I have some feed where the ip column is more ip:port and this is impossible to manage directly and the actual sieve and modify are often not enough or too limited in my opinion. I usually need both :(
Let's talk about it in a new thread maybe ?

Anyway, I push this change before the next monday.

wagner-certat:
[...]

PR #1418, opened 2019-06-16
fixes abusech csv certtools/intelmq#1417

kalyparker:


wagner-certat:
Thanks, fixed tests and added docs in fbbc3478c0e0d7293eafafc39d344e23e231b853

PR #1419, opened 2019-06-26
New ViriBack parser

dargen3:
There is new ViriBack parser. These parser takes data from http collector and it takes it from website http://tracker.viriback.com/. Urls are published without protocol so "http://" is added by parser.


dargen3:
I don`t understand the reason of bugs. When i run test locally, ViriBack test correctly run.

dargen3:
Hello, 
there is some bug in test_db_ipv4_lookup (intelmq.tests.bots.experts.ripe.test_expert.TestRIPEExpertBot), but i made no changes with these expert. So i don`t know, what is wrong. Could you tell me, please, what i should do? Or it is internal bug?

wagner-certat:
Did you try to use the generic html parser?

`raw_report.splitlines()[78].strip()[8:-10]` looks very prone to minimal changes on the provider's side.

dargen3:
Thank you for the remark. Now it use html parser.

wagner-certat:
It was actually a question.

We have an existing parser for feeds like this one: https://github.com/certtools/intelmq/blob/develop/docs/Bots.md#html-table-parser
I fiddled a bit around and it can be used for that. I however extended it in two ways: Parsing dates and converting them to date time, like we did in #1400 , seems like you were facing the same issue and decided to ignore it (I now provided such functionality - generically - in the core)
Second the html table parser now supports multiple parsers (in the beautifulsoup context), because the viriback website has incorrect html and the `html.parser` can't handle that properly.

I reused your tests, the mapping and the docs of course. I hope you are not disappointed that I replaced your proposed parser.

cc @kbrajneesh I also moved the list of available time conversions to the core (and added a few). Still lot of room for improvement, if you have some time :)

Full diff now with all my follow-up commits (I ended up doing a bit more work :D ): https://github.com/certtools/intelmq/compare/ea8b0926c5c61c8d7192fd4c14a0d90ac4bc8028...010c22c9bd779077959b81e1d797120a2dfb9ed6

dargen3:
No, i am really not disappointed. Solution with table-parser is better. Thank you for help and improvement! :)

PR #1448, opened 2019-09-06
SQLBot

e3rd:
**OUTDATED:** Using SQL connection in a bot can be now more generic. PostgreSQL connection, execution and exception handling code was duplicated. Now, I've pulled it to the PostgreSQLBot that can be inherited instead and that inherits SQLBot.
Like that, we may easily adopt for example SQLite engine which is heavily used in CSIRTMalta workflow.

Could you take look at it please and tell me what design fails you see or that it's alright and I may develop further and do some SQLite bots?

**UPDATE:** This PR is a draft (just take a look at the code, tests fails due to not be edited yet).
* I made a class SQLBot in `lib/bot.py` that handles connection to relational databases – PostgreSQL and SQLite (heavily used in CSIRTMalta workflow). Now, I had two options. Either make a separate expert and output for every relational database (so that user can add "PostgreSQLOutput", "PostgreSQLLookupExpert", "SQLiteOutput", "SQLiteLookupExpert") or make a generic expert and generic output. I've chosen the latter.
* Any bot using SQLBot will need to have an extra parameter "engine" with the value "postgresql" or "sqlite".
* Generic DB Lookup is now generic
* PostgreSQL output is now generic and should be IMHO renamed to SQLOutput

What do you think? That way, we may easily adopt any DB our users need.


wagner-certat:
> What do you think? That way, we may easily adopt any DB our users need.

Cool work, definitely appreciated :)

e3rd:
Thanks, I'll do some changes and let you know :)

intelmq/bots/BOTS
```diff
@@ -1005,6 +1006,9 @@
         "PostgreSQL": {
             "description": "PostgreSQL is the bot responsible to send events to a PostgreSQL Database. When activating autocommit, transactions are not used: http://initd.org/psycopg/docs/connection.html#connection.autocommit",
             "module": "intelmq.bots.outputs.postgresql.output",
+            "help": {
```
wagner-certat: I'm not sure if we really want to have the help for parameters in that file. (#757 proposes some major steps in how parameters are handled including types and docs/help.)

e3rd: Yeah, sorry, that's definitely not part of this PR, I've just wondered how to implement help description to the Manager and been designing some ways... Then I discovered #757 and other and saw that nothing is decided yet and gave up :D

Issue #1442, opened 2019-09-03
Shadowserver parser: determine feed by report's extra fields


wagner-certat:
HTTP, RT and Mail collectors provide the file name in a field (`extra.file_name`). This can be used to determine the feedname and thus would obsolete having lots of different shadowserver parsers with different feedname parameters.

grodriguezl:
I'm working on this. I will assume the following:
1.   report's field `extra.file_name` is not modified (the same as if I download the file manually), so it could be something like `2019-09-05-botnet_drone-$suffix.csv`. The important value here is `botnet_drone` which is after the date and before any other '-'

Parse function will determine which config has to be used to process that report.

Any comments or ideas are appreciated

e3rd:
Thanks!

These are the file name → feed name mapping I've been able to guess. I wonder if there is a complete list somewhere.

```python
feeds = ["Accessible-HTTP",
"Accessible-ADB",
"Accessible-AFP",
"Accessible-Cisco-Smart-Install",
"Accessible-CWMP",
"Accessible-FTP",
"Accessible-RDP",
"Accessible-Rsync",
"Accessible-SMB",
"Accessible-Telnet",
"Accessible-Ubiquiti-Discovery-Service",
"Accessible-VNC",
"Amplification-DDoS-Victim",
"DNS-Open-Resolvers",
"Drone-Brute-Force",
"Drone",
"NTP-Monitor",
"NTP-Version",
"Open-Chargen",
"Open-IPMI",
"Open-LDAP",
"Open-mDNS",
"Open-MongoDB",
"Open-MSSQL",
"Open-NATPMP",
"Open-NetBIOS-Nameservice",
"Open-Portmapper",
"Open-QOTD",
"Open-Redis",
"Open-SNMP",
"Open-SSDP",
"Open-TFTP",
"Sandbox-URL",
"Sinkhole-HTTP-Drone",
"SSL-FREAK-Vulnerable-Servers",
"SSL-POODLE-Vulnerable-Servers",
"Vulnerable-ISAKMP"]

files= ["scan_http",
"scan_adb",
"scan_afp",
"cisco_smart_install",
"scan_cwmp",
"scan_ftp",
"scan_rdp",
"scan_rsync",
"scan_smb",
"scan_telnet",
"scan_ubiquiti",
"scan_vnc",
"ddos_amplification",
"scan_dns",
"drone_brute_force",
"botnet_drone",
"scan_ntpmonitor",
"scan_ntp",
[...]

PR #1450, opened 2019-09-11
Update Shadowserver parser

grodriguezl:
This PR is based on https://github.com/certtools/intelmq/issues/1442

I modified the bot so now it uses `extra.file_name` field that collectors add to the report. 

Main changes: 
* now only one Parser is needed instead of one parser per type of report
* report's field  `extra.file_name` MUST be present
* parameters `feedname` and `overwrite` are no longer needed
* new parameter `keep_feedname` is added to check if feedname should be overwritten
* Depending on `extra.file_name` the corresponding configuration is loaded
* Re-arrange of tests, almost all supported type of reports are tested
* I changed the values of `feed.name`, most changes are '-' for spaces 

Supported reports are in bot's README

Besides updated tests, I tested it with reports we receive and a File Collector bot.

TODOS:
* I don't have the filenames of these reports https://github.com/certtools/intelmq/issues/1442#issuecomment-529577029
* I don't have samples of `cwsandbox_url` so it is not tested.

Due to the changes I made, **it is not** back compatible with older versions of the bot.

Please have a look


e3rd:
It seems when running this awkward command, a test failed.

It seems like something changed for CYMRU bot.

`INTELMQ_TEST_DATABASES=1 INTELMQ_TEST_LOCAL_WEB=1 INTELMQ_TEST_EXOTIC=1 nosetests --with-coverage --cover-package=intelmq --cover-branches; find contrib/ -name "test*.py" -exec nosetests {} \+; elif [[ $requirements == false ]]; then INTELMQ_TEST_LOCAL_WEB=1 nosetests --with-coverage --cover-package=intelmq --cover-branches;`

Maybe some change of the testing message, seems like a small error

```
'protocol.transport': 'icmp',
?                         ^ -
+  'protocol.transport': 'tcp',
```

Note you can examine by just running the single test, ex: `python3 test_cap_program_new.py`. Note that you can amend a commit (so that a new commit doesn't appear) and force push the branch. (Sorry if you know all this, I just saw "first time contributor" in the top bar.)


130 files changed, wow! :o

grodriguezl:
I think that error was introduced in 39179824eedc88ef1ebea3f0fe1d177e67d29453 and fixed later in dc81d8fe581144942b9e7a020f653293e7b70d17

wagner-certat:
Just removing the previously provided parameters is something we cannot do. That breaks the compatibility with previous versions.

I will need some time to review the code and recover the previous functionalities.

wagner-certat:
Thank you @grodriguezl for all that work! I saw that you must have invested lots of time and cleaned up quite some things and also added new tests etc.

Based on that I made two new commits 5e23101c1135c401ece2ebdcc5f8881130b71d72 and 78b22a3a1c270312b6410237c0c1078a4968117f -  the latter one for the new `netis_router` tests

PR #1451, opened 2019-09-13
Improved parser to better handle description

amojamo:
- Changed _value_ to _values_ as it makes more sense.
- Use of regex to read the description for better robustness. As it stands now, there is a conflict when the Bambenek parser reads the IP list. This is because there is a slight change in the Bambenek IP list, where they have a longer description with more commas than usual.

### DIFF
--- a/intelmq/bots/parsers/bambenek/parser.py
+++ b/intelmq/bots/parsers/bambenek/parser.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 """ IntelMQ parser for Bambenek DGA, Domain, and IP feeds """
 
+import re
 from intelmq.lib.bot import ParserBot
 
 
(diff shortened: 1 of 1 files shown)


intelmq/bots/parsers/bambenek/parser.py
```diff
@@ -32,32 +33,33 @@ def parse_line(self, line, report):
             self.tempdata.append(line)
 
         else:
-            value = line.split(',')
+            m = re.match(r"(?P<ip>\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}),(?P<description>.*), \
```
wagner-certat: Does not seem to work with IPv6

intelmq/bots/parsers/bambenek/parser.py
```diff
@@ -32,32 +33,33 @@ def parse_line(self, line, report):
...
+                (?P<timestamp>\d{4}-\d{2}-\d{2}[ ]\d{2}[:]\d{2}),(?P<url>.*)", line)
+            values = m.groups()
```
wagner-certat: That line raises an exception in the tests

intelmq/bots/parsers/bambenek/parser.py
wagner-certat: Could we change `.*` (greedy) to `.*?` (non-greedy) here?

intelmq/bots/parsers/bambenek/parser.py
```diff
@@ -32,32 +33,33 @@ def parse_line(self, line, report):
             self.tempdata.append(line)
 
         else:
-            value = line.split(',')
+            m = re.match(r"(?P<ip>\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}),(?P<description>.*), \
+                (?P<timestamp>\d{4}-\d{2}-\d{2}[ ]\d{2}[:]\d{2}),(?P<url>.*)", line)
```
wagner-certat: Could we change `.*` (greedy) to `.*?` (non-greedy) here?

Can we `$` at the end to indicate the regular expression should match the full line?

amojamo:
> What was the actual problem you were facing? E.g. a line which could not be parsed?

They (Bambenek) have since last time edited the IP list, so it parses the line without error as of today. The problem before their last edit was on the following line(s):

> 64.183.187.20,IP resolved by necurs C&C, uses encoded IP, this is not the C2 IP, 2019-09-17 06:06,http://osint.bambenekconsulting.com/manual/necurs.txt

This has since been changed to the following:
> 64.183.187.20,IP resolved by necurs C&C uses encoded IP - this is not the C2 IP, 2019-09-17 06:06,http://osint.bambenekconsulting.com/manual/necurs.txt

The problem was that the parsing failed because there were more commas than anticipated, so `event.add('event_description.url', value[3])` contained the test "this is not the C2 IP" instead of the anticipated URL.

This kinda makes the change obsolete in a way, but without a regex expression the parser is more fragile than it needs to be.

intelmq/bots/parsers/bambenek/parser.py
amojamo: Ah, I'm sorry, I didn't consider IPv6. Regex for an IPv6 pattern is out of my scope. 
The greedy to non-greedy works for the description group, but not for the URL group.

intelmq/bots/parsers/bambenek/parser.py
amojamo: It could be my line split. I'm not sure how to break the regex line into two lines in order not to trigger the "Line too long" warning when it comes to code style checking.

wagner-certat: Just split the line like this:
```python
            m = re.match(r"(?P<ip>\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}),(?P<description>.*), "
                         r"(?P<timestamp>\d{4}-\d{2}-\d{2}[ ]\d{2}[:]\d{2}),(?P<url>.*)", line)
```

wagner-certat:
> > What was the actual problem you were facing? E.g. a line which could not be parsed?
> 
> They (Bambenek) have since last time edited the IP list, so it parses the line without error as of today. The problem before their last edit was on the following line(s):
> 
> > 64.183.187.20,IP resolved by necurs C&C, uses encoded IP, this is not the C2 IP, 2019-09-17 06:06,http://osint.bambenekconsulting.com/manual/necurs.txt
> 
> This has since been changed to the following:
> 
> > 64.183.187.20,IP resolved by necurs C&C uses encoded IP - this is not the C2 IP, 2019-09-17 06:06,http://osint.bambenekconsulting.com/manual/necurs.txt
> 
> The problem was that the parsing failed because there were more commas than anticipated, so `event.add('event_description.url', value[3])` contained the test "this is not the C2 IP" instead of the anticipated URL.

Well, that's obvious.

> This kinda makes the change obsolete in a way, but without a regex expression the parser is more fragile than it needs to be.

If the regular expression itself is stable - yes. I opened https://github.com/amojamo/intelmq/pull/1 for some tests which you could use for development.

In case the format is proper CSV (using commas is ok then if properly escaped), we can use the csv parser of python itself like with any csv-based feed. That's IMHO the best option.

wagner-certat:
Are you still working on this? The tests are still failing on `values = m.groups()`

Issue #1452, opened 2019-09-17
Reverse DNS and Cymru Whois default overwrite


grodriguezl:
Reverse DNS and Cymru Whois experts by default overwrite events. 
Is it intended to be like this?

wagner-certat:
Should be configurable - Pull requests welcome :) (With the default behaviour like it is currently)

grodriguezl:
Doing...

PR #1455, opened 2019-09-18
ENH: default overwrite on reverse_dns and cymru_whois

grodriguezl:
Fixes https://github.com/certtools/intelmq/issues/1452

Default overwrite is set to true.


wagner-certat:
Thanks!

Based on this I further added some docs and tests, also a upgrade function and a warning, also for the future change. The default for new bots is now False.
6843167de7355486b15e82b98441e5220c04fe4e

PR #1456, opened 2019-09-24
Set logging_level_stream when defaults are unavailable

d-lord:
Hi! I'm keen to try out IntelMQ but ran into an error straight after installation. Here's a simple fix.

This error occurs when invoking `intelmqctl` on my plain Ubuntu 18.04 LTS install (vagrant, ubuntu/bionic64, dependencies installed as per the install guide):

>UnboundLocalError: local variable 'logging_level_stream' referenced before assignment

The pull request simply sets that to DEBUG, allowing intelmqctl to run. Not exhaustively tested. I tried running the tests in the repo but 26 of them failed, perhaps because of Vagrant? This pull request doesn't change the test results, anyway.

### DIFF
--- a/intelmq/bin/intelmqctl.py
+++ b/intelmq/bin/intelmqctl.py
@@ -651,6 +651,7 @@ class IntelMQController():
             self.load_defaults_configuration()
         except Exception:
             log_level = DEFAULT_LOGGING_LEVEL
+            logging_level_stream = 'DEBUG'
         else:
             log_level = self.parameters.logging_level.upper()
             logging_level_stream = log_level if log_level == 'DEBUG' else 'INFO'


wagner-certat:
Thanks for reporting this with a proposed patch. I applied some further changes based on it.

However it's interesting that you actually ran into this, because that would mean that the loop or a setattr threw an error which I'd consider rather interesting. In c1ad087b4a696e54c4d1fc4f78cc4f1eabce7df6 the error should be given as output, could you maybe share this (only if it is relevant of course).

d-lord:
Thanks for taking a look! I'll investigate tonight and see what error was raised.

d-lord:
Sorry to say that I haven't been able to reproduce this with full confidence. It may have been a permission error while opening `/opt/intelmq/var/log/intelmqctl.log` for writing, but I'm out of time and can't prove it for sure.

That error, if it's the culprit, is definitely a quirk of my setup. I mounted `/opt/intelmq` to be the repo on my host machine via Vagrant so it's less permissive than normal.

Thank you!

wagner-certat:
> Sorry to say that I haven't been able to reproduce this with full confidence. It may have been a permission error while opening `/opt/intelmq/var/log/intelmqctl.log` for writing, but I'm out of time and can't prove it for sure.

That should be handled by https://github.com/certtools/intelmq/blob/e4d7fe964ba4e0d36a59031c17c8f59cfc296cfa/intelmq/bin/intelmqctl.py#L663-L671

> That error, if it's the culprit, is definitely a quirk of my setup and not an issue with IntelMQ. I mounted `/opt/intelmq` to be the repo on my host machine via Vagrant so it's less permissive than normal.

Okay. If there's anything you can contribute to [contrib/vagrant](https://github.com/certtools/intelmq/tree/develop/contrib/vagrant), please let me know, these files were last updated 4 years ago.

PR #1485, opened 2020-01-04
Fix unresolved conflict in user guide

faheel:
The conflict was introduced in commit e7e76cc2a200166e59e7e598f220f7e7a4840806


wagner-certat:
Thanks!

However, I think the correct one is the other, as that commit merged maintenance into develop and the version from develop should win.

PR #1490, opened 2020-01-21
Fix usage description for intelmqctl

bernhardreiter:
* Extend the filename for logging in the same way as lib/utils.py .


wagner-certat:
Thanks!

PR #1510, opened 2020-03-03
hidden bug

e3rd:
Hi, I've found this hidden bug. There was no effect thanks to `harmonisation.py/import dns.resolver`, if you comment this line, the unit test fails.

See:

```python3
import dns
dns.resolver, dns.reversename, dns.exception.DNSException # AttributeError: module 'dns' has no attribute 'resolver'
```

```python3
import dns.resolver
dns.resolver, dns.reversename, dns.exception.DNSException # no problem
```

### DIFF
--- a/intelmq/bots/experts/reverse_dns/expert.py
+++ b/intelmq/bots/experts/reverse_dns/expert.py
@@ -2,7 +2,7 @@
 
 from datetime import datetime
 
-import dns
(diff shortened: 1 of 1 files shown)


wagner-certat:
Thanks.

To be on the safe side (in case of some changes in the library upstream) I now added the other imports too in d6a5aae1

PR #1514, opened 2020-04-14
Adds ELK setup docs.

gethvi:
As I had promised via dev mailing list I wrote a guide for setting up Logstash and Elasticsearch for IntelMQ.

### DIFF
--- a/docs/ELK-Stack.md
+++ b/docs/ELK-Stack.md
@@ -1,0 +1,368 @@
+# ELK Stack
+
+If you wish to run IntelMQ with ELK (Elasticsearch, Logstash, Kibana) it is entirely possible. This guide assumes the reader is familiar with basic configuration of ELK and does not aim to cover using ELK in general. It is based on the version 6.8.0 (ELK is a fast moving train therefore things might change). Assuming you have IntelMQ (and Redis) installation in place, lets dive in.
+
+## Configuring IntelMQ for Logstash
+
+In order to pass IntelMQ events to Logstash we are going to need an output queue, particulary orphaned queue (which simply means there is no IntelMQ bot processing events from this queue). However because IntelMQ system does not like to have an orphaned queue (and the IntelMQ Manager does not allow to create one), we will simply add a new bot at the end of the pipeline (or wherever you see fit) and set it's configuration `enabled` to `false`. The bot type does not matter as it's only a forever disabled dummy bot (it's recommended to choose an output bot, e.g. Rest API Output Bot).
+
+## Configuring Logstash
+
+Logstash defines pipeline as well. In the pipeline configuration of Logstash you need to specify where it should look for IntelMQ events, what to do with them and where to pass them.
+
+### Input
+
+This part describes how to receive data from Redis queue.
+See the example configuration and comments below:
+
+```
+input {
+  redis {
+    host => "10.10.10.10"
+    port => 6379
+    db => 2 
+    data_type => "list"
+    key => "restapi-output-queue"
+  }
+}
+```
+* `host` - Redis hostname or IP
+* `port` - Redis port
+* `db` - Redis db used (same value as `source_pipeline_db` from defaults.conf of IntelMQ)
+* `data_type` - data type is usually `list`
+* `key` - name of the queue (found in pipeline.conf of IntelMQ as source queue for our dummy bot)
+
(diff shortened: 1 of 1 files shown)


docs/ELK-Stack.md
```diff
@@ -0,0 +1,368 @@
+# ELK Stack
+
+If you wish to run IntelMQ with ELK (Elasticsearch, Logstash, Kibana) it is entirely possible. This guide assumes the reader is familiar with basic configuration of ELK and does not aim to cover using ELK in general. It is based on the version 6.8.0 (ELK is a fast moving train therefore things might change). Assuming you have IntelMQ (and Redis) installation in place, lets dive in.
+
+## Configuring IntelMQ for Logstash
+
+In order to pass IntelMQ events to Logstash we are going to need an output queue, particulary orphaned queue (which simply means there is no IntelMQ bot processing events from this queue). However because IntelMQ system does not like to have an orphaned queue (and the IntelMQ Manager does not allow to create one), we will simply add a new bot at the end of the pipeline (or wherever you see fit) and set it's configuration `enabled` to `false`. The bot type does not matter as it's only a forever disabled dummy bot (it's recommended to choose an output bot, e.g. Rest API Output Bot).
```
navtej, line 7: We dont need an orphaned queue. We have a redis output bot, which can be used here. Please see this https://github.com/certtools/intelmq/tree/develop/intelmq/bots/outputs/redis 
Using a redis output bot avoids all configuration warnings/errors and gives you the ability to have a distributed setup where you dont use shared redis.

docs/ELK-Stack.md
```diff
@@ -0,0 +1,368 @@
...
+* `key` - name of the queue (found in pipeline.conf of IntelMQ as source queue for our dummy bot)
+
+**Extra**
+
+You can also use syntax like this: `host => "${REDIS_HOST:10.10.10.10}"`
+
+The value will then be taken from environment variable `$REDIS_HOST`. If the environment variable is not defined then the default value of `10.10.10.10` will be used instead.
+
+### Filter (optional)
+
```
navtej, line 42: we should try to keep all field modifications inside the IntelMQ ecosystem. We have an expert bot for the very purpose, https://github.com/certtools/intelmq/blob/develop/docs/Bots.md#format-field . The reason is that if the user has multiple outputs configured, he can apply changes at one place to have consistent output across the various output bots.

docs/ELK-Stack.md
```diff
@@ -0,0 +1,368 @@
...
+* `index` - name of the index where to insert data
+
+Our experience, hardware equipment and the amount of events collected led us to having a separate index for each month. This might not necessarily suit your needs.
+
+## Configuring Elasticsearch
+
+Configuring Elasticsearch is entirely up to you and should be consulted with the [official documentation](https://www.elastic.co/guide/en/elasticsearch/reference/index.html). What you will most likely need is something called [index template](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates.html) mappings which I can provide. Again it might not suit your needs so feel free to adjust it for your use case, it is only included as an inspiration.
+
+```
+{
```
navtej, line 78: We have an es mapper tool which should be used here.

docs/ELK-Stack.md
gethvi: Thank you, I will look into it, definitely looks like a much cleaner solution.

docs/ELK-Stack.md
gethvi: I don't see the connection here. Format Fields Bot is used for for fields inside IntelMQ events. The note about environment variables is a different situation and concerns only the configuration of Logstash.

docs/ELK-Stack.md
gethvi: This is excellent, thanks for pointing it out, I will update the guide accordingly.

docs/ELK-Stack.md
navtej: The context is
Filter (optional)

i added the comment below that

gethvi: Ok, I didn't understand before. I agree, the filter part was only an example that it can be done, not necessarily that it should be done. I will change it. Thanks.

wagner-certat:
Thank you very much for your contribution!

Issue #1409, opened 2019-05-14
mismatch of our types with upstream RSIT


wagner-certat:
There are a couple of types which are not in RSIT. What to do with them?

 - [ ] information content security > dropzone: asked on RSIT ML for input -> PR #2006 
 - [x] information content security > leak: data-leak https://github.com/enisaeu/Reference-Security-Incident-Taxonomy-Task-Force/pull/61
 - [X] intrusions > backdoor: merged into system-compromise in PR #1995
 - [x] intrusions > compromised: replace by system-compromise https://github.com/enisaeu/Reference-Security-Incident-Taxonomy-Task-Force/pull/59 -> PR #1993
 - [x] intrusions > defacement: replace by unauthorised-information-modification and a specific identifier https://github.com/enisaeu/Reference-Security-Incident-Taxonomy-Task-Force/pull/73 -> PR #1994
 - [x] intrusions > unauthorized-command not accepted upstream: https://github.com/enisaeu/Reference-Security-Incident-Taxonomy-Task-Force/pull/49 replacement system-compromise: https://github.com/enisaeu/Reference-Security-Incident-Taxonomy-Task-Force/pull/59  -> PR #1993
 - [x] intrusions > unauthorized-login not accepted upstream: https://github.com/enisaeu/Reference-Security-Incident-Taxonomy-Task-Force/pull/49 replacement system-compromise: https://github.com/enisaeu/Reference-Security-Incident-Taxonomy-Task-Force/pull/59  -> PR #1993
 - [x] malicious code > dga domain: ~https://github.com/enisaeu/Reference-Security-Incident-Taxonomy-Task-Force/pull/32~ Rejected by RSIT, see #1613; replace by other > dga-domain -> PR #1992
 - [x] malicious code > malware: replace by infected-system or malware-distribution, PR #1917 
 - [x] malicious code > ransomware: replace by infected-system
 - [x] other > proxy: Other can deviate from RSIT
 - [x] other > tor: Other can deviate from RSIT
 - [x] other > unknown: different name https://github.com/enisaeu/Reference-Security-Incident-Taxonomy-Task-Force/pull/46
 - [x] vulnerable > vulnerable client: vulnerable-system
 - [x] vulnerable > vulnerable service: vulnerable-system
 - [x] `information content security` -> `information-content-security` (taxonomy expert)

https://github.com/enisaeu/Reference-Security-Incident-Taxonomy-Task-Force

cc @aaronkaplan @th-certbund

PR #1533, opened 2020-05-14
set sandbox_url to malware-distribution

n-marques:
Change Sandbox URL classification type from malware to malware-distribution

### DIFF
--- a/intelmq/bots/parsers/shadowserver/config.py
+++ b/intelmq/bots/parsers/shadowserver/config.py
@@ -1382,7 +1382,7 @@ sandbox_url = {
     ],
     'constant_fields': {
         'classification.taxonomy': 'malicious code',
-        'classification.type': 'malware',
+        'classification.type': 'malware-distribution',
         'classification.identifier': 'sandbox-url',
     },
 }


wagner-certat:
related to #1409

wagner-certat:
Hi,

First of all: I agree with the change itself.

This change is also part of a larger commit getting rid of the `malware` type completely: c9e89e762c4111372315c81e98926d323cf74b9d That commit is part of the develop-3.0 branch which will result in the next major release (not a previous patch/maintenance or minor release).

We can prepone this change if we agree this fits into a patch or minor release. I am very hesitant to do this in a patch release as it can cause trouble in any downstream application and dataprocessing. _Maybe_ in minor release? (that would be the develop branch - but it needs some changes/adaptions)

@th-certbund what do you think on that?

th-certbund:
I would prefer to make this change as part of the larger commit for the next major release (3.0).

wagner-certat:
@n-marques Thanks for your input and motivation to contribute and sorry for the begin able to merge your contribution :/

In the meantime I can propose you to use e.g. the modify expert or sieve expert to fix the value in your setup. Example for the modify bot configuration:

```json
[
    {
        "rulename": "Sandbox URL classification.type fix",
        "if": {
            "classification.identifier": "sandbox-url",
            "classification.type": "malware"
        },
        "then": {
            "classification.type": "malware-distribution"
        }
    }
]
```
(untested)

PR #1536, opened 2020-05-14
ENH: improve MISP API Output bot

### LINKED ISSUE #1527: MISP API output enhancements
Some enhancements worth considering:

* BUG: Avoid timeouts then doing lookup on configured ("significant_fields") fields that aren't in the intelmq item itself. This caused a long delay (timeout?) to occur, and a false-positive match on the search.
* ENH: Add a option for feed.name as tag (add_feed_name_as_tag)  
* ENH: Support Bulk output (Events grouped by feed.provider / feed.name ?)
   Add option for bulking multiple intelmq items in the queue, into a single misp event
* ENH: if a MISP event attribute is matched, update the event timestamp rather than doing nothing

bernhardreiter:
* Change behaviour on matched MISP events to update their timestamp if they have the same feed.provider and match closely.
 * Add new option `add_feed_name_as_tag`
 * Add safeguarding against empty significant_fields.

resolves #1527


wagner-certat:
Thanks, I also added the new parameter in the BOTS file 0df8a82c5c9ab3ba00c281177ab211237c41b966

bernhardreiter:
@wagner-certat Thank you very much!

PR #1540, opened 2020-05-19
Version 3.0 ideas

otmarlendl:
This is part feedback, part change suggestions

Slight re-working of the structure to make the main aims clearer. 

We need to think about the microservices part. IMHO it doesn't provide sufficient value.


aaronkaplan:
I think we clarified that the microservice part is optional

PR #1545, opened 2020-06-11
add redis_cache_host to defaults.conf

dominik335:
As it's assumed it's 127.0.0.1

### DIFF
--- a/intelmq/etc/defaults.conf
+++ b/intelmq/etc/defaults.conf
@@ -36,4 +36,5 @@
     "statistics_host": "127.0.0.1",
     "statistics_password": null,
     "statistics_port": 6379
+    "redis_cache_host": "127.0.0.1"
 }


wagner-certat:
For what is this necessary in particular (it's in the templates for all bots needing a cache) and why not also adding the password and the port? (DB and TTL are specific to the exact usage)

dominik335:
My case was that I had redis running in another container than intelmq so host was different. Not really necessary but helpful. And pass and port would fit too.

wagner-certat:
> My case was that I had redis running in another container than intelmq so host was different. Not really necessary but helpful. And pass and port would fit too.

All bots which need the redis cache should have this parameter it's template/documentation. Where is it missing?

dominik335:
Not aware of any place but for me it was much easier to put it in defaults so I thought it could be good to have that along other redis settings in default.conf. Just that ;)

wagner-certat:
Ok. However, I wouldn't consider this worthwhile (at least in it's current state). Nevertheless, thanks for your willingness to contribute :)

(btw it's invalid JSON syntax, the previous line needs a trailing comma)

PR #1554, opened 2020-06-25
ESET ETI collector/parser bots

monoidic:
This PR contains code for two new bots for interfacing with the ESET ETI API, for the JSON-based URL and domain feeds. A new collector is used as there does not currently seem to be a generic TAXII collector.
Any feedback is welcome.

PR #1556, opened 2020-07-01
ESET collector: distribute message sending, no redundant debug messages

monoidic:
Just some small improvements based on feedback in #1554

### DIFF
--- a/intelmq/bots/collectors/eset/collector.py
+++ b/intelmq/bots/collectors/eset/collector.py
@@ -20,7 +20,6 @@ class ESETCollectorBot(CollectorBot):
         self.collection = self.parameters.collection
 
     def process(self):
-        domains = []
         end = datetime.datetime.now(datetime.timezone.utc)
         start = end - datetime.timedelta(seconds=self.time_delta)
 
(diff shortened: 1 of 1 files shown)


wagner-certat:
And then using `event.add('raw', self.recover_line_json_stream())` also fixes the test fails will then use the original message 1:1 (dict key ordering is not consistent in Python)

monoidic:
Iterating over `client.poll` gives a list of dictionaries, otherwise `domains += data` wouldn't have worked previously as concatenating a dict to a list only adds the keys
I suppose maybe giving a list of dicts for the test case might fix it?

wagner-certat:
> Iterating over `client.poll` gives lists of dictionaries, otherwise `domains += data` wouldn't have worked previously as concatenating a dict to a list only adds the keys.

Ah, of course -.-

wagner-certat:
Thanks!

I again did some work based on that: 967e4621f7085da7dd9d9ae1e47ab25a3b898121 As I am unable to test the collector, I hope it still works.

from the commit msg:
> do not save the eset feed in the data anymore but use an extra field
> - collector does not need to interpred, modify and serialize raw data anymore, improves speed and stability (no json errors)
> 
> in parser, add the feed to the event too

PR #1557, opened 2020-07-01
Adds support for ASN (RFC5398)

mlmarkovic:


intelmq/bots/experts/rfc1918/expert.py
```diff
@@ -31,7 +33,11 @@
             "255.255.255.255/32", "fe80::/64", "2001:0db8::/32")
 DOMAINS = ('.test', '.example', '.invalid', '.localhost', 'example.com',
            'example.net', 'example.org')
-
+ASNS = (64496, 64497, 64498, 64499, 64500, 64501, 64502, 64503, 64504, 64505,
```
wagner-certat, line 36: `tuple(range(start, end))` would be more readable

mlmarkovic:
The last commit (a3da9b78fc7ea5c9e30dce173a52c2b0a21d28ca) changes/adds the following:

- the variable `check` was renamed to `netcheck` -> collision with static `check` method
- initialize the `netcheck` variable
- create `ip_networks` instances in `init` (improved performance)
- fixes `fooexample.com` in fqdn and url check
- check for host in ip ranges added
- `Bots.md` and `CHANGELOG.md`updated

wagner-certat:
>     * the variable `check` was renamed to `netcheck` -> collision with static `check` method

Are you sure that they collided? `check` was a local variable, while the check method is `self.check` (even though it's static).

mlmarkovic:
> > ```
> > * the variable `check` was renamed to `netcheck` -> collision with static `check` method
> > ```
> 
> Are you sure that they collided? `check` was a local variable, while the check method is `self.check` (even though it's static).

'collision' might be the incorrect term, but it's still not a bad idea to have different names.

wagner-certat:
a594051727fffeacce92f4e3513ede7ebe15d0c4
38d8b8fe6633fbdffd2d05b468426d7c65fbe396

PR #1560, opened 2020-07-02
ENH: Adds CZ.NIC HaaS (Honeypot as a Service) feed and parser

gethvi:
I assume there is an automatic way of generating updated `Feeds.md` from `feeds.yaml`?


wagner-certat:
> I assume there is an automatic way of generating updated `Feeds.md` from `feeds.yaml`?

Yes, it’s in bin/intelmq_gen_docs.py but maybe it should be moved to contrib/development-Tools. But the generating tool is actually documented in the developer’s guide as well as the feeds.md itself

gethvi:
Thanks! My bad, I totally missed it.

wagner-certat:
Follow-up changes:
0b5ffcd8
3885ffb53bdf2d097eaea8f6e909d46763770bdc
14af72e69e1a8d309e7182a5a6d4dc79a532a814

PR #1572, opened 2020-07-07
FIX: feed revision: Abuse.ch URLhaus

### LINKED ISSUE #1571: Abuse.ch URLhaus feed parsing error
Feeds documentation suggests using generic CSV parser for this feed. However the mapping for columns brings error for lines, where there is no FQDN. Example:

```
"2020-04-03 13:11:03","http://ucto-id.cz/binr.image","offline","malware_download","ucto-id.cz","87.236.194.255","35592","CZ"
"2020-04-04 13:17:12","http://80.211.217.221/hakka/helios.x86","offline","malware_download","80.211.217.221","80.211.217.221","24806","CZ"
```

The first line is parsed fine, because the fifth column contains valid FQDN. However the second line throws InvalidValue exception, because `80.211.217.221` is obviously not valid FQDN.

Secondly, `type_translation` is ignored because the `malware_download` string is assigned to key `extra.urlhaus.threat_type`.

Also `delimeter` parameter is missing.

Fixes incoming.

gethvi:
Fixes #1571.

I kept the `type_translation` even tho the feed always (as far as I can say) sets `Threat` column to `malware_download` (I found very few occasions where the `Threat` column was empty). If there ever appears any other `Threat` value, it will be ignored (the feed documentation doesn't list them therefore we can not map them before they appear and so far it seems very unlikely anyway).

### DIFF
--- a/docs/Feeds.md
+++ b/docs/Feeds.md
@@ -110,8 +110,8 @@ To add feeds to this file add them to `intelmq/etc/feeds.yaml` and then run
 ## URLhaus
 
 * **Public:** yes
-* **Revision:** 2019-02-14
-* **Documentation:** https://urlhaus.abuse.ch/
+* **Revision:** 2020-07-07
+* **Documentation:** https://urlhaus.abuse.ch/feeds/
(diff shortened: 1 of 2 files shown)


wagner-certat:
Thank you! The test fail looks unrelated.

wagner-certat:
Follow-up changes:
0b5ffcd872ec5d170f1a12fbd85716dcd38a1d5d
3885ffb53bdf2d097eaea8f6e909d46763770bdc
14af72e69e1a8d309e7182a5a6d4dc79a532a814

PR #1573, opened 2020-07-08
ESET parser: workaround NXDOMAIN in place of IP, add recover_line

monoidic:
Occasionally, `@ipv4_NXDOMAIN` (and presumably the same with ipv6) may pop up in place of an IP or null in the `ip` field. This is a simple workaround to still add those entries.
Also, a recover_line won't hurt.

### DIFF
--- a/intelmq/bots/parsers/eset/parser.py
+++ b/intelmq/bots/parsers/eset/parser.py
@@ -61,7 +61,7 @@ class ESETParserBot(ParserBot):
         event.add('source.fqdn', line['domain'], raise_failure=False)  # IP addresses may be passed in line['domain']
 
         ip = line['ip']
-        if not ip.endswith('_NXDOMAIN'):
+        if ip and not ip.endswith('_NXDOMAIN'):
             event.add('source.ip', ip)
 
     recover_line = ParserBot.recover_line_json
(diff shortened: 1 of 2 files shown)


intelmq/bots/parsers/eset/parser.py
```diff
@@ -60,7 +60,9 @@ def common_parse(self, event, line):
         event.add('classification.type', type)
         event.add('source.fqdn', line['domain'], raise_failure=False)  # IP addresses may be passed in line['domain']
 
-        event.add('source.ip', line['ip'])
+        event.add('source.ip', line['ip'], raise_failure=False)
```
wagner-certat, line 63: As we know that only `@ipv4_NXDOMAIN` and `@ipv6_NXDOMAIN` appear as values to be ignored, please specifically *only* ignore these. Otherwise we may actually miss something due to other errors.

monoidic: Done with `ip.endswith('_NXDOMAIN')`

wagner-certat:
Thanks.

Follow up commits: a9135c70755e1d5736a4d6e7b77ef0cfe7536e14 950b5ab8c5c06d2337166c1093c52a84c29993c8 (tests)

PR #1574, opened 2020-07-08
FIX: feed revision: DynDNS Infected Domains

gethvi:
* Changes the feed provider name to it's true author.
* Changes feed url to csv file with shadowserver-like syntax (it includes more information).
  * This allows using generic csv parser, therefore `intelmq.bots.parsers.dyn.parser` could be deprecated.

### DIFF
--- a/docs/Feeds.md
+++ b/docs/Feeds.md
@@ -22,7 +22,6 @@ To add feeds to this file add them to `intelmq/etc/feeds.yaml` and then run
 - [DShield](#dshield)
 - [Danger Rulez](#danger-rulez)
 - [Dataplane](#dataplane)
-- [DynDNS](#dyndns)
 - [Fraunhofer](#fraunhofer)
 - [HPHosts](#hphosts)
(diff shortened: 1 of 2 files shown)


gethvi:
Meh, this might be bad PR, the data seems different than at the original url, but generated at the same time. :disappointed: 

Still investigating.

wagner-certat:
> The shadowserver Compromised-Website parser would match too I guess, it would also add the URL (using the Host and URL path). Would that also be sensible?

That gives a lot of `Optional key '...' not found in feed 'Compromised-Website'. Possible change in data format or misconfiguration.` errors.

But having a compound `source.url` field would help a lot.

wagner-certat:
In order to get forward with this, I
* kept the "old" feed format and parser bot
* added a new parameter `compose_fields` to the CSV Parser which allows to compose the source.url from the two columns (see details in bot's docs)
* adapted the feed documentation
and merged your PR finally :)

Thanks for the contribution!

PR #1585, opened 2020-07-21
BUG: parser fails when IP address comes in the domain column

MariusUrkis:
Some alerts from malwareurl.com contain IP address in the Domain column.  Adding IP address value into source.fqdn field results in error.


wagner-certat:
Thanks!

PR #1586, opened 2020-07-24
gethostbyname can use URLs

### LINKED ISSUE #1511: gethostbyname expert: Also get host for URLs if FQDN does not exist
Usecase: I want every record to have `source.ip`. Some records have `source.fqdn`, some `source.url`.
I could easily use `url2fqdn` and `gethostbyname` experts. However, this would create a superfluous `source.fqdn` record for every `source.url` record.

Way 1: 
This way needs no more implementation, however, it will pollute the botnet with many bots.

`Sieve bot`
  * Does the record have `source.ip` → continue
  * Or does the record have `source.fqdn` → pass to `gethostbyname` and continue
  * Or does the record have `source.url` → pass to `url2fqdn` + `gethostbyname` and continue

Way 2:

Implement `gethostbyname` parameter **`accept_url`**. If True and no `source.fqdn` present, `source.url` would be taken to directly produce `source.ip`.

**Would you allow me to implement way 2?**

e3rd:


intelmq/bots/experts/gethostbyname/expert.py
```diff
@@ -21,29 +21,35 @@
 import socket
 
 from intelmq.lib.bot import Bot
+from intelmq.lib.harmonization import URL
 
 
 class GethostbynameExpertBot(Bot):
 
+    def init(self):
+        self.fallback_to_url = getattr(self.parameters, 'fallback_to_url', True)
```
wagner-certat, line 30: Hm, for backwards compatibility, I'd favor `False` here. I'm fine with the default value of True in BOTS.

e3rd: Clever! Changed and added a comment so that future reviewer gets a chance to understand why we are no using the default value from BOTS.

wagner-certat (approved):

PR #1588, opened 2020-07-28
Shadowserver 'blacklist' report renamed to 'blocklist'

th-certbund:
The "Blacklist" report name was changed to "Blocklist".


wagner-certat:
Follow-up commit: c1c9f847

PR #1598, opened 2020-08-17
ENH: CERT-Bund feed for the Avalanche botnet

sinus-x:


wagner-certat:
Hi @sinus-x and thanks for contributing so much!

This PR is marked as Draft, is it ready for review/merge?

At CERT.at we receive two different type of mails by CERT BUND:
* Subject `[CB-Report#20200807-10009061] Malware infections in country AT`
* Subject `[CB-Report#20200819-10000781] Malware infections (Avalanche) in country AT`

Your regular expression matches both feeds, is this intentional?

The format looks very much like plain IntelMQ, possibly we can create and use a generic CSV parser with IntelMQ format?

cc @th-certbund

sinus-x:
Hi.
I must have missed the parser, so I've switched to the generic one.

Both feeds are reports on malware and their e-mail annotations are very similar (they share all their fields and the reported data were indistinguishable, at least for us), we haven't found any reason for keeping them separate. I've altered the bot description to reflect on this.

They have not been tested yet, because I personally do not have access to the data directly. We're hoping on doing it tommorow. (This applies to the rest of my bot PRs, too.)

th-certbund:
The CSV data format of our reports on Avalanche related and other malware infections is identical.
The only difference is the text of the emails with background information.

So you should be able to use IntelMQ's generic csv parser for both reports like this:
```
source.asn,source.ip,time.source,malware.name,source.port,destination.ip,destination.port,
destination.fqdn,protocol.transport
```

For the email collectors, you can use the following regexp's for the subject lines:
```
^\[CB-Report#.* Malware infections \(Avalanche\) in country
^\[CB-Report#.* Malware infections in country
```

wagner-certat (changes requested):
Thank you for your contribution! I have a few comments

PR #1599, opened 2020-08-17
ENH: CZ.NIC's Proki parser

sinus-x:


intelmq/bots/parsers/cznic/parser_proki.py
```diff
@@ -0,0 +1,32 @@
...
+        events = json.loads(raw_report).get("data")
+        for line in events:
+            yield line
+
+    def parse_line(self, line, report):
+        event = self.new_event(report)
+
+        # json keys map 1:1 to harmonization fields
+        for field in line:
+            if field == "feed.name":
```
JakubOnderka, line 23: The similar problem is also with `time.observation` field. 

Parsing fails with exception:
```
  File "/opt/dev_intelmq/intelmq/lib/bot.py", line 978, in process
    events = list(filter(bool, value))
  File "/opt/dev_intelmq/intelmq/bots/parsers/cznic/parser_proki.py", line 23, in parse_line
    event.add(field, line.get(field))
  File "/opt/dev_intelmq/intelmq/lib/message.py", line 224, in add
    raise exceptions.KeyExists(key)
intelmq.lib.exceptions.KeyExists: key 'time.observation' already exists
```

intelmq/bots/parsers/cznic/parser_proki.py
```diff
@@ -0,0 +1,32 @@
...
+from intelmq.lib.bot import ParserBot
+
+
+class CZNICProkiParserBot(ParserBot):
+
+    recover_line = ParserBot.recover_line_json
+
+    def parse(self, report):
+        raw_report = utils.base64_decode(report.get("raw"))
+        events = json.loads(raw_report).get("data")
```
JakubOnderka, line 14: This can be problem when importing data from dump. If message could not be parsed, IntelMQ saves data to dump file, but in that file, message is stored as:

```json
[{"malware.name": "XXX", ...}]
```

So when recovering, this fails, because there is no `data` field.

```
2020-08-20 17:25:07,140 - proki-parser - ERROR - Bot has found a problem.
  File "/opt/dev_intelmq/intelmq/lib/bot.py", line 269, in start
    self.process()
  File "/opt/dev_intelmq/intelmq/lib/bot.py", line 968, in process
    for line in self.parse(report):
  File "/opt/dev_intelmq/intelmq/bots/parsers/cznic/parser_proki.py", line 14, in parse
    events = json.loads(raw_report).get("data")
AttributeError: 'list' object has no attribute 'get'
```

intelmq/bots/parsers/cznic/parser_proki.py
wagner-certat: Is this issue resolved? (at least the tests are ok)

wagner-certat: Is this issue resolved? (at least the tests are ok)

PR #1601, opened 2020-08-18
BUG: danger rulez parser: correctly skip malformed rows

tomas321:
Define variables before referencing

add the same behavior to all dynamic fields i.e. timestamp and ip.


wagner-certat:
Hi,

Thanks. Why not using `None` instead of empty strings?

tomas321:
I used empty strings for no particular reason.. changed it to None and force pushed.

wagner-certat:
Thanks!

PR #1602, opened 2020-08-19
Add GPG signature verification

sinus-x:
Add support for PGP signature verification to HTTP collector.

Requires [`python-gnupg`](https://pypi.org/project/python-gnupg/). Public keys have to be added manually to the `intelmq` user's keyring.

New parameters:
- `verify_pgp_signatures`: Whether to download and verify signatures
- `signature_url`: HTTP location of signature file
- `signature_url_formatting`: formatting option for signature URL; behaves exactly like `http_url_formatting`
- `pgp_keyring`: path to keyring, defaults to `None`, in which case the default path to GPG keyring is used

Tested with (and usable by) Turris Greylist feed.

_Edit 2020-09-03: Adapted to match requested changes._


.travis.yml
```diff
@@ -27,11 +27,11 @@ install:
   - set -e
   - if [[ -v requirements ]]; then sudo apt-get install polipo lighttpd; fi
   - if [[ $mode == debian ]]; then sudo apt-get install dpkg-dev dh-python python-setuptools python3-setuptools python3-all debhelper quilt fakeroot dh-systemd safe-rm; pip3 install requests; pip3 install redis; pip3 install dnspython; pip3 install psutil; pip3 install python-dateutil; pip3 install termstyle; pip3 install pytz; pip3 install typing; fi
-  - if [[ $requirements == true ]]; then for file in intelmq/bots/*/*/REQUIREMENTS.txt; do pip install -r $file; done; fi
-  - if [[ -v requirements ]]; then pip install Cerberus!=1.3 codecov pyyaml requests_mock; fi
-  - if [[ $mode == codestyle ]]; then pip install pycodestyle; fi
+  - if [[ $requirements == true ]]; then for file in intelmq/bots/*/*/REQUIREMENTS.txt; do pip3 install -r $file; done; fi
```
wagner-certat, line 30: Travis only installs python3 there, so `pip` is not ambiguous. Once Python 4 is released, that will break again. Therefore I see no advantage in changing it.

docs/Bots.md
```diff
@@ -258,6 +258,9 @@ The `tornado` library is required.
 * `http_url`: location of information resource (e.g. https://feodotracker.abuse.ch/blocklist/?download=domainblocklist)
 * `http_url_formatting`: (`bool|JSON`, default: `false`) If `true`, `{time[format]}` will be replaced by the current time in local timezone formatted by the given format. E.g. if the URL is `http://localhost/{time[%Y]}`, then the resulting URL is `http://localhost/2019` for the year 2019. (Python's [Format Specification Mini-Language](https://docs.python.org/3/library/string.html#formatspec) is used for this.)
 You may use a `JSON` specifying [time-delta](https://docs.python.org/3/library/datetime.html#datetime.timedelta) parameters to shift the current time accordingly. For example use `{"days": -1}` for the yesterday's date; the URL `http://localhost/{time[%Y-%m-%d]}` will get translated to "http://localhost/2018-12-31" for the 1st Jan of 2019.
+* `verify_gpg_signatures`: `bool`, defaults to `false`. If `true`, signature file is downloaded and report file is checked. On error (missing signature, mismatch, ...), the error is logged and the report is not processed. Public key has to be imported in local keyring.
```
wagner-certat, line 261: Please add that the python library python-gnupg is also needed.

intelmq/bots/BOTS
```diff
@@ -107,7 +107,10 @@
                 "name": "__FEED__",
                 "provider": "__PROVIDER__",
                 "rate_limit": 3600,
-                "ssl_client_certificate": null
+                "ssl_client_certificate": null,
+                "verify_gpg_signatures": false,
```
wagner-certat, line 111: This might be nitpicking, but the format is called PGP, not GPG (which is an implementation). Also applies to `gpg_signature_suffix`

intelmq/bots/collectors/http/collector_http.py
```diff
@@ -75,6 +86,30 @@ def process(self):
 
         self.logger.info("Report downloaded.")
 
+        # GPG verification
+        use_gpg = getattr(self.parameters, 'verify_gpg_signatures', False)
+        if use_gpg:
+            if gnupg is None:
+                raise MissingDependencyError("gnupg")
+
+            result = self.verify_signature(data=resp._content)
```
wagner-certat, line 95: Why do you use `resp._content` and not the public method `resp.content`?

intelmq/bots/collectors/http/collector_http.py
```diff
@@ -100,5 +135,39 @@ def process(self):
...
+    def verify_signature(self, data: bytes):
+        """
+        Download signature file and verify the report data.
+        """
+
+        # get GPG parameters
+        suffix = getattr(self.parameters, "gpg_signature_suffix", ".asc")
+        keyring = getattr(self.parameters, "gpg_keyring", None)
+
+        http_url = self.parameters.http_url + suffix
```
wagner-certat, line 147: That won't work for URL formatting, better provide the actually used URL also as parameter to this method

wagner-certat (changes requested):
Very cool, Thank you very much for your efforts!

Do you know of any feed where we can use this functionality?

sinus-x:
I've updated the code as requested:
- Signature URL is defined in the same way as the main payload URL
- String `PGP` is used instead of `GPG`

All new builds is failing on misconfigured Shadowserver (`ValueError: Could not get a config for 'scan_caida'`), all local tests for this collector are passing.

PR #1607, opened 2020-09-11
Add key-value-parser bot

creideiki:
Bot for parsing key=value strings, e.g. in firewall logs.

### DIFF
--- a/docs/Bots.md
+++ b/docs/Bots.md
@@ -36,6 +36,7 @@
   - [Cymru CAP Program](#cymru-cap-program)
   - [Cymru Full Bogons](#cymru-full-bogons)
   - [HTML Table Parser](#html-table-parser)
+  - [Key-Value Parser](#key-value-parser)
   - [Twitter](#twitter)
   - [Shadowserver](#shadowserver)
   - [Shodan](#shodan)
@@ -1218,6 +1219,31 @@ Parses breaches and pastes and creates one event per e-mail address. The e-mail 
  * `"time_format"`: Optional. If `"timestamp"`, `"windows_nt"` or `"epoch_millis"` the time will be converted first. With the default `null` fuzzy time parsing will be used.
  * `"type"`: set the `classification.type` statically, optional
  * `"html_parser"`: The HTML parser to use, by default "html.parser", can also be e.g. "lxml", have a look at https://www.crummy.com/software/BeautifulSoup/bs4/doc/
+
+* * *
+
+### Key-Value Parser
+
+#### Information:
+* `name:` intelmq.bots.parsers.key_value.parser
+* `lookup:` no
+* `public:` no
+* `cache (redis db):` none
+* `description:` Parses text lines in key=value format.
+
+#### Configuration Parameters:
+
+* `pair_separator`: String separating key=value pairs, default "` `" (space).
+* `kv_separator`: String separating key and value, default `=`.
+* `keys`: Array of string->string, names of keys to propagate mapped to IntelMQ event fields. Example:
+   ```json
+   "keys": {
+       "srcip": "source.ip",
+       "dstip": "destination.ip"
+   }
+   ```
+* `strip_quotes`: Boolean, remove opening and closing quotes from values, default true.
+* `timestamp_key`: String, optional, key containing event timestamp.
 
 * * *
 
--- a/intelmq/bots/BOTS
+++ b/intelmq/bots/BOTS
@@ -535,7 +535,7 @@
             "parameters": {
                 "pair_separator": " ",
                 "kv_separator": "=",
-                "keys": { },
+                "keys": {},
                 "strip_quotes": true,
                 "timestamp_key": null
             }
--- a/intelmq/bots/BOTS
+++ b/intelmq/bots/BOTS
@@ -529,6 +529,17 @@
                 "splitlines": false
             }
         },
+        "Key-Value": {
+            "description": "Parse key=value strings.",
+            "module": "intelmq.bots.parsers.key_value.parser",
+            "parameters": {
+                "pair_separator": " ",
+                "kv_separator": "=",
+                "keys": { },
+                "strip_quotes": true,
+                "timestamp_key": null
+            }
+        },
         "MISP": {
             "description": "Parse MISP events.",
             "module": "intelmq.bots.parsers.misp.parser",
--- a/intelmq/bots/parsers/key_value/parser.py
+++ b/intelmq/bots/parsers/key_value/parser.py
@@ -71,7 +71,7 @@ class KeyValueParserBot(ParserBot):
                         value = DateTime.from_timestamp(int(value))
                     else:
                         value = parse(value, fuzzy=True).isoformat() + " UTC"
-                except:
+                except ValueError:
                     value = None  # Will be ignored by event.add()
             if key in self.keys:
                 event.add(self.keys[key], value, raise_failure=False)
--- a/intelmq/bots/parsers/key_value/parser.py
+++ b/intelmq/bots/parsers/key_value/parser.py
@@ -52,11 +52,6 @@ class KeyValueParserBot(ParserBot):
         if self.timestamp_key:
             self.keys[self.timestamp_key] = "time.source"
 
-    def parse(self, report):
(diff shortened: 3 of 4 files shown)


docs/Bots.md
```diff
@@ -1221,6 +1222,31 @@ Parses breaches and pastes and creates one event per e-mail address. The e-mail
...
+* `pair_separator`: String separating key=value pairs, default "` `" (space).
+* `kv_separator`: String separating key and value, default `=`.
+* `keys`: Array of string->string, names of keys to propagate mapped to IntelMQ event fields. Example:
+   ```json
+   "keys": {
+       "srcip": "source.ip",
+       "dstip": "destination.ip"
+   }
+   ```
+* `strip_quotes`: Boolean, remove opening and closing quotes from values, default true.
```
wagner-certat, line 1245: This is only valid for double quotes (not single quotes annd backticks). Please specify this here.

intelmq/bots/parsers/key_value/parser.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def parse(self, report):
+        raw_report = utils.base64_decode(report.get('raw'))
+        for row in raw_report.splitlines():
+            yield row
+
+    def parse_line(self, row, report):
+        event = self.new_event(report)
+        for kv_pair in row.split(self.pair_separator):
+            (key, _, value) = kv_pair.rpartition(self.kv_separator)
```
wagner-certat, line 62: That doesn't work if the line has (quoted) occurrences of the separator. For example:
```
key="long value" key2="other value"
```
I don't know if your use-case (some firewall logs) can every include that. However it is either a limitation (which should be documented) or a bug.

intelmq/bots/parsers/key_value/parser.py
```diff
@@ -0,0 +1,81 @@
...
+            if not key:
+                continue
+            if self.strip_quotes and value.startswith('"') and value.endswith('"'):
```
wagner-certat, line 65: See my comment above on the parameter. Extending it to other quotes is easy, but this issue is also related to the parsing issue.

intelmq/bots/parsers/key_value/parser.py
```diff
@@ -0,0 +1,81 @@
...
+                value = value[1:-1]
+            if key == self.timestamp_key:
+                try:
+                    if value.isnumeric():
+                        value = DateTime.from_timestamp(int(value))
+                    else:
+                        value = parse(value, fuzzy=True).isoformat() + " UTC"
+                except:
+                    value = None  # Will be ignored by event.add()
```
wagner-certat, line 74: If we can't parse it, do we *really* want to ignore this? That drops data.

intelmq/bots/parsers/key_value/parser.py
```diff
@@ -0,0 +1,81 @@
...
+                continue
+            if self.strip_quotes and value.startswith('"') and value.endswith('"'):
+                value = value[1:-1]
+            if key == self.timestamp_key:
+                try:
+                    if value.isnumeric():
+                        value = DateTime.from_timestamp(int(value))
+                    else:
+                        value = parse(value, fuzzy=True).isoformat() + " UTC"
+                except:
```
wagner-certat, line 73: Don't use a bare except.

wagner-certat (changes requested):
Cool addition! Thank you for this good PR (including docs and tests).

What firewall format is that? (Then we could document how to parse it)

intelmq/bots/parsers/key_value/parser.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def init(self):
+        self.pair_separator = getattr(self.parameters, 'pair_separator', ' ')
+        self.kv_separator = getattr(self.parameters, 'kv_separator', '=')
+        self.keys = getattr(self.parameters, 'keys', {})
+        if not self.keys:
+            raise ConfigurationError('Key extraction', 'No keys specified.')
+        self.strip_quotes = getattr(self.parameters, "strip_quotes", True)
+        self.timestamp_key = getattr(self.parameters, "timestamp_key", None)
+        if self.timestamp_key:
```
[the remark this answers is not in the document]
creideiki: `time.source` is special because all other values are just copied, while the timestamp is parsed.

intelmq/bots/BOTS
```diff
@@ -529,6 +529,17 @@
                 "splitlines": false
             }
         },
+        "Key-Value": {
+            "description": "Parse key=value strings.",
+            "module": "intelmq.bots.parsers.key_value.parser",
+            "parameters": {
+                "pair_separator": " ",
+                "kv_separator": "=",
+                "keys": { },
```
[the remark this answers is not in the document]
creideiki: Fixed.

intelmq/bots/parsers/key_value/parser.py
creideiki: This is a limitation, which does not affect my use-case. The logs I'm processing contains fields like `srcip=192.0.2.1` and `srcmac="00:00:5e:00:17:17"`, for which this handling suffices and keeps the parser extremely simple. But you're right that it should be documented.

docs/Bots.md
creideiki: Documentation added.

creideiki:
> What firewall format is that? (Then we could document how to parse it)

This is a FortiGate, but I'm pretty sure I've seen other brands of firewall use similar formats.

intelmq/bots/parsers/key_value/parser.py
creideiki: Changed to the `ValueError` which `dateutil.parser.parse` actually raises on a parse error.

intelmq/bots/parsers/key_value/parser.py
creideiki: This is correct. However, it only drops the timestamp, and lets the rest of the fields through. A garbage timestamp will fail harmonisation and lead to the entire event being dropped.

I think. But apparently my development environment is halfway between working states at the moment, so I haven't actually tried.

intelmq/bots/parsers/key_value/parser.py
```diff
@@ -0,0 +1,81 @@
...
+            self.keys[self.timestamp_key] = "time.source"
+
+    def parse(self, report):
```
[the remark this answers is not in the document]
creideiki: Looking at the code, it should be. But, as below, my development environment has an unrelated problem at the moment, so I can't actually check right now.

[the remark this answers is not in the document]
creideiki: I pushed a change removing my `parse()` method, so let's see what happens.

intelmq/bots/parsers/key_value/parser.py
creideiki: I have now repaired my development environment. What happens when trying to add an unparseable timestamp field is that if `event.add()` is called with `raise_failure=False`, the field is silently discarded. If `raise_failure=True` (which is the default), the whole message is discarded and an error is noted in the bot's log file. So I think that throwing away the timestamp is better than throwing away the entire message.

intelmq/bots/parsers/key_value/parser.py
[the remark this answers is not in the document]
wagner-certat: I changed to code a bit to make this parameter obsolete and removed it. It is now detected from the mapping if the value needs to be parsed for `time.source`.

intelmq/bots/parsers/key_value/parser.py
wagner-certat: I documented this in the Bots.md file

intelmq/bots/parsers/key_value/parser.py
wagner-certat: I now added a warning log message in case this happens.

wagner-certat:
> > What firewall format is that? (Then we could document how to parse it)
> 
> This is a FortiGate, but I'm pretty sure I've seen other brands of firewall use similar formats.

Thanks, I also added this to the docs.

Once again, thank you very much for your efforts and your contribution!

PR #1620, opened 2020-10-12
sqlite docs updated

e3rd:


e3rd:
( + I pushed `import argparse` to the top because it should be ordered alphabetically)

wagner-certat:
Thanks!

PR #1622, opened 2020-10-14
Integrate sphinx to build documentation

schacht-certat:


docs/requirements.txt
```diff
@@ -0,0 +1,12 @@
+recommonmark==0.6.0
+Sphinx==3.2.1
+sphinx-markdown-tables==0.0.15
+PyYAML==5.3.1
```
wagner-certat, line 4: I don't approve those exact version requirements. Can we use `>=` instead?

schacht-certat: done ;)

PR #1623, opened 2020-10-15
parser_ctip: Update AZURE fields

monoidic:
This commit adds support for some fields which are used in a CTIP feed.


intelmq/bots/parsers/microsoft/parser_ctip.py
```diff
@@ -91,9 +91,21 @@
...
+    "DestinationIpInfo.DestinationIpAsnOrgName": "source.as_name",
+    "DestinationIpInfo.DestinationIpCountryCode": "source.geolocation.cc",
+    "DestinationIpInfo.DestinationIpRegion": "source.geolocation.region",
+    "DestinationIpInfo.DestinationIpCity": "source.geolocation.city",
+    "DestinationIpInfo.DestinationIpPostalCode": "extra.source.geolocation.postal_code",
+    "DestinationIpInfo.DestinationIpLatitude": "source.geolocation.latitude",
+    "DestinationIpInfo.DestinationIpLongitude": "source.geolocation.longitude",
+    "DestinationIpInfo.DestinationIpMetroCode": "extra.source.geolocation.metro_code",
+    "DestinationIpInfo.DestinationIpAreaCode": "extra.source.geolocation.area_code",
+    "DestinationIpInfo.DestnationIpConnectionType": "protocol.application",
```
wagner-certat, line 104: I think we are mixing here source and destination. `DestinationIp` results in `destination.ip` but the corresponding asn `DestinationIpInfo.DestinationIpAsnNumber` is saved as `source.asn`.

wagner-certat (changes requested):
Thanks for the PR!

I initially only tested the parser with the ctip-infected-summary feed. It looks like these fields are only present in the ctip-c2 blob container/feed, is this correct?

intelmq/bots/parsers/microsoft/parser_ctip.py
monoidic: Indeed, this would have been a part of an ugly hack, but I'll just map destination to destination properly instead.

monoidic:
Indeed, they are from the c2 feed.

wagner-certat (approved):

wagner-certat:
Thanks again @monoidic 

Merged and picked for next bugfix release

PR #1627, opened 2020-10-21
Update URLs in README to the new canonical documentation URL

schacht-certat:


setup.py
```diff
@@ -24,10 +24,6 @@
         module = bot['module']
         BOTS.append('{0} = {0}:BOT.run'.format(module))
 
-with open(os.path.join(os.path.dirname(__file__), 'README.rst')) as handle:
-    README = handle.read().replace('<docs/',
-                                   '<https://github.com/certtools/intelmq/blob/master/docs/')
```
wagner-certat, line 29: The variable `README` is used below

wagner-certat (changes requested):

setup.py
schacht-certat: fixed

wagner-certat (approved):

PR #1637, opened 2020-10-26
Rotating file handler added as another logging option

Narzhan:
The purpose of this PR is to add the support for log rotation to the intelmq. The need for this is, at least in my case, for log running deployments typically in Docker. Since the processing info logs can take a large portion of disk space this feature could be used as a solution. I'm well aware that the amount of logs could be reduced by restricting the log level to warning of error but this solution is in my opinion better since it allows for a greater control over disk usage and keeping the info logs, which might still hold some value for troubleshooting.

To enable rotation of logs I've added two parameters _logging_max_size_ and _logging_max_copies_ which directly map to the values for limiting the sizes of logs and backup counts from Python's [RotatingFileHandler](https://docs.python.org/3/library/logging.handlers.html#rotatingfilehandler). If both parameters are set the Rotating file handler is used instead of the default FileHandler.

I haven't customised the handler class, like it is done for example for [FileHandler](https://github.com/certtools/intelmq/blob/develop/intelmq/lib/utils.py#L259), since this would require to handle the log rotation in some way and I wasn't sure whether there is still any need for overriding the emit. Also I haven't added any test cases yet since I wasn't sure whether it is needed for logging purposes. If any of these two features are needed just let me know and I will add them.

### DIFF
--- a/intelmq/lib/bot.py
+++ b/intelmq/lib/bot.py
@@ -729,7 +729,9 @@ class Bot(object):
             syslog = False
         self.logger = utils.log(self.__bot_id_full, syslog=syslog,
                                 log_path=self.parameters.logging_path,
-                                log_level=self.parameters.logging_level)
+                                log_level=self.parameters.logging_level,
+                                log_max_size=getattr(self.parameters, "logging_max_size", None),
+                                log_max_copies=getattr(self.parameters, "logging_max_copies", None))
 
     def __load_pipeline_configuration(self):
         self.logger.debug("Loading pipeline configuration from %r.", PIPELINE_CONF_FILE)
--- a/intelmq/lib/utils.py
+++ b/intelmq/lib/utils.py
@@ -307,7 +307,7 @@ def log(name: str, log_path: Union[str, bool] = intelmq.DEFAULT_LOGGING_PATH,
         log_level: str = intelmq.DEFAULT_LOGGING_LEVEL,
         stream: Optional[object] = None, syslog: Union[bool, str, list, tuple] = None,
         log_format_stream: str = LOG_FORMAT_STREAM,
-        logging_level_stream: Optional[str] = None):
+        logging_level_stream: Optional[str] = None, log_max_size: Optional[int] = None, log_max_copies: Optional[int] = None):
     """
     Returns a logger instance logging to file and sys.stderr or other stream.
     The warnings module will log to the same handlers.
@@ -350,7 +350,11 @@ def log(name: str, log_path: Union[str, bool] = intelmq.DEFAULT_LOGGING_PATH,
         logging_level_stream = log_level
 
     if log_path and not syslog:
-        handler = FileHandler("%s/%s.log" % (log_path, name))
+        if not all((log_max_size, log_max_copies)):
+            handler = FileHandler("%s/%s.log" % (log_path, name))
+        else:
+            handler = logging.handlers.RotatingFileHandler("%s/%s.log" % (log_path, name), maxBytes=log_max_size,
+                                                           backupCount=log_max_copies)
         handler.setLevel(log_level)
         handler.setFormatter(logging.Formatter(LOG_FORMAT))
     elif syslog:


wagner-certat:
Thanks @Narzhan for the contribution!

If I understand the docs of RotatingFileHandler correctly, we can use it in place of the normal file handler and if the additional parameters are not set (which would be the default), the behavior would stay the same.

> I haven't customised the handler class, like it is done for example for [FileHandler](https://github.com/certtools/intelmq/blob/develop/intelmq/lib/utils.py#L259), since this would require to handle the log rotation in some way

I don't think that log rotation needs to be handled in the derived class when still the parent's method is called by using `super()`.

> and I wasn't sure whether there is still any need for overriding the emit.

The overrides have two reasons:
- To remove potentially used shell coloring. A quick search showed that this is only using by intelmqctl and intelmqdumo tools to give colored feedback.
- In case of errors writing the logs to the file, the traceback is given on stdout to at least show it somewhere. (Actually it should be stderr).

As the `logging.FileHandler` and `logging.handlers.RotatingFileHandler` classes are very similar and if we can use the RotatingFileHandler in place of the FileHandler, just changing the parent class for the overriding class should be fine I guess.

waldbauer-certat:
Hey @Narzhan,

if you want to use docker & want to store logs, I would recommend to use [Filebeat](https://www.elastic.co/beats/filebeat) instead. Docker best practice would be [stdout/stderr](https://docs.docker.com/config/containers/logging/) ( currently discussed, due to breaking changes ), but [Filebeat](https://www.elastic.co/beats/filebeat) is currently also fine.

I'm currently trying to figure out a solution which fills the gap. In the current state of the software I would, as mentioned above, use [Filebeat](https://www.elastic.co/beats/filebeat).

Narzhan:
@wagner-certat My understanding is that _RotatingFileHandler_ behaves similarly to the _FileHandler_ only difference is that in _emit()_ the rotating version does the log rotation itself(checks file sizes, if they are near the max size, create new, ...) just as is mentioned [here](https://docs.python.org/3/library/logging.handlers.html#logging.handlers.RotatingFileHandler.emit). 

I think that my approach is optimal in the sense that if you set the rotating handler's parameters it is used if not everything stays the same.

@waldbauer-certat I'm not really interested in storing the logs, I just want to keep some of them for eventual troubleshooting. The problem right now is the logs could get so big that they fill up the storage space on a longer running instance of Intelmq.

wagner-certat:
I added documentation and removed the FileHandler. I have set the log_max_bytes value to 0 as None would raise an error in the handler

35e3aeaf6d4dcd79bd0421b6248090beaaacf66e

Thanks again @Narzhan  for the contribution!

aaronkaplan:
I tend to agree with @waldbauer-certat  here. From an architecture point of view, adding functionality (because it might be needed in docker) into intelmq which is actually *better* solved outside of docker (see the comment on best practices of logging in docker via stderr/stdout and then using a *standard* docker logfile transport such as syslog output or ELK output or ... you name it.. .filebeat, etc.) is better.
Why?

1. Because it reduces complexity.
2. Because it keeps IntelMQ cleaner, smaller, compact
3. Because it's a nice _separation of concerns_ approach: let the pros do the log transportation and -rotation

Not sure if this should have been merged if it is *because of docker*. See my point?

wagner-certat:
> I tend to agree with @waldbauer-certat here. From an architecture point of view, adding functionality (because it might be needed in docker) into intelmq which is actually _better_ solved outside of docker (see the comment on best practices of logging in docker via stderr/stdout and then using a _standard_ docker logfile transport such as syslog output or ELK output or ... you name it.. .filebeat, etc.) is better.

No-one contributed Filebeat-integration yet.

> Not sure if this should have been merged if it is _because of docker_. See my point?

Despite that I do not defend myself why I review and care about pull requests: The change is minimal, backwards compatible and resolves the needs of some users by reducing the complexity of their set-ups.

Issue #758, opened 2016-10-27
Notify flag in the DHO


sykaeh:
I would like to add a "notify" flag to the data harmonization ontology (DHO). This field specifies whether the responsible parties should be notified about this event. Our use case is that an expert then decides based on the attributes of the event (classification.type, classification.identifier, ip, time.source, etc.) whether people should be notified about this event and sets the flag appropriately. This can then be used to filter events when sending notifications/uploading to a database etc. Is anyone else interested in such a flag?

aaronkaplan:
> On 27 Oct 2016, at 18:08, Sybil Ehrensberger <notifications@github.com> wrote:
> 
> I would like to add a "notify" flag to the data harmonization ontology (DHO). This field specifies whether the responsible parties should be notified about this event.

okay. So the notify flag actually already exists in cert.at's version.
We were not sure if it is generally relevant. But it seems to be. So how about putting that "upstream" to the main certtools/intelmq repo?


> Our use case is that an expert then decides based on the attributes of the event (classification.type, classification.identifier, ip, time.source, etc.) whether people should be notified about this event and sets the flag appropriately. This can then be used to filter events when sending notifications/uploading to a database etc. Is anyone else interested in such a flag?
> 
Excellent.
Yes, we have it already.

Anyone else? If so, I'll merge that upstream.

Best,
a.

bernhard-herzog:
We're going to need something similar, but more refined. A simple boolean flag would not enough. We have to be able to decide and record in the event who to notify in what way. Our current plans basically require some JSON structure for this.

dmth:
> So am i hearing that you need some Kind of variable which tracks some state of an event? ( and that notify Boolean would not be enough)?

I'll try to elaborate a little bit: Currently we are "mis-using" the `extras` field in order to store information about contacts and how they should be informed. This information can contain multiple contacts with a different set of rules. We'd appreciate a field which is intended to contain such information. But this field cannot be a boolean, because we support the notification of multiple contacts.

`notify` Is a good name, but maybe a little bit to specific for a use-case were this field might contain specific processing information, like provenance of data.

Such a `processing`-field might contain something like this (fictional structure below):

```
"processing_info": [
  "dns_lookup_expert": {
    "used_nameserver": "ns1.example.com",
    "lookup_timestamp": 4711
  },
  "flow-monitor": {
    "timestamp": "4712"
  },
  "notifications": [{
     "who": "mail@example.com"
     "how": "e-mail"
[...]

PR #1649, opened 2020-11-04
MISP parser: added functionality to honor/filter the "to_ids" attribute of MISP

exitnode:
Added the following to honor the "to_ids" attribute of MISP:

* added the new field "misp.to_ids" to harmonization.conf (Boolean)
* added the new parameter "only_ids" to the MISP parser
* MISP parser now inserts the MISP attribute "to_ids" into the resulting IntelMQ events
* When parameter "only_ids" is set to "true", the MISP parser only processes and forwards MISP events if their "to_ids" attribute's value is true

### DIFF
--- a/intelmq/etc/harmonization.conf
+++ b/intelmq/etc/harmonization.conf
@@ -205,6 +205,10 @@
             "regex": "^[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[0-9a-z]{12}$",
             "type": "LowercaseString"
         },
+        "misp.to_ids": {
+            "description": "MISP - Malware Information Sharing Platform & Threat Sharing IDS flag",
+            "type": "Boolean"
+        },
         "output": {
             "description": "Event data converted into foreign format, intended to be exported by output plugin.",
             "type": "JSON"
--- a/docs/user/bots.rst
+++ b/docs/user/bots.rst
@@ -1341,6 +1341,10 @@ for processing. Supported MISP event categories and attribute types are
 defined in the `SUPPORTED_MISP_CATEGORIES` and `MISP_TYPE_MAPPING` class
 constants.
 
+**Configuration Parameters**
+
+* `only_ids`: If set to true, MISP events will be discardrd if their to_ids attribute is set to false
+
 n6
 ^^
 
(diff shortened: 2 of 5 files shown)


wagner-certat:
This introduces a new field in the harmonization (internal data format) -> @aaronkaplan

Rafiot:
It seems fine to me.

wagner-certat:
> It seems fine to me.

Thanks!

wagner-certat:
@exitnode Can you please better explain what the `to_ids` field is for (the given explanation in `harmonization.conf` is way too generic`) and why we need a separate field for it in IntelMQ?

exitnode:
> 
> 
> @exitnode Can you please better explain what the `to_ids` field is for (the given explanation in `harmonization.conf` is way too generic`) and why we need a separate field for it in IntelMQ?

This is from the MISP core format documentation:

```
to_ids represents whether the attribute is meant to be actionable.
Actionable defined attributes that can be used in automated processes
as a pattern for detection in Local or Network Intrusion Detection
System, log analysis tools or even filtering mechanisms
```

We use the "to_ids" flag inside of MISP to decide if an IOC will be later on be included in a blocking list e.g. for a proxy server. We use this to manually verify what will be blocked and what not while working on the MISP events.

Now we can filter inside IntelMQ if an event' property is meant for blocking or not and therefore put it in a pipeline that eventually generates the blocking list.

FYI: I'm working on additional enhancements to the MISP collector that will also include this property and some more functionality. As soon as it is stable enough, I will send a PR for it, too.

intelmq/etc/harmonization.conf
```diff
@@ -205,6 +205,10 @@
             "regex": "^[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[0-9a-z]{12}$",
             "type": "LowercaseString"
         },
+        "misp.to_ids": {
+            "description": "MISP - Malware Information Sharing Platform & Threat Sharing IDS flag",
```
wagner-certat, line 209: For me, this is not descriptive enough. I can't understand the meaning of the field from this text.

wagner-certat (commented):
If I got the meaning of the `to_ids` field right, I think this is similar to the discussion #758. Both fields have equivalent meanings, i.e. that the IoC is actionable. IMHO we should only have one field, not one field per input/output as long as they do not conflict (I don't see this risk at the moment)

PR #1654, opened 2020-11-09
Implement a simple Kafka Collector

### LINKED ISSUE #1634: Kafka collector
To collect data from (non-IntelMQ) Kafka Message Buses, we need a collector for kafka

schacht-certat:
Closes #1634

### DIFF
--- a/intelmq/bots/collectors/kafka/collector.py
+++ b/intelmq/bots/collectors/kafka/collector.py
@@ -1,0 +1,52 @@
+"""
+Kafka Collector Bot
+
+Collects information from the Apache Kafka distributed stream processing system.
+
+Args:
+    topic (str): topic to collect information from
+    bootstrap_servers (str): the ‘host[:port]’ string of the Apache Kafka system. Defaults to `localhost:9092`
+"""
+
+from intelmq.lib.bot import CollectorBot
+from intelmq.lib.exceptions import MissingDependencyError
+
+try:
+    import kafka
+except ImportError:
+    kafka = None
+
+
+class KafkaCollectorBot(CollectorBot):
+
+    def init(self):
+        if kafka is None:
+            raise MissingDependencyError("kafka")
+
+        self.topic = []
+        if getattr(self.parameters, 'topic', '') != '':
+            self.topic = self.parameters.topic
+
+        self.bootstrap_servers = 'localhost:9092'
+        if getattr(self.parameters, 'bootstrap_servers', '') != '':
+            self.bootstrap_servers = self.parameters.bootstrap_servers
+
+        self.ssl_cafile = getattr(self.parameters, 'ssl_ca_certificate', None)
+        self.ssl_certfile = getattr(self.parameters, 'ssl_client_certificate', None)
+        self.ssl_check_hostname = getattr(self.parameters, 'ssl_check_hostname', False)
+
+        self.logger.debug("Topic set to {}, bootstrap_servers set to {}".format(self.topic, self.bootstrap_servers))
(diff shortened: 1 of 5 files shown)


intelmq/bots/collectors/kafka/collector.py
```diff
@@ -0,0 +1,45 @@
...
+
+        self.logger.debug(f"topic set to {self.topic}, bootstrap_servers set to {self.bootstrap_servers}")
+
+    def process(self):
+        " Fetch messages from the Kafka server and pass them on one by one "
+        consumer = kafka.KafkaConsumer(self.topic, bootstrap_servers=self.bootstrap_servers, auto_offset_reset='earliest', consumer_timeout_ms=100, group_id=self._Bot__bot_id)
+        for msg in consumer:
+            consumer.commit()
+            report = self.new_report()
+            self.logger.debug(f"Received msg with offset {msg.offset} from kafka topic {msg.topic}: {msg.value.decode()}")
```
wagner-certat, line 41: that requires python 3.6, but intelmq supports 3.5 as well. See also the test fail: https://travis-ci.org/github/certtools/intelmq/builds/742395055

Instead of requiring a different python version for this bot and excluding it from the tests, I propose not using format strings as long as 3.5 is supported.

intelmq/bots/collectors/kafka/collector.py
```diff
@@ -0,0 +1,45 @@
...
+
+        self.bootstrap_servers = 'localhost:9092'
+        if getattr(self.parameters, 'bootstrap_servers', '') != '':
+            self.bootstrap_servers = self.parameters.bootstrap_servers
+
+        self.logger.debug(f"topic set to {self.topic}, bootstrap_servers set to {self.bootstrap_servers}")
+
+    def process(self):
+        " Fetch messages from the Kafka server and pass them on one by one "
+        consumer = kafka.KafkaConsumer(self.topic, bootstrap_servers=self.bootstrap_servers, auto_offset_reset='earliest', consumer_timeout_ms=100, group_id=self._Bot__bot_id)
```
wagner-certat, line 37: Here we could use some generic parameters:
- `ssl_client_certificate` as `ssl_certfile`
- `ssl_ca_certificate` as `ssl_cafile`
- `http_verify_cert` as `ssl_check_hostname`
- _maybe_ `http_timeout_sec` as `request_timeout_ms`, `session_timeout_ms`, `heartbeat_interval_ms` and `consumer_timeout_ms`

Maybe we need to add authentication as well, but later

wagner-certat (changes requested):
Please add user documentation to `docs/user/bots.rst`
and you get (useless) bonus points if you also add the entry to the changelog :)
minor codestyle issues: https://travis-ci.org/github/certtools/intelmq/jobs/742395065#L345

intelmq/bots/collectors/kafka/collector.py
schacht-certat: fixed

intelmq/bots/collectors/kafka/collector.py
schacht-certat: I did that, but one thing: the doc says about `http_verify_cert`: `path to trusted CA bundle or directory, false to ignore verifying SSL certificates, or true (default) to verify SSL certificates` -  but`ssl_check_hostname`only allows `true` or `false`. Maybe we should not use the generic variable for the kafka collector?

wagner-certat: you're right, that can cause conflicts and confusion

schacht-certat: I have now changed the variable to `ssl_check_hostname`. I think it confusing to have a variable with `http` in the name used for something that has nothing to do with http.

PR #1655, opened 2020-11-11
Initialize logging_level attribute

schacht-certat:


intelmq/bin/intelmqctl.py
```diff
@@ -681,6 +681,7 @@ def __init__(self, interactive: bool = False, return_type: str = "python", quiet
             no_file_logging: do not log to the log file
             drop_privileges: Drop privileges and fail if it did not work.
         """
+        self.logging_level = None
```
wagner-certat, line 684: I would prefer `intelmq.DEFAULT_LOGGING_LEVEL` or some other fixed sane default value

wagner-certat (changes requested):

intelmq/bin/intelmqctl.py
schacht-certat: fixed

PR #1657, opened 2020-11-12
Modify BotDebugger to return output instead of dumping it to stdout

schacht-certat:
This way we can use the BotDebugger methods and its return values
internall, resp. it can be used by another tool that wants to somehow
present the output differently (like an API).
One downside is, that the output is being `buffered`, i.e. *all* the
messages are only returned when the BotDebugger `run` method returns.

This also updates the `intelmqctl` script to use the new BotDebugger
interface.

### DIFF
--- a/intelmq/bin/intelmqctl.py
+++ b/intelmq/bin/intelmqctl.py
@@ -171,10 +171,12 @@ class IntelMQProcessManager:
         with open(filename, 'w') as fp:
             fp.write(str(os.getpid()))
 
+        output = ""
         try:
-            BotDebugger(self.__runtime_configuration[bot_id], bot_id, run_subcommand,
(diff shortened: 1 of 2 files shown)


wagner-certat:
> One downside is, that the output is being `buffered`, i.e. _all_ the
> messages are only returned when the BotDebugger `run` method returns.

The buffering only applies to the messages of the BotDebugger class itself, not to the messages/logging of the bot if I interpret the code correctly, right?

schacht-certat:
On 11/12/20 12:21 PM, Wagner wrote:
>     One downside is, that the output is being |buffered|, i.e. /all/ the
>     messages are only returned when the BotDebugger |run| method returns.
> 
> The buffering only applies to the messages of the BotDebugger class 
> itself, not to the messages/logging of the bot if I interpret the code 
> correctly, right?

yes!

PR #1666, opened 2020-11-24
Splunk saved search expert bot

creideiki:
An expert bot that enriches events based on Splunk saved searches.

### DIFF
--- a/docs/user/bots.rst
+++ b/docs/user/bots.rst
@@ -2653,7 +2653,7 @@ Splunk saved search
 
     * `send` and `drop` are mutually exclusive
 
-* `duplicates`: List of strings, default ``[ "warn", "use_first", "send" ]``, what to do if the search returns more than one result. All specified actions are performed. Valid values are:
+* `multiple_result_handling`: List of strings, default ``[ "warn", "use_first", "send" ]``, what to do if the search returns more than one result. All specified actions are performed. Valid values are:
 
   * `limit`: limit the search so that duplicates are impossible
   * `warn`: log a warning message
--- a/docs/user/bots.rst
+++ b/docs/user/bots.rst
@@ -2655,12 +2655,14 @@ Splunk saved search
 
 * `duplicates`: List of strings, default ``[ "warn", "use_first", "send" ]``, what to do if the search returns more than one result. All specified actions are performed. Valid values are:
 
+  * `limit`: limit the search so that duplicates are impossible
   * `warn`: log a warning message
   * `use_first`: use the first search result
   * `ignore`: do not modify the event
   * `send`: send the event on
   * `drop`: drop the message
 
+    * `limit` cannot be combined with any other value
     * `send` and `drop` are mutually exclusive
     * `ignore` and `use_first` are mutually exclusive
 
--- a/docs/user/bots.rst
+++ b/docs/user/bots.rst
@@ -2680,7 +2680,7 @@ The saved search should take parameters according to the `search_parameters` con
 
 The time window used is the one saved with the search.
 
-Waits for Splunk to return an answer for each message, so slow searches will delay the entire botnet.
+Waits for Splunk to return an answer for each message, so slow searches will delay the entire botnet. If you anticipate a load of more than one search every few seconds, consider running multiple load-balanced copies of this bot.
 
 Taxonomy
 ^^^^^^^^
--- a/docs/user/bots.rst
+++ b/docs/user/bots.rst
@@ -2670,6 +2670,8 @@ Splunk saved search
 
 Runs a saved search in Splunk using fields in an event, adding fields from the search result into the event.
 
+Splunk documentation on saved searches: https://docs.splunk.com/Documentation/Splunk/latest/Report/Createandeditreports
+
 The saved search should take parameters according to the `search_parameters` configuration and deliver results according to `result_fields`. The examples above match a saved search of this format:
 
 ::
--- a/docs/user/bots.rst
+++ b/docs/user/bots.rst
@@ -2645,7 +2645,7 @@ Splunk saved search
          "username": "source.account"
      }
 
-* `not_found`: List of strings, default ``[ "warn", "send" ]``, what to do if the search returns zero results. Valid values are:
+* `not_found`: List of strings, default ``[ "warn", "send" ]``, what to do if the search returns zero results. All specified actions are performed. Valid values are:
 
   * `warn`: log a warning message
   * `send`: send the event on unmodified
@@ -2653,7 +2653,7 @@ Splunk saved search
 
     * `send` and `drop` are mutually exclusive
 
-* `duplicates`: List of strings, default ``[ "warn", "use_first", "send" ]``, what to do if the search returns more than one result. Valid values are:
+* `duplicates`: List of strings, default ``[ "warn", "use_first", "send" ]``, what to do if the search returns more than one result. All specified actions are performed. Valid values are:
 
   * `warn`: log a warning message
   * `use_first`: use the first search result
--- a/docs/user/bots.rst
+++ b/docs/user/bots.rst
@@ -2611,6 +2611,73 @@ Use the following command to validate your sieve files:
    optional arguments:
      -h, --help  show this help message and exit
 
+Splunk saved search
+^^^^^^^^^^^^^^^^^^^
+
+**Information**
+
+* `name`: splunk_saved_search
+* `lookup`: splunk database
+* `public`: no
+* `cache (redis db)`: none
+* `description`: Enrich an event from Splunk search results.
+
+**Configuration Parameters**
+
+* **HTTP parameters** (see above)
+* `auth_token`: String, Splunk API authentication token
+* `url`: String, base URL of the Splunk REST API
+* `retry_interval`: Integer, optional, default 5, number of seconds to wait between attempts to retrieve search results
+* `saved_search`: String, name of Splunk saved search to run
+* `search_parameters`: Array of string->string, optional, default ``{}``, IntelMQ event fields containing the data to search for, mapped to parameters of the Splunk saved search. Example:
+
+  .. code-block:: json
+
+     "search_parameters": {
+         "source.ip": "ip"
+     }
+
+* `result_fields`: Array of string->string, optional, default ``{}``, Splunk search result fields mapped to IntelMQ event fields to store the results in. Example:
+
+  .. code-block:: json
+
+     "result_fields": {
+         "username": "source.account"
+     }
+
+* `not_found`: List of strings, default ``[ "warn", "send" ]``, what to do if the search returns zero results. Valid values are:
+
+  * `warn`: log a warning message
+  * `send`: send the event on unmodified
+  * `drop`: drop the message
+
+    * `send` and `drop` are mutually exclusive
+
+* `duplicates`: List of strings, default ``[ "warn", "use_first", "send" ]``, what to do if the search returns more than one result. Valid values are:
+
+  * `warn`: log a warning message
+  * `use_first`: use the first search result
+  * `ignore`: do not modify the event
+  * `send`: send the event on
+  * `drop`: drop the message
+
+    * `send` and `drop` are mutually exclusive
+    * `ignore` and `use_first` are mutually exclusive
+
+**Description**
+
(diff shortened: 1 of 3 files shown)


creideiki:
I have no idea how to test this functionality without a Splunk instance filled with live data, so there are currently no tests.

docs/user/bots.rst
```diff
@@ -2611,6 +2611,75 @@ Use the following command to validate your sieve files:
...
+  * `warn`: log a warning message
+  * `use_first`: use the first search result
+  * `ignore`: do not modify the event
+  * `send`: send the event on
+  * `drop`: drop the message
+
+    * `send` and `drop` are mutually exclusive
+    * `ignore` and `use_first` are mutually exclusive
+
+* `overwrite`: Boolean or absent, whether search results overwrite values already in the message or not. If not specified, attempting to add a field that already exists throws an exception.
```
wagner-certat, line 2667: please add the default value

intelmq/bots/experts/splunk_saved_search/expert.py
```diff
@@ -0,0 +1,192 @@
...
+        if "send" in self.duplicates and "drop" in self.duplicates:
+            raise ConfigurationError("Processing", "Cannot both drop and send messages with duplicate search results")
+        if "ignore" in self.duplicates and "use_first" in self.duplicates:
+            raise ConfigurationError("Processing", "Cannot both ignore and use duplicate search results")
+
+        self.overwrite = getattr(self.parameters, "overwrite", None)
+
+        self.set_request_parameters()
+
+        self.http_header.update({"Authorization": f"Bearer {self.auth_token}"})
```
wagner-certat, line 103: This means that Python 3.5 is not supported by the bot. We need to decide if we drop Python 3.5 in the next IntelMQ release as more and more new contributions use format strings and therefore require Python 3.6

cc @schacht-certat

intelmq/bots/experts/splunk_saved_search/expert.py
```diff
@@ -0,0 +1,192 @@
...
+    def update_event(self, event, search_result):
+        self.logger.info("Updating event: %s",
+                         dict([(field, search_result[field]) for field in self.result_fields]))
+        for result, field in self.result_fields.items():
+            event.add(field, search_result[result], overwrite=self.overwrite)
+
+    def process(self):
+        event = self.receive_message()
+
+        self.logger.info("Received event, searching for %s",
```
wagner-certat, line 117: Please use debug log level here.

intelmq/bots/experts/splunk_saved_search/expert.py
```diff
@@ -0,0 +1,192 @@
...
+                req.raise_for_status()
+
+        hits = req.json()['results']
+        if len(hits) == 0:
+            if "warn" in self.not_found:
+                self.logger.warning("No results returned")
+            if "drop" in self.not_found:
+                self.logger.debug("Dropping message")
+                self.acknowledge_message()
+            if "send" in self.not_found:
```
wagner-certat, line 169: This can be an `elif`

intelmq/bots/experts/splunk_saved_search/expert.py
```diff
@@ -0,0 +1,192 @@
...
+                self.logger.warning("Multiple results returned: %s", hits)
+            if "use_first" in self.duplicates:
+                self.logger.debug("Using first search result")
+                self.update_event(event, hits[0])
+            if "ignore" in self.duplicates:
+                self.logger.debug("Ignoring search results")
+            if "drop" in self.duplicates:
+                self.logger.debug("Dropping message")
+                self.acknowledge_message()
+            if "send" in self.duplicates:
```
wagner-certat, line 183: can be `elif` as well

wagner-certat (changes requested):
Please add a link to Splunk documentation on saved searches to the bot docs

> I have no idea how to test this functionality without a Splunk instance filled with live data, so there are currently no tests.

Yeah, that would be tricky I guess.

creideiki:
> Please add a link to Splunk documentation on saved searches to the bot docs

Added. Note that the documentation is headlined "reports", and only mentions saved searches in passing. The terminology it uses is unclear, but a report is a presentation of the result of a saved search, with an optional schedule for running the saved search periodically.

docs/user/bots.rst
```diff
@@ -2611,6 +2611,75 @@ Use the following command to validate your sieve files:
...
+* `public`: no
+* `cache (redis db)`: none
+* `description`: Enrich an event from Splunk search results.
+
+**Configuration Parameters**
+
+* **HTTP parameters** (see above)
+* `auth_token`: String, Splunk API authentication token
+* `url`: String, base URL of the Splunk REST API
+* `retry_interval`: Integer, optional, default 5, number of seconds to wait between attempts to retrieve search results
```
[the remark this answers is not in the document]
creideiki: Changed, hopefully clarified.

docs/user/bots.rst
```diff
@@ -2611,6 +2611,75 @@ Use the following command to validate your sieve files:
...
+
+The saved search should take parameters according to the `search_parameters` configuration and deliver results according to `result_fields`. The examples above match a saved search of this format:
+
+::
+
+   index="dhcp" ipv4address="$ip$" | ... | fields _time username ether
+
+The time window used is the one saved with the search.
+
+Waits for Splunk to return an answer for each message, so slow searches will delay the entire botnet.
```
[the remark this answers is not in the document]
creideiki: I am not prepared to promise that multithreading works, but load balancing through multiprocessing shouldn't be a problem, so I added that.

docs/user/bots.rst
```diff
@@ -2611,6 +2611,75 @@ Use the following command to validate your sieve files:
...
+
+* `result_fields`: Array of string->string, optional, default ``{}``, Splunk search result fields mapped to IntelMQ event fields to store the results in. Example:
+
+  .. code-block:: json
+
+     "result_fields": {
+         "username": "source.account"
+     }
+
+* `not_found`: List of strings, default ``[ "warn", "send" ]``, what to do if the search returns zero results. Valid values are:
```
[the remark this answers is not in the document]
creideiki: Added.

docs/user/bots.rst
```diff
@@ -2611,6 +2611,75 @@ Use the following command to validate your sieve files:
...
+
+  * `warn`: log a warning message
+  * `send`: send the event on unmodified
+  * `drop`: drop the message
+
+    * `send` and `drop` are mutually exclusive
+
+* `duplicates`: List of strings, default ``[ "warn", "use_first", "send" ]``, what to do if the search returns more than one result. Valid values are:
```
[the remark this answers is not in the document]
creideiki: I cannot thing of a better single word, and would prefer not to have to extend configuration parameters to phrases. But I'm not a native speaker, so if you have a better suggestion, go ahead.

There is [a "head" command](https://docs.splunk.com/Documentation/Splunk/latest/SearchReference/Head) that limits the return to a number of results, but I had never heard of it before your comment made me go look. My impression is that idiomatic Splunk would limit the number of results by constraining the time window of the search (which cannot be done through the API - [the documentation for running saved searches](https://docs.splunk.com/Documentation/Splunk/latest/SearchReference/Savedsearch) references the "time picker", a web UI widget with no API counterpart). My impression is also that Splunk is heavily weighted towards interactive use in the web UI, with the REST API being at best an afterthought.

I would prefer not having to limit my search results when writing saved searches, since that makes them less useful for interactive use or creating reports. It would probably be possible to have a configuration parameter to let the bot add a `| head 1` command at the end of the query.

docs/user/bots.rst
creideiki: Fixed. I have no idea why I thought JSON couldn't handle an explicit null value.

intelmq/bots/experts/splunk_saved_search/expert.py
creideiki: Changed to `"".format()`. Also changed another f-string in a log statement below to %-syntax.

intelmq/bots/experts/splunk_saved_search/expert.py
creideiki: Changed.

intelmq/bots/experts/splunk_saved_search/expert.py
creideiki: It can, however I find the code easier to read this way. If you strongly prefer `elif`, I'll change it.

intelmq/bots/experts/splunk_saved_search/expert.py
creideiki: Same reply as above.

intelmq/bots/experts/splunk_saved_search/expert.py
```diff
@@ -0,0 +1,202 @@
...
+
+import intelmq.lib.utils as utils
+from intelmq.lib.bot import Bot
+from intelmq.lib.exceptions import MissingDependencyError, ConfigurationError
+import time
+
+
+class SplunkSavedSearchBot(Bot):
+
+    is_multithreadable = False
```
wagner-certat, line 72: Why is it not multithreadable?

intelmq/bots/BOTS
```diff
@@ -952,6 +952,25 @@
...
+                "saved_search": "search_name",
+                "search_parameters": {
+                    "event field": "search parameter"
+                },
+                "result_fields": {
+                    "result field": "event field"
+                },
+                "not_found": [ "warn", "send" ],
+                "duplicates": [ "warn", "use_first", "send" ],
+                "overwrite": true
```
wagner-certat, line 971: For consistency I suggest to change this to null. That's what is documented in the bots.rst and the default value for overwrite in other bots as well

wagner-certat (changes requested):

intelmq/bots/BOTS
creideiki: Changed.

intelmq/bots/experts/splunk_saved_search/expert.py
creideiki: Everything in imperative languages is not multithreadable until proven otherwise, and given that the bot object holds state between invocations of `process()` I'm not comfortable enough with Python threading to promise anything.

wagner-certat: Where does the bot hold a state?

creideiki: `self.session` is created in `init()` and then continually reused in `process()`.

wagner-certat: I looks like the question if the requests session is thread-safe is not so easy to answer
- https://stackoverflow.com/questions/18188044/is-the-session-object-from-pythons-requests-library-thread-safe
- https://github.com/psf/requests/issues/1871
- https://github.com/psf/requests/issues/2766

I have read a bit, bot definitely not everything and it's not clear to me yet if the relevant issues affects us here. Anyway, that's a topic for a separate issue and does not need to be solved here. Thanks for pointing this out, that affects a few other bots as well.

docs/user/bots.rst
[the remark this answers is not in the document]
creideiki: I tested, and blindly adding `| head 1` at the end of the query works, so I added an option to specify that to guarantee there will be no duplicate results. That way, you can have a saved search that returns an arbitrary number of results interactively, but just one when run from the bot.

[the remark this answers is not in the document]
wagner-certat: > I cannot thing of a better single word, and would prefer not to have to extend configuration parameters to phrases. But I'm not a native speaker, so if you have a better suggestion, go ahead.

What do you think about `duplicate_handling`?



> I tested, and blindly adding `| head 1` at the end of the query works, so I added an option to specify that to guarantee there will be no duplicate results. That way, you can have a saved search that returns an arbitrary number of results interactively, but just one when run from the bot.

Nice!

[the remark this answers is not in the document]
creideiki: I find `multiple_result_handling` to be clearer for a human reading the configuration, but less elegant, especially when used as a variable name in the code. But I added a commit changing the name, so you can see which one you prefer.

wagner-certat (approved):

wagner-certat:
Thank you very much again! :)

PR #1667, opened 2020-11-26
Docker documentation

waldbauer-certat:
All docker related changes in documentation.

### DIFF
--- a/docs/user/installation.rst
+++ b/docs/user/installation.rst
@@ -20,6 +20,7 @@ Supported and recommended operating systems are:
 * Debian 9 and 10
 * OpenSUSE Leap 15.1, 15.2
 * Ubuntu: 16.04, 18.04, 20.04
+* Docker Engine: 18.x and higher
 
 Other distributions which are (most probably) supported include CentOS 8, RHEL, Fedora and openSUSE Tumbleweed.
 
@@ -67,6 +68,17 @@ Optional dependencies:
 
    zypper in bash-completion jq
    zypper in python3-psycopg2 python3-pymongo python3-sleekxmpp
+
+Docker (beta)
+^^^^^^^^^^^^^
+
+**ATTENTION** Currently the version published on docker hub is not tagged with the same IntelMQ version. During beta, each version is published with tag `1.0`.
+**ATTENTION** Currently you cant manage your botnet via :doc:`intelmqctl`. You need to use `IntelMQ-Manager <https://github.com/certtools/intelmq-manager>`_ currently!
+
+Follow `Docker Install <https://docs.docker.com/engine/install/>`_ and 
+`Docker-Compose Install <https://docs.docker.com/compose/install/>`_ instructions.
+
+The latest image is hosted on `Docker Hub <https://hub.docker.com/r/certat/intelmq-full>`_
 
 Installation
 ------------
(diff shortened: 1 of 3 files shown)


docs/user/installation.rst
```diff
@@ -68,6 +69,17 @@ Optional dependencies:
    zypper in bash-completion jq
    zypper in python3-psycopg2 python3-pymongo python3-sleekxmpp
 
+Docker (beta)
+^^^^^^^^^^^^^
+
+**ATTENTION** Currently docker version can differ from intelmq version namings.
```
schacht-certat, line 75: I'm not sure I get the meaning of that sentence. Docker version is for example 1.5 and IntelMQs version is 2.2 - do you mean the version of the docker container does not follow the same version schema as IntelMQ does? Then I would write something like `Currently the version of the docker container published on docker hub is not tagged with the same version as the IntelMQ release it contains. It is tagged with version foo, but contains IntelMQ version bar.`

docs/user/installation.rst
```diff
@@ -68,6 +69,17 @@ Optional dependencies:
...
+**ATTENTION** If you use Docker you cant access :doc:`intelmqctl` via cli. You need to use `IntelMQ-Manager <https://github.com/certtools/intelmq-manager>`_ currently!
+
```
schacht-certat, line 77: s/access/run
s/via/from the
and I would specify: You can't manage you IntelMQ botnet (which is inside the container) using `intelmqctl` from outside the container.

docs/user/installation.rst
```diff
@@ -121,6 +133,53 @@ PyPi
...
+Docker **with** docker-compose (recommended)
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Navigate to your preferred installation directory, i. e. use ``mkdir ~/intelmq && cd ~/intelmq``
+
+.. code-block:: bash
+
+   git clone https://github.com/certat/intelmq-docker.git
+
+   sudo docker pull certat/intelmq-full:1.0
```
schacht-certat, line 145: doesn't docker compose automatically pull the image when we do `docker-compose up`?

docs/user/installation.rst
```diff
@@ -121,6 +133,53 @@ PyPi
...
+
+   mkdir intelmq_logs
```
schacht-certat, line 147: If we do `mkdir intelmq_logs` in the parent directory of the repository, but then run `docker-compose up` *in* the repository, the container won't be able to access the `intelmq_logs` folder, because the `docker-compose.yml` file points to `./intelmq-logs` which means the folder should be in the directory the `docker-compose.yml` file resides in. It seems docker-compose creates that directory with root permissions if it does not exist, therefore I would simply ship an empty `intelmq_logs` directory in the git repository.

When testing this I also had to change the image path in the `docker-compose.yml` file from `intelmq-full:1.0` to `certat/intelmq-full:1.0`.

docs/user/installation.rst
waldbauer-certat: This is only the case during the beta phase. After the beta is intended to follow the original intelmq versioning.

You can find a good reference here https://hub.docker.com/_/alpine or https://hub.docker.com/_/mariadb

docs/user/installation.rst
```diff
@@ -121,6 +133,53 @@ PyPi
...
+
+   cd intelmq-docker
+
+   chown -R $USER:$USER example_config
```
schacht-certat, line 151: This is only needed if the directory `example_config` does not exist. In that case docker-compose (which is run as root) creates the directory with root permissions. Given that by default the permissions should be oke, but apparently there can occur permission problems in some situations, I would add a sentence below: "Note: If you get a `Permission denied` error from one of the containers, please change the permissions back using `chown -R $USER:$USER example_config`"

docs/user/installation.rst
waldbauer-certat: With docker pull we force to pull the latest image from dockerhub.

If you only wanna work with docker-compose, I would recommend using `docker-compose pull`.

[Docker-compose up not pulling](https://github.com/docker/compose/issues/3574)

docs/user/installation.rst
waldbauer-certat: 1) Done.
2) Well, if you want to build your own image you should not use `certat/` namespace. Image should be built without any namespaces, i. e. `intelmq-full:1.0`

waldbauer-certat: Changed `intelmq-full:1.0` to `certat/intelmq-full:1.0` as default. Added additional developer note in README

schacht-certat (changes requested):

PR #1669, opened 2020-12-01
Generic CSV Parser: Ignore tabs and spaces

kbrajneesh:
Need to ignore line having spaces or tabs only or comment having leading tabs or spaces.


wagner-certat:
Thanks! Also cherry-picked for the bugfix release 2.2.3

PR #1670, opened 2020-12-02
Remediate some warnings

schacht-certat:
- BUG: Replace dns.query() with dns.resolve()
- BUG: Close the request session after its use
- TST: use a context manager for the mocked files

### DIFF
--- a/intelmq/bots/experts/national_cert_contact_certat/expert.py
+++ b/intelmq/bots/experts/national_cert_contact_certat/expert.py
@@ -51,6 +51,7 @@ class NationalCERTContactCertATExpertBot(Bot):
                     'sep': 'semicolon',
                 }
                 req = self.session.get(URL, params=parameters)
+                self.session.close()
                 if not req.text:
                     # empty response
                     continue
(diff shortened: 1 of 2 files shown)


intelmq/bots/experts/abusix/lib.py
```diff
@@ -22,7 +22,7 @@ def query(ip):
         query = rev + QUERY_HOST
 
         try:
-            response = dns.resolver.query(query, 'TXT')
+            response = dns.resolver.resolve(query, 'TXT')
```
wagner-certat, line 25: That method is new in dnspython 2.0.0 (see the changelog https://dnspython.readthedocs.io/en/stable/whatsnew.html). I'm therefore not sure if we should break compatibility with the old versions.

schacht-certat: Yeah, you're right. I've removed that commit.

wagner-certat: Once the compatibility layer gets removed, we'd need to support both dnspython 1.x and >= 2.x by using different function names, or a compat layer.

wagner-certat (approved):

PR #1673, opened 2020-12-03
A couple of bug fixes and updates of the Debian package

schacht-certat:
d/changelog:
  * Fix bogus-mail-host-in-debian-changelog
    https://lintian.debian.org/tags/bogus-mail-host-in-debian-changelog.html
  * Fix depends-on-essential-package-without-using-version
    This referres to `debianutils`. See
    https://lintian.debian.org/tags/build-depends-on-essential-package-without-using-version.html
    "The only reason to list an explicit dependency on an essential package
    is if you need a particular version of that package"
  * Fix: description-starts-with-package-name
    https://lintian.debian.org/tags/description-starts-with-package-name.html
    This also adds a synopsis, because according to policy, a Debian
    package description consists 'of two parts, the synopsis or the
    short description, and the long description'
  * Fix: syntax-error-in-debian-changelog
    There was a changed-by line missing, this got somehow introduced in
    eb6b38091d865f5e36920df57984eec638fed7a0; not a big deal, but fixing
    it makes lintian happy.
 d/control:
  * Run wrap-and-sort on d/control
  * Use sphinx and the sphinx rdt theme to build and manage
    documentation. This adds the necessary dependencies to the control
    file
  * Replaces the old style debian/compat file with a build-dependency on
    debhelper-compat.
 d/rules:
  * Enable the sphinx debhelper
    The --with systemd argument for debhelper is no longer provided in
    compat >= 11 - Debian packages use dh_installsystemd nowadays
    This also removes the systemd override in d/rules, because this is
    now default behaviour since compat 10.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,22 +2,58 @@ Source: intelmq
 Maintainer: Sebastian Wagner <wagner@cert.at>
 Section: python
 Priority: optional
-Build-Depends: debhelper (>= 4.1.16), python3-all, python3-setuptools, quilt, dh-python, python-setuptools, dh-systemd, safe-rm, python3-requests, python3-redis, python3-dnspython, python3-psutil, python3-dateutil, python3-termstyle, python3-tz, lsb-release
+Build-Depends: debhelper (>= 4.1.16),
+               dh-python,
+               dh-systemd,
+               lsb-release,
+               python-setuptools,
+               python3-all,
+               python3-cerberus,
+               python3-dateutil,
+               python3-dnspython,
+               python3-yaml,
+               python3-psutil,
+               python3-redis,
+               python3-requests,
+               python3-requests-mock,
+               python3-setuptools,
+               python3-sphinx,
+               python3-sphinx-rtd-theme,
+               python3-termstyle,
+               python3-tz,
+               quilt,
+               safe-rm
 X-Python3-Version: >= 3.5
 Standards-Version: 3.9.6
 Homepage: https://github.com/certtools/intelmq/
 
 Package: intelmq
 Architecture: all
-Depends: ${misc:Depends}, ${python3:Depends},
- python3-dateutil (>= 2.5), python3-dnspython (>= 1.11.1),
- python3-openssl, python3-psutil (>= 1.2.1), python3-redis (>= 2.10),
- python3-requests (>= 2.2.1), python3-termstyle (>= 0.1.10), python3-tz,
- redis-server, cron, bash-completion, jq, systemd, debianutils
-Suggests: python3-imbox (>= 0.8), python3-pyasn (>= 1.5.0),
- python3-stomp.py (>= 4.1.9), python3-sleekxmpp (>= 1.3.1),
- python3-geoip2 (>= 2.2.0), python3-pymongo, python3-psycopg2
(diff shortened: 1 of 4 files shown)


debian/control
```diff
@@ -2,22 +2,56 @@ Source: intelmq
 Maintainer: Sebastian Wagner <wagner@cert.at>
 Section: python
 Priority: optional
-Build-Depends: debhelper (>= 4.1.16), python3-all, python3-setuptools, quilt, dh-python, python-setuptools, dh-systemd, safe-rm, python3-requests, python3-redis, python3-dnspython, python3-psutil, python3-dateutil, python3-termstyle, python3-tz, lsb-release
+Build-Depends: debhelper-compat (= 13),
```
wagner-certat, line 5: With that requirement, we cannot build the package anymore on Ubuntu 16.04, 18.04, 20.04 and Debian 9, 10.

wagner-certat (changes requested):

debian/control
schacht-certat: I've changed that to `debhelper-compat (=10)` which should work for all of those, backports considered

debian/control
```diff
@@ -2,22 +2,56 @@ Source: intelmq
 Maintainer: Sebastian Wagner <wagner@cert.at>
 Section: python
 Priority: optional
-Build-Depends: debhelper (>= 4.1.16), python3-all, python3-setuptools, quilt, dh-python, python-setuptools, dh-systemd, safe-rm, python3-requests, python3-redis, python3-dnspython, python3-psutil, python3-dateutil, python3-termstyle, python3-tz, lsb-release
+Build-Depends: debhelper-compat (= 10),
```
wagner-certat, line 5: That doesn't work with Debian 9.0 and Ubuntu 16.04 now. If we gonna drop support for them, as they use python 3.5, that would be fine.

wagner-certat:
I've update the unstable repo to commit 1cc4f11 now to get all the build results: https://build.opensuse.org/package/show/home:sebix:intelmq:unstable/intelmq

debian/control
schacht-certat: Oke, reverted to `9`

schacht-certat:
I have reverted the change regarding `debhelper-compat` and added some more build-dependencies. The build should work on Debian 9 now, the only thing one has to do is to install `python3-cerberus` by hand.

wagner-certat (changes requested):
The Debian build fails on travis: https://travis-ci.org/github/certtools/intelmq/jobs/747603328#L501

> The build should work on Debian 9 now, the only thing one has to do is to install `python3-cerberus` by hand.

Cerberus is only needed at build time for the tests, not at runtime. So it shouldn't be an issue

I updated the unstable package to commit 01af64fcc: https://build.opensuse.org/package/show/home:sebix:intelmq:unstable/intelmq

wagner-certat:
> The Debian build fails on travis: https://travis-ci.org/github/certtools/intelmq/jobs/747603328#L501

Maybe using GitHub Actions, like for intelmq-api, works around this issue? I don't see this failure on OBS.
While Travis uses the command `dpkg-buildpackage -us -uc -d`, the GitHub Action uses `dpkg-buildpackage -us -uc -b`

wagner-certat (approved):

PR #1674, opened 2020-12-07
[FIX] Issue #1651

### LINKED ISSUE #1651: CSV Parser not treating ignored values correctly leading to exception
If the generic csv parser encounters a `-` as value in a column, the data is rejected by:
```
  File "/usr/lib/python3/dist-packages/intelmq/lib/bot.py", line 978, in process
    events = list(filter(bool, value))
  File "/usr/lib/python3/dist-packages/intelmq/bots/parsers/generic/parser_csv.py", line 130, in parse_line
    raise InvalidValue(key, value)
intelmq.lib.exceptions.InvalidValue: invalid value '-' (<class 'str'>) for key 'extra.cert_orgunit'
```

Example data is as simple as:
```
someheader
-
```

The reason is in these lines of code:
https://github.com/certtools/intelmq/blob/4d10b2bcfd932b4ae3275e0df00f83325255e574/intelmq/bots/parsers/generic/parser_csv.py#L129-L130

If `Message.add` gets a value to be ignored (like `-`), the return value is `None`, which evaluates to False. I think changing the if-clause to `... is not False` is enough.

Please add a testcase for this.

waldbauer-certat:
Currently CsvParser handles "-" values incorrectly, throwing an error, which was not intended. 
As @wagner-certat already mentioned, the fix is to check `is not False`.

Fixes #1651


intelmq/tests/bots/parsers/generic/test_parser_csv_invalid_values.py
```diff
@@ -45,13 +49,13 @@ def set_bot(cls):
...
 
-    def test_error(self):
-        """ Test if the error is raised. """
+    def test_event(self):
+        """ Test if parser treatens ignored values correctly """
         self.allowed_error_count = 1
         self.run_bot()
-        self.assertMessageEqual(0, EXAMPLE_EVENT)
-        self.assertLogMatches("Failed to parse line.")
-        self.assertLogMatches("intelmq.lib.exceptions.InvalidValue: invalid value '-' \(<class 'str'>\) for key 'source.fqdn'")
```
wagner-certat, line 54: I'm not comfortable removing this check. We can use the invalid FQDN value `.` for testing this error behavior.

wagner-certat (changes requested):

intelmq/tests/bots/parsers/generic/test_parser_csv_invalid_values.py
waldbauer-certat: Done

Issue #1614, opened 2020-10-05
Replace sleekxmpp with slixmpp


schacht-certat:
Hi, the sleekxmpp repository states: "SleekXMPP is deprecated in favor of Slixmpp, a fork which takes full advantage of Python 3 and asyncio."

SleekXMPP is used for an xmpp output and an xmpp collector, they should be migrated to slixmpp (https://lab.louiz.org/poezio/slixmpp)

cheers,
Birger

wagner-certat:
I'll ask on the mailing list if someone is using this bot before we put our efforts in adapting it.

wagner-certat:
https://lists.cert.at/pipermail/intelmq-users/2020-October/000177.html

PR #1682, opened 2020-12-14
Remove the XMPP collector and output bots

schacht-certat:
This was evaluated on the mailinglist in
https://lists.cert.at/pipermail/intelmq-users/2020-October/000177.html
This removes both the XMPP collector bot and the XMPP output bot,
any documentation regarding those bots and the test for those bots.
It also updates installation instructions to not list python-sleekxmpp
as dependency.

Closes: #1614


wagner-certat (changes requested):
Please mark them as deprecated before we completely remove the code in a future version (i propose 3.0.0 as it is a breaking change). See for example the postgresql output bot, which is deprecated as well.

Please also add a note to the changelog and the news file as well as the `bots.rst`

PR #1684, opened 2020-12-14
New Parser for JSON data

kbrajneesh:
This parser can translate json key into harmonized key field and can create events from list of dict within key of JSON.


docs/user/bots.rst
```diff
@@ -1243,6 +1243,38 @@ HTML Table Parser
...
+* `"json_data_format"`: Boolean, if list of data is within key of json object, optional. Default: false.
+* `"json_data_key"`: Key of json object where data list is present. string should be flatten_key, optional. To be used in conjunction with `"json_data_format"`. Default: `""`.
+   E.g.
+
+  .. code-block:: json
+
+     "json_data_format": true,
+     "json_data_key": "data.ipdata"
+
+  With above configuration, list of dict will be created from list present in json["data"]["ipdata"]. Each dict will then create atleast an event.
```
wagner-certat, line 1260: What if the data is in `json["data.ipdata"]`?

docs/user/bots.rst
```diff
@@ -1243,6 +1243,38 @@ HTML Table Parser
  * `"type"`: set the `classification.type` statically, optional
  * `"html_parser"`: The HTML parser to use, by default "html.parser", can also be e.g. "lxml", have a look at https://www.crummy.com/software/BeautifulSoup/bs4/doc/
 
+JSON Custom Parser
+^^^^^^^^^^^^^^^^^
+
+**Configuration parameters**
+
+* `"json_data_format"`: Boolean, if list of data is within key of json object, optional. Default: false.
```
wagner-certat, line 1251: Please expand the explanation, examples are also helpful.

docs/user/bots.rst
```diff
@@ -1243,6 +1243,38 @@ HTML Table Parser
...
+* `"json_data_key"`: Key of json object where data list is present. string should be flatten_key, optional. To be used in conjunction with `"json_data_format"`. Default: `""`.
+   E.g.
+
+  .. code-block:: json
+
+     "json_data_format": true,
+     "json_data_key": "data.ipdata"
+
+  With above configuration, list of dict will be created from list present in json["data"]["ipdata"]. Each dict will then create atleast an event.
+* `"splitlines"`: Boolean, spit multiline data into list, optional. Default: `"false"`. Either `"json_data_format"` or `"splitlines"` can be used.`
```
wagner-certat, line 1261: Please expand the explanation, examples are also helpful.

wagner-certat (changes requested):
Thanks for the contribution! That's definitely a long-awaited parser :)

I only had a look at the added documentation so far.

docs/user/bots.rst
kbrajneesh: json_data_format will be **True** if there is a single json and list of data is inside any of its child nodes. For below json, we find data inside key 'data'
`{
    "complete": false,
    "count": 9632240,
    "data": [
        {
            "ip": "179.124.36.196",
            "classification": "malicious",
            "first_seen": "2020-01-13",
            "last_seen": "2020-12-14",
            "tags": [
                "SSH Scanner",
                "SSH Worm"
            ],
            "cve": []
        },
        {
            "ip": "189.86.227.150",
            "classification": "malicious",
            "first_seen": "2019-01-17",
            "last_seen": "2020-12-14",
            "tags": [
                "Eternalblue",
                "SMB Scanner"
            ],
            "cve": [
                "CVE-2017-0144"
            ]
        }
    ],
    "message": "ok"
}`

docs/user/bots.rst
kbrajneesh: Again in above json, if we have config `json_data_key`: _data_ and `json_data_format `: _True_, it will create two event with list of data present in json["data"] with proper harmonized field as per given in `translate_field` parameter.

docs/user/bots.rst
kbrajneesh: it will still work. if we have both json["data"]["ipdata"] and json["data.ipdata"], the one which is later defined, will override the other and its value will be assigned in key "data.ipdata" of flattened json.

schacht-certat:
Hi, thanks a lot for your contribution! That looks like a very useful parser. Could you make your commit messages a bit more verbose? It would be great if the information you added in the pull request would be part of the commit message, i.e. 
```
Add a new JSONCustom parser Bot

JSON Custom Parser allows can translate json key into harmonized key field 
and can create events from list of dict within key of JSON
```
and for the documentation commit (if you want to keep that separate - you could also simply squash to one commit!)
```
DOC: add documentation for the JSON Custom Parser

The JSON Custom Parser converts from a custom JSON-String into an Event.
This adds an explanation and examples to the Bot documentation.
```

Also there are **a lot** of unnecessary changes in the `bots.rst` file, it seems your editor somehow tries to change spaces in the whole file? See https://github.com/certtools/intelmq/pull/1684/commits/35ff026a6c7f989b1a25aa7904146bb95e14208f - it would be great if that could be fixed.

kbrajneesh:
Hi, all requested changes have been made in previous commit.

docs/user/bots.rst
wagner-certat: If `json_data_key` can only be used in conjunction with `json_data_format`, why use two different parameters?

docs/user/bots.rst
wagner-certat: Why are `json_data_format` and `splitlines` mutually exclusive?

Btw, I think there's no test that covers actual line splitting (the tests with `splitlines=true` only have one line)

docs/user/bots.rst
wagner-certat: Can you please add that example to the docs?

sebix:
Looking at the mapping of values from source document to IntelMQ format, I see a difference to what the CSV parser uses.

The Generic CSV Parser maps $csvcolumnname -> $intelmqname 
That also allows to define fallback fields, eg if the data is not a URL, try it as hostname, if that fails too try it as IP address.

I think, the JSON parser could use the same scheme too, for consistency and future additions could add this fallback (alternative mappings) feature.

PR #1708, opened 2021-01-11
Fixes sieve timezone "utc" & sieve variable bug

waldbauer-certat:
Timezone was not interpreted correctly. Now we localize time to UTC in sieve expert.

Signed-off-by: Sebastian Waldbauer <waldbauer@cert.at>


wagner-certat:
It works and it's already deployed.

*But* maybe this fix is better placed in `parse_utc_isoformat` to fix the issue in the core? Otherwise every bot using this function would need the extra lines as well.

PR #1709, opened 2021-01-11
remove undesired quotes in harmonization.conf

tomas321:
change unicode right/left double quotes to utf-8 (imho as it is intend to)

`event.classification.identifier.description` in harmonization.conf served as an example to correct usage.

### DIFF
--- a/intelmq/etc/harmonization.conf
+++ b/intelmq/etc/harmonization.conf
@@ -10,7 +10,7 @@
             "type": "LowercaseString"
         },
         "classification.type": {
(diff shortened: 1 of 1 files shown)


wagner-certat:
Just curious: Did this lead to any problems? (The file needs to be read with UTF-8, but this should be the case for all intelmq-code)

tomas321:
To be honest, I cannot remember, but I did have to change it a while ago (that's why I don't remember). Now when I encountered the change and saw the similar chars in the harmonization.conf, I thought it's a small typo.

schacht-certat:
That would change how the text is rendered on https://intelmq.readthedocs.io/en/latest/dev/harmonization-fields.html - the double backticks are the syntax for literal text: https://docutils.sourceforge.io/docs/ref/rst/roles.html#literal

I would simply use one tick before the string and one tick after or one asterisk before and one after to emphasize.

schacht-certat:
This was implemented in e9640383286a7b8dbee6c23697a46e822275b222

PR #1723, opened 2021-01-29
[FIX] Unittest checks if rsync is installed

waldbauer-certat:
Currently this test will fail and throw an exception if rsync is not installed. So we should check first if rsync is installed, before running this test.

### DIFF
--- a/intelmq/tests/bots/collectors/rsync/test_collector.py
+++ b/intelmq/tests/bots/collectors/rsync/test_collector.py
@@ -2,6 +2,7 @@
 import tempfile
 import os
 import unittest
+import shutil
 
 import intelmq.lib.test as test
 import intelmq.lib.utils as utils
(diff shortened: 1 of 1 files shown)


intelmq/tests/bots/collectors/rsync/test_collector.py
```diff
@@ -17,7 +17,7 @@
           "raw": utils.base64_encode(EXAMPLE_FILE),
           }
 
-
+@unittest.skipIf(not os.path.exists("/opt/rsync"), "RSync is not installed")
```
schacht-certat, line 20: On Linux commands are not usually installed on `/opt`, but in `/bin` or `/usr/bin`

schacht-certat (changes requested):

intelmq/tests/bots/collectors/rsync/test_collector.py
wagner-certat: What about using the return value of `type which`? Should be rather universal

schacht-certat:
LGTM, but you might want to make the error message more correct- i.e. 'rsync command not found'
(rsync can still be installed but it might be installed somewhere that's not in the PATH of the Python interpreter)

schacht-certat (approved):

PR #1728, opened 2021-02-04
Shadowserver Block List Feed: Parse CIDR-notation entries correctly

waldbauer-certat:
ShadowServer parses not CIDR notation networks correctly.

#Fixes 1720


wagner-certat (changes requested):
pycodestyle check fails:
intelmq/bots/parsers/shadowserver/config.py:198:1: E302 expected 2 blank lines, found 1
intelmq/bots/parsers/shadowserver/config.py:200:8: E713 test for membership should be 'not in'

wagner-certat (approved):

PR #1749, opened 2021-02-18
Insert date when email was received

aleksejsv:
Sometimes we receive email reports like `this is happening right now` and there is no date/time included. So if we process emails once per hour - we don't have info about event time. Additional field `extra.email_received` in the mail body collector would help.

### DIFF
--- a/intelmq/bots/collectors/mail/collector_mail_body.py
+++ b/intelmq/bots/collectors/mail/collector_mail_body.py
@@ -29,6 +29,7 @@ class MailBodyCollectorBot(MailCollectorBot):
                 report["extra.email_subject"] = message.subject
                 report["extra.email_from"] = ','.join(x['email'] for x in message.sent_from)
                 report["extra.email_message_id"] = message.message_id
+                report["extra.email_received"] = message.date
 
                 self.send_message(report)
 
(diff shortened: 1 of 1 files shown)


intelmq/bots/collectors/mail/collector_mail_body.py
```diff
@@ -29,6 +29,7 @@ def process_message(self, uid, message):
                 report["extra.email_subject"] = message.subject
                 report["extra.email_from"] = ','.join(x['email'] for x in message.sent_from)
                 report["extra.email_message_id"] = message.message_id
+                report["extra.email_received"] = message.date
```
wagner-certat, line 32: Is `message.date` really the *reception* date, not the sending date?

wagner-certat:
Hm, could we use the `time.observation` field for this? Then it would match the behaviour of other collectors (e.g. the RT collector) to hold the date when the message reached the target system.

aleksejsv:
Isn't `time.observation` the time when event was processed by IntelMQ? `message.date` is more like `time.source` in this case - when event has occurred. Later in the parser I do `event.add('time.source', ts)`, and setting `time.source` automatically by the collector is a bad idea I guess.

intelmq/bots/collectors/mail/collector_mail_body.py
aleksejsv: I rechecked the `imbox` library you use for emails. It seems it just takes mail header field `Date`, so you should be right

aleksejsv: Renamed field to the `extra.email_date`

wagner-certat (approved):

wagner-certat:
Thanks! I extended the change a bit in 802e77fe5668a08b31180a5aeb1f7d68bef071f2 (add it to all mail collectors, update docs and tests).

PR #1751, opened 2021-02-22
Generate the BOTS list on demand

### LINKED ISSUE #368: Where bot configs should be stored in repo?

### LINKED ISSUE #552: Enable separate packaging of bots by allowing addition and removals to the config.

### LINKED ISSUE #644: Errors when already configured bots gain additional options through upgrade

### LINKED ISSUE #668: Very long BOTS file

### LINKED ISSUE #757: Clearly define all parameters used in a bot

### LINKED ISSUE #972: Make IntelMQ plugable
We've already discussed this some times in the dev - calls but I couldn't find an issue which states this demand.

From my point of view  IntelMQ will get better when using a plug-able structure, allowing to add own plug-ins with more ease.

I try to start anew with a collection of already existing Issues and code which are / is covering this topic.

 -  Is #552 
 -  Is #368 
 -  Is #757 
 -  Is #668 
 -  Replace-Bots branch of @sykaeh https://github.com/sykaeh/intelmq/tree/replace-bots https://github.com/certtools/intelmq/issues/757#issuecomment-270442055
 -  ML http://lists.cert.at/pipermail/intelmq-dev/2016-April/000033.html

I understood that we've already reached some kind of consensus that this structure is something we want to pursue. So let's just start to discuss and collect what's required to achieve this goal.

Maybe we can start by collecting some, even fictional, requirements:

- _As an operator of an IntelMQ instance, I'd like to add my own code for bots which processes internal information without the need of manually editing the bots file._
- _As an admin I want to install IntelMQ without the need of installing libraries which are not required for my use-case. Code for bots that is not necessary should not exist on my systems._
 

_please continue..._

### LINKED ISSUE #1069: BOTS is not a config file -> /var/lib
Currently it resides in `/etc/intelmq/` or `/opt/intelmq/etc/` but it's not a config file. It should rather be in `/var/lib/intelmq/` or `/opt/intelmq/var/lib/`.

Could be solved together with #552 or earlier, together with a symlink for backwards compatibility.

### LINKED ISSUE #1750: Get rid of the BOTS file
Removing the `BOTS file` and generate it on demand using i. e. `intelmqctl list bots`.

waldbauer-certat:
Instead of using a BOTS file on disk, we are now able to generate it on demand.

Closes #1750 
Closes #1069 
Closes #972 
Closes #757 
Closes #668
Closes #644 
Closes #552 
Closes #368


wagner-certat:
Please add in the commit message and a changelog entry that this fixes the following issues:

#1750
#1069
#972
#757
#668
#644
#552
#368

intelmq/bin/intelmqctl.py
```diff
@@ -1219,11 +1221,11 @@ def botnet_status(self, group=None):
                 retval = 1
         return retval, botnet_status
 
-    def list(self, kind=None, non_zero=False, count=False):
+    def list(self, kind=None, non_zero=False, count=False, configured=False):
```
wagner-certat, line 1224: The list of parameters for the `list` command gets a bit clunky and confusing now. I think we should split the `list` subcommand (maybe not part of this step, but soon)

intelmq/lib/utils.py
```diff
@@ -808,3 +812,51 @@ def file_name_from_response(response: requests.Response) -> str:
...
+    bot_types = [
+        'CollectorBot',
+        'ParserBot',
+        'OutputBot',
+        'Bot'
+    ]
+    root = mod.__path__[0]
+    root_dict = {}
+
+    for d in dirs:
```
schacht-certat, line 830: please don't use single character variables, that does not help the readability of the code at all. Just write `for directory in directories`

intelmq/lib/utils.py
```diff
@@ -808,3 +812,51 @@ def file_name_from_response(response: requests.Response) -> str:
     except KeyError:
         file_name = response.url.split("/")[-1]
     return file_name
+
+
+def list_all_bots() -> str:
+    mod = importlib.import_module('intelmq.bots')
+    dirs = ['collectors', 'parsers', 'experts', 'outputs']
+    allowed_types = ['bool', 'NoneType', 'str', 'int']
```
schacht-certat, line 820: Why do we have to keep a list of allowed types for configuration settings?

intelmq/lib/utils.py
```diff
@@ -808,3 +812,51 @@ def file_name_from_response(response: requests.Response) -> str:
...
+    root = mod.__path__[0]
+    root_dict = {}
+
+    for d in dirs:
+        root_dict[d.capitalize()] = {}
+        for dirs in os.listdir(os.path.join(root, d)):
+            bot = Path(os.path.join(root, d, dirs))
+            if bot.is_dir() and "__" not in bot.name:
+                for files in os.listdir(bot):
+                    f = Path(files)
```
schacht-certat, line 836: Please don't use single character variable names. Its not clear at all what `f` should contain

intelmq/lib/utils.py
```diff
@@ -808,3 +812,51 @@ def file_name_from_response(response: requests.Response) -> str:
...
+        'OutputBot',
+        'Bot'
+    ]
+    root = mod.__path__[0]
+    root_dict = {}
+
+    for d in dirs:
+        root_dict[d.capitalize()] = {}
+        for dirs in os.listdir(os.path.join(root, d)):
+            bot = Path(os.path.join(root, d, dirs))
```
schacht-certat, line 833: You're mixing `pathlib` and `os` at will, you can simple concatenate `pathlib` paths with `/` and iterate over directories using `iterdir`.
You could use something like this to get all the python files in the bot directories:
```
botfiles = [botfile for botfile in pathlib.Path('intelmq/bots').glob('**/*.py') if botfile.is_file() and botfile.name != '__init__.py']
```

Then you have a flat list of python files as `Path` objects and you don't have to use multiple nested for loops, which makes the code way more readable

intelmq/lib/utils.py
```diff
@@ -808,3 +812,51 @@ def file_name_from_response(response: requests.Response) -> str:
...
+                                if prop.startswith("_"):
+                                    continue
+
+                                if type(getattr(mod.BOT, prop)).__name__ not in allowed_types:
+                                    continue
+
+                                keys[prop] = val
+
+                            for t in bot_types:
+                                if name == t:
```
schacht-certat, line 853: That can be written as `if name in bot_types`

intelmq/lib/utils.py
```diff
@@ -808,3 +812,51 @@ def file_name_from_response(response: requests.Response) -> str:
...
+        'CollectorBot',
+        'ParserBot',
+        'OutputBot',
+        'Bot'
+    ]
+    root = mod.__path__[0]
+    root_dict = {}
+
+    for d in dirs:
+        root_dict[d.capitalize()] = {}
```
schacht-certat, line 831: can you use a meaningful name instead of `root_dict`?

intelmq/lib/utils.py
```diff
@@ -808,3 +812,51 @@ def file_name_from_response(response: requests.Response) -> str:
...
+        for dirs in os.listdir(os.path.join(root, d)):
+            bot = Path(os.path.join(root, d, dirs))
+            if bot.is_dir() and "__" not in bot.name:
+                for files in os.listdir(bot):
+                    f = Path(files)
+
+                    if f.suffix == '.py' and '__' not in f.name:
+                        mod = importlib.import_module("intelmq.bots.{}.{}.{}".format(d, bot.parts[-1:][0], f.name[0:-3]))
```
schacht-certat, line 839: If you want the filename from a `Path` object you can use `stem` - no need to manually slice the string.
Also isn't `d` already part of the `bot` path object- why use `d` directly in the format string instead of using the bot part?

intelmq/lib/utils.py
```diff
@@ -808,3 +812,51 @@ def file_name_from_response(response: requests.Response) -> str:
...
+                        if hasattr(mod, 'BOT'):
+                            name = mod.BOT.__name__
+                            keys = {}
+                            for prop, val in vars(mod.BOT).items():
+                                if prop.startswith("_"):
+                                    continue
+
+                                if type(getattr(mod.BOT, prop)).__name__ not in allowed_types:
+                                    continue
```
schacht-certat, line 848: Maybe its again easier to use a flat list instead of a for loop and if statements with continue:
```
bot = mod.BOT
variables = sorted(name for name, value in inspect.getmembers(bot) if not inspect.ismethod(value) and not inspect.isfunction(value) and not name.startswith('_'))
```
Maybe that's why there is the `allowed_types`?

schacht-certat (changes requested):

wagner-certat (changes requested):
Currently `intelmqctl list bots` returns an empty string instead of a human-readable bot index.

`BOTS_FILE` is still used in
- `intelmq/bin/intelmqsetup.py`
- `intelmq/__init__.py`

As the BOTS file is present in installations, but now obsolete, please also add a method to `intelmq/lib/upgrades.py`:
- if identical to the packaged version (use pkg_resources to get that file), the file can be removed
- if different, print an error/warning message to the admin (first item of the returned tuple can be a string).

The output has an arbitrary order it seems. Please sort the result - at least for parsers, experts and output this is possible. I'm not sure yet how to best solve the sorting issue for collectors.

The extraction of names has some weird results, e.g.:
```
    "ESMIP": {
      "module": "intelmq.bots.outputs.mcafee.output_esm_ip",
      "description": null,
      "parameters": {}
    },
```
That should be ideally "McAfee ESM IP"

And here:
```
    "MISPAPI": {
      "module": "intelmq.bots.outputs.misp.output_api",
      "description": null,
      "parameters": {
        "is_multithreadable": false
      }
    },
```
Name should rather be `MISP API` and `is_multithreadable` is not a parameter, so it needs to be renamed to `_is_multithreadable`

```
    "ShadowserverJSON": {
      "module": "intelmq.bots.parsers.shadowserver.parser_json",
      "description": "\n    Shadowserver JSON Parser\n\n    Parameters\n    ----------\n    feedname: str\n        The name of the feed\n\n    ",
      "parameters": {
        "reporttype_fn": null,
        "feedname": null,
        "sparser_config": null
      }
    },
```
For the description using [`textwrap.dedent`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) results in less newlines and whitespace.

Could you please rebase on develop? Then the result is more complete and we can better compare the result with the BOTS file.

intelmq/lib/utils.py
```diff
@@ -808,3 +814,46 @@ def file_name_from_response(response: requests.Response) -> str:
...
+                not inspect.isroutine(value) and
+                not name.isupper() and
+                not name.startswith('_')
+            )
+            for key, value in variables:
+                keys[key] = value
+
+            for t in bot_types:
+                if name == t:
+                    continue
```
schacht-certat, line 851: I don't think that is needed anymore, using the code above bots whose name are the same as one of the `bot_types` are not listed. Also `t` is not a meaningful variable name. Given that `bot_types` is not needed anywhere else in the method, you can also simpy write
```
# Some bots have their parent class names as part of their name,
# we are removing that to make the name more readable.
for bot_type in ['CollectorBot','ParserBot','OutputBot','Bot']:
    name = name.replace(bot_type, '')
```

intelmq/lib/utils.py
```diff
@@ -808,3 +814,46 @@ def file_name_from_response(response: requests.Response) -> str:
...
+                name = name.replace(t, '')
+
+            bots[file.parts[-3:][0].capitalize()][name] = {
```
schacht-certat, line 854: why use `file.parts[-3:][0]` instead of `file.parts[3]`?

intelmq/lib/utils.py
```diff
@@ -808,3 +814,46 @@ def file_name_from_response(response: requests.Response) -> str:
...
+                "module": mod.__name__,
+                "description": textwrap.dedent(mod.BOT.__doc__),
```
schacht-certat, line 856: This crashes if the description is missing.
```
"description": "Missing description" if mod.BOT.__doc__ is None else textwrap.dedent(mod.BOT.__doc__),
```

schacht-certat (changes requested):
If we change the variables to be constants, they should be moved to the module level. An alternative would be to make them private.

intelmq/lib/upgrades.py
```diff
@@ -531,6 +533,21 @@ def v230_csv_parser_parameter_fix_1(defaults, runtime, harmonization, dry_run):
                 changed = True
     return changed, defaults, runtime, harmonization
 
+def v300_bots_file_removal(defaults, runtime, harmonization, dry_run):
+    """
+    Remove BOTS file
+    """
+    changed = None
+    messages = []
+    bots_file = Path(resource_filename('intelmq', 'bots/BOTS'))
```
wagner-certat, line 542: That will resolve to the filename in the packaging source, not the installed one. Use the deleted code of `__init__.py` (`os.path.join(intelmq.CONFIG_DIR, "BOTS")`)

wagner-certat (changes requested):
> Currently `intelmqctl list bots` returns an empty string instead of a human-readable bot index.

Still applies.

`intelmq@medan:~> intelmqctl -t json list bots` now returns for me (with an additional rebase on develop):
```
{"Collectors": {}, "Parsers": {}, "Experts": {}, "Outputs": {}}
```

intelmq/lib/upgrades.py
```diff
@@ -588,8 +607,5 @@ def v230_feed_changes(defaults, runtime, harmonization, dry_run):
     ((2, 2, 1), (v221_feed_changes, )),
     ((2, 2, 2), (v222_feed_changes, )),
     ((2, 2, 3), ()),
-    ((2, 3, 0), (v230_csv_parser_parameter_fix_1, v230_feed_changes, v230_deprecations,)),
-    ((3, 0, 0), ()),
+    ((2, 3, 0), (v230_csv_parser_parameter_fix_1, )),
```
wagner-certat, line 610: There was a problem with the rebase here

intelmq/bots/experts/sieve/expert.py
```diff
@@ -85,8 +85,27 @@ def check(parameters):
...
+            if not metamodel:
+                raise TextXError('Metamodel not loaded.')
+
+            if not os.path.exists(parameters['file']):
+                raise FileExistsError('File does not exist: %r' % parameters['file'])
+
+            try:
+                metamodel.model_from_file(parameters['file'])
+            except TextXError as e:
+                raise TextXSemanticError('Could not process sieve file %r. Error in (%d, %d).' % (parameters['file'], e.line, e.col))
```
waldbauer-certat, line 108: Change TextXSemanticError to TextXError?

wagner-certat: If we are at it, `intelmq.lib.exceptions.ConfigurationError` could be better

intelmq/bots/experts/sieve/expert.py
```diff
@@ -85,8 +85,27 @@ def check(parameters):
...
+
+            metamodel = None
+
+            try:
+                metamodel = metamodel_from_file(grammarfile)
+            except TextXError as e:
+                raise TextXSemanticError('Could not process sieve grammar file. Error in (%d, %d).' % (e.line, e.col))
+
+            if not metamodel:
+                raise TextXError('Metamodel not loaded.')
```
wagner-certat, line 100: If I am not mistaken, this does not serve any purpose. (I know that it was in the original validator, but we can get rid of it)

intelmq/bots/experts/sieve/expert.py
```diff
@@ -85,8 +85,27 @@ def check(parameters):
...
+
+            if not os.path.exists(parameters['file']):
+                raise FileExistsError('File does not exist: %r' % parameters['file'])
```
wagner-certat, line 103: `FileExistsError`? It does not exist. We can use `intelmq.lib.exceptions.ConfigurationError` instead.

intelmq/lib/upgrades.py
```diff
@@ -571,6 +575,22 @@ def v230_feed_changes(defaults, runtime, harmonization, dry_run):
...
+    Remove BOTS file
+    """
+    changed = None
+    messages = []
+    bots_file = Path(os.path.join(CONFIG_DIR, "BOTS"))
+    if bots_file.exists():
+        bots_file.unlink()
+        changed = True
+    else:
+        messages.append("Can't find BOTS file, you have to remove it yourself.")
```
wagner-certat, line 589: In case the user already removed the file, the user would see the (error) message here.

But if `BOTS_FILE` does not exist, we do no longer need to remove it. So we can just remove the else-clause I think.

wagner-certat (changes requested):
Looks very good, we are near the goal :)

I leave the review of the utils and the bots for @schacht-certat :)

schacht-certat (changes requested):
Please get rid of the added `parse_docstring` method. We nearly had a mergable PR, now there are 20 more lines of code (which you yourself describe as *hacky* and *needing a lot of rework*). In addition to that, you try to introduce the idea of a `IntelMQ-Bot-Name` metadata field, which we discussed but there was not consensus about that. This PR should be about getting rid of the `BOTS` file and not introduce any additional ideas (which, given that no bot contains an `IntelMQ-Bot-Name` field at the moment, only adds complexity which no added functionality).

Please comment your code.

Please write meaningful commit messages. The [PR Template](https://github.com/certtools/intelmq/blob/develop/.github/pull_request_template.md) give hints about how to write good commit messages. A commit message simply saying `Docstringparser and BOTS list generator` with no other information does not explain what problems you are trying to solve. A commit message saying `Removed BOTS file, fixed upgrade` does not explain what bug was fixed in the upgrade. It can also help to list the bugs a commit closes, like @wagner-certat pointed out in https://github.com/certtools/intelmq/pull/1751#issuecomment-786867188

wagner-certat:
> Please get rid of the added `parse_docstring` method.

This is an attempt to get the bots' names right. Let's discuss this on Monday.

wagner-certat:
Tests and `intelmqctl list bots` fail

intelmq/lib/utils.py
```diff
@@ -808,3 +816,44 @@ def file_name_from_response(response: requests.Response) -> str:
     except KeyError:
         file_name = response.url.split("/")[-1]
     return file_name
+
+
+def list_all_bots() -> str:
```
wagner-certat, line 821: This function returns dict, not str

intelmq/lib/utils.py
```diff
@@ -808,3 +816,44 @@ def file_name_from_response(response: requests.Response) -> str:
...
+
+            for key, value in variables:
+                keys[key] = value
+
+            for bot_type in ['CollectorBot', 'ParserBot', 'OutputBot', 'Bot']:
+                name = name.replace(bot_type, '')
+
+            bots[file.parts[2].capitalize()[:-1]][name] = {
+                "module": mod.__name__,
+                "description": "Missing description" if __doc__ not in mod.BOT else textwrap.dedent(mod.BOT.__doc__),
```
wagner-certat, line 856: Fixes the fail:
```suggestion
                "description": "Missing description" if hasattr(mod.BOT, '__doc__') else textwrap.dedent(mod.BOT.__doc__),
```

wagner-certat:
IMHO we should migrate the descriptions in the BOTS file to the bot classes.

intelmq/lib/utils.py
```diff
@@ -808,3 +816,44 @@ def file_name_from_response(response: requests.Response) -> str:
...
+
+    botfiles = [botfile for botfile in pathlib.Path(base_path).glob('**/*.py') if botfile.is_file() and botfile.name != '__init__.py']
+    for file in botfiles:
+        file = Path(file.as_posix().replace(base_path, 'intelmq/bots'))
+        mod = importlib.import_module('.'.join(file.with_suffix('').parts))
+        if hasattr(mod, 'BOT'):
+            name = mod.BOT.__name__
+            keys = {}
+            variables = sorted(
+                (key, value) for key, value in vars(mod.BOT).items()
```
wagner-certat, line 839: `vars()` only returns the members of this class, excluding the parent classes. This is problem for the Bot classes and the mixins as well. `dir()`, which I proposed earlier, returns all members - but not a dict, just the names.

wagner-certat:
BOTS file needs to be removed then.
But we will then get the next error, as `setup.py` also accesses the BOTS file. Therefore it needs to be adapted, otherwise intelmq is not installable otherwise.

wagner-certat:
I fixed some issues in #1784 
But still `intelmq.bots.parsers.microsoft.parser_bingmurls` is not detected. Couldn't yet figure out why.

wagner-certat:
> My (hacky) idea for the setup.py fix. In utils.py add an optional parameter `base_path` to `list_all_bots`.
> In setup.py open utils.py, do regular search on `(def list_all_bots.*?)(def|$)`, save the first match with `re.MULTILINE`. Do `eval([necessary imports] $matchingstring, list_all_bots(base_path))`.
> 
> But that may not work because of all the imports with importlib? Otherwise I see no chance to detect the executables in setup.py.

Forget what I wrote :)

Actually, for setup.py we don't need all the information generated by `list_all_bots`, we only need the module names. For this purpose, we can iterate `intelmq/bots/*/*.py`, relative to the location of setup.py and add all the found modules. We only need to exclude the bot-libraries, e.g. files starting with `_` which requires renaming some `lib.py` to `_lib.py`, but that's trivial.

schacht-certat (approved):
Not happy about the IMO unrelated commits on top, but so be it. Lets get this over with.

PR #1754, opened 2021-02-22
BUG: twitter collector: proper URL parsing

wagner-certat:
- Proper input validation in URLs using urllib. CWE-20, found by GitHub's CodeQL.
- Limit replacement ("pastebin.com", "pastebin.com/raw") to a maximum of one.

@hugeox (original author/maintainer) please review as well

### DIFF
--- a/intelmq/bots/collectors/twitter/collector_twitter.py
+++ b/intelmq/bots/collectors/twitter/collector_twitter.py
@@ -33,6 +33,7 @@ To get api login data see: https://python-twitter.readthedocs.io/en/latest/getti
 """
 
 import time
+from urllib.parse import urlsplit
 
 from intelmq.lib.bot import CollectorBot
 from intelmq.lib.utils import create_request_session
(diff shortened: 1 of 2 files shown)


intelmq/bots/collectors/twitter/collector_twitter.py
```diff
@@ -80,11 +81,12 @@ def init(self):
         self.session = create_request_session(self)
 
     def get_text_from_url(self, url: str) -> str:
-        if "pastebin.com" in url:
+        netloc = urlsplit(url)
```
schacht-certat, line 84: `urlsplit` returns a named tuple, so houln't this be `urlsplit(url).netloc`?

intelmq/bots/collectors/twitter/collector_twitter.py
```diff
@@ -80,11 +81,12 @@ def init(self):
...
+        if netloc == "pastebin.com":
```
schacht-certat, line 85: In theory the netloc could also contain the port, so `netloc.startswith('pastebin.com')` would also work in case the url is `https://pastebin.com:443/...`

schacht-certat (changes requested):

intelmq/bots/collectors/twitter/collector_twitter.py
wagner-certat: But it should not match `pastebin.com.example.com`, so we need to split the port off and compare to exact match (or: endswith). And to not split IPv6-Addresses, we need `rsplit(':', maxsplit=1)`

schacht-certat (approved):

waldbauer-certat (approved):

PR #1761, opened 2021-02-25
Remove the XMPP collector and output bots

schacht-certat:
This removes both the XMPP collector bot and the XMPP output bot,
any documentation regarding those bots and the test for those bots.
It also updates installation instructions to not list python-sleekxmpp
as dependency.
This was evaluated on the mailinglist in
https://lists.cert.at/pipermail/intelmq-users/2020-October/000177.html
and the XMPP bots were depracated in 391d625

Closes: #1614


wagner-certat (changes requested):
Could you please add changelog and news entries? Also explaining that we asked on the ML for any users.

schacht-certat:
> Could you please add changelog and news entries? Also explaining that we asked on the ML for any users.

done

wagner-certat (approved):

PR #1769, opened 2021-03-02
Some Shadowserver API Collector and Parser fixes

wagner-certat:
MAINT: shadowserver api collector: remove unneeded imports
the requests library is not explicitly needed in the module
and the MissingDependencyError is not needed and wasn't used

BUG: shadowserver api collector: change file name to json
The API returns file names with the extension `.csv`, although the files are JSON, not CSV. Therefor, for clarity and better error detection in the parser, the file name in `extra.file_name` uses `.json` as extension.

DOC: fix grouping in BOTS with shadowserver api collector
the collectors in the BOTS file are grouped: first the generic, not
feed(-provider)-specific bots, the then the feed-specific ones.

BUG: shadowserver parser: handle 0 for asn fields


schacht-certat (approved):
Not sure if a regex is the fastest way to do string replacement, but on some superficial research the internet has no clear answer on that, so I'm okey with how it is ;)

wagner-certat:
> Not sure if a regex is the fastest way to do string replacement

I wanted to make sure to only replace the string at the end, and that's not possible otherwise. As that operation is only done once per collected file and the expression is pre-compiled also, I think the impact is negligible.

PR #1770, opened 2021-03-02
MAINT: ripe expert: fix mode parameter usage

wagner-certat:
very minor fix

after the automatic conversion of self.parameters -> self, the member
__mode is obsolete and the same as mode.


schacht-certat (approved):
LGTM

PR #1771, opened 2021-03-02
MAINT: Remove Malware Domain List feed

### LINKED ISSUE #1762: Malware Domain List feed broken
As reported on the mailing list, the feed Malware Domain List is broken, it returns an empty page.
https://lists.cert.at/pipermail/intelmq-users/2021-February/000202.html ff.

It is not clear if this is a temporary issue or permanent.

If it's temporary, the parser should be able to handle that, otherwise we can remove it.

schacht-certat:
This commit removes the malwaredomainlist parser bot and all relevant
unit tests. It removes the relevant feed entry in the feeds.yaml file
and drops the malwaredomainlist bot entry in the BOTS file and in the
runtime.conf.
Last but not least it adds two methods to the upgrades.py removing bots
that use the malwaredomainlist parser module and pointing out collectors
that have the malwaredomain list url set.

Closes #1762


intelmq/lib/upgrades.py
```diff
@@ -532,6 +534,41 @@ def v230_csv_parser_parameter_fix_1(defaults, runtime, harmonization, dry_run):
...
+    """
+    Deprecate malwaredomainlist parser
+    """
+    changed = None
+    malwaredomainlistparser = []
+    for bot_id, bot in runtime.items():
+        if bot["module"] == "intelmq.bots.parsers.malwaredomainlist.parser":
+            malwaredomainlistparser.append(bot_id)
+    for bot_id in malwaredomainlistparser:
+        del runtime[bot_id]
```
wagner-certat, line 547: That could cause troubles as the pipeline and (to a lesser degree) the positions need to be adapted as well.

intelmq/lib/upgrades.py
```diff
@@ -550,7 +587,7 @@ def v230_csv_parser_parameter_fix_1(defaults, runtime, harmonization, dry_run):
     ((2, 2, 2), (v222_feed_changes, )),
     ((2, 2, 3), ()),
     ((2, 3, 0), (v230_csv_parser_parameter_fix_1, )),
-    ((3, 0, 0), ()),
+    ((3, 0, 0), (v300_feed_changes, v300_deprecations,)),
```
wagner-certat, line 590: Can you please move the functions to 2.3.0? They are a bug fix IMHO.

wagner-certat (changes requested):
Please also remove the bots from `intelmq/etc/pipeline.conf`.

Could you please add a test case for the upgrade function?

intelmq/lib/upgrades.py
schacht-certat: Fixed, it doesn't change anything, but emits a warning

intelmq/lib/upgrades.py
schacht-certat: done

schacht-certat:
> Please also remove the bots from `intelmq/etc/pipeline.conf`.
> 
> Could you please add a test case for the upgrade function?

done and done

wagner-certat (approved):

PR #1787, opened 2021-03-08
extend intelmqsetup to api

### LINKED ISSUE #1783: Extend intelmqsetup to API
- [x] Check group write permissions for `CONFIG_DIR`
- [x] Create manager dir in `CONFIG_DIR` with group writable.
- [x] Check if group is `intelmq` each.
- [x] Find out the webserver user name
- [x] allow giving webserver username by parameter
- [x] check if group intelmq exists
- [x] check if webserver is in group intelmq

wagner-certat:
Small intelmqsetup refactoring

intelmqsetup: extend to intelmq-api. fixes #1783

### DIFF
--- a/intelmq/bin/intelmqsetup.py
+++ b/intelmq/bin/intelmqsetup.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 """
-© 2019 Sebastian Wagner <wagner@cert.at>
+© 2019-2021 nic.at GmbH <intelmq-team@cert.at>
 
-SPDX-License-Identifier: AGPL-3.0
+SPDX-License-Identifier: AGPL-3.0-only
 
 Sets up an intelmq environment after installation or upgrade by
  * creating needed directories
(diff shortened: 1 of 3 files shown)


waldbauer-certat (approved):
LGTM

wagner-certat:
update to address the test fails

wagner-certat:
The two test fails are because of `test_invalid_ptr2` (temporary glitch?)

intelmq/bin/intelmqsetup.py
```diff
@@ -17,59 +22,135 @@
...
+        getpwnam('intelmq')
+    except KeyError:
+        sys.exit(red("User 'intelmq' does not exist. Please create it and then re-run this program."))
+    try:
+        getgrnam('intelmq')
+    except KeyError:
+        sys.exit(red("Group 'intelmq' does not exist. Please create it and then re-run this program."))
+
+
+def create_directory(directory: str, octal_mode: int, readable_mode: str):
```
schacht-certat, line 66: You can use [stat.filemode(mode)](https://docs.python.org/3/library/stat.html#stat.filemode) to calculate the readable_mode. That would lead to one less arguments in the method

intelmq/bin/intelmqsetup.py
```diff
@@ -17,59 +22,135 @@
...
+    if not os.path.isdir(directory):
+        os.makedirs(directory, mode=octal_mode,
+                    exist_ok=True)
```
schacht-certat, line 69: why the newline?

intelmq/bin/intelmqsetup.py
```diff
@@ -17,59 +22,135 @@
...
+        print(f'Created directory {directory!r} with permissions {readable_mode}.')
+    else:
+        actual_mode = stat.filemode(os.stat(directory).st_mode)
```
schacht-certat, line 72: why compare the string representation if you can compare the actual mode?  (`actual_mode` should probably be `current_mode`)

intelmq/bin/intelmqsetup.py
```diff
@@ -17,59 +22,135 @@
...
+
+
+def intelmqsetup_core(ownership=True, state_file=STATE_FILE_PATH):
+    directories_modes = ((FILE_OUTPUT_PATH, 0o755, 'drwxr-xr-x'),
+                         (VAR_RUN_PATH, 0o755, 'drwxr-xr-x'),
+                         (DEFAULT_LOGGING_PATH, 0o755, 'drwxr-xr-x'),
+                         (CONFIG_DIR, 0o775, 'drwxrwxr-x'),
+                         )
+    for directory, octal_mode, readable_mode in directories_modes:
+        create_directory(directory, octal_mode, readable_mode)
```
schacht-certat, line 106: There are 4 directories- so if you simply write `create_directory` for every directory you have 4 lines of code and it is readable. Why put the directories in a list of tuples and then iterate through them? IMO that makes it only more complex.

schacht-certat (changes requested):
I guess 11b1bab25e0c8d92752a956dcb745dbffa71b284 should not actually be part of that PR?

We can use [`pathlib`](https://docs.python.org/3/library/pathlib.html) now that we are using python >= 3.6. Its way more elegant than using os for most operations.

intelmq/bin/intelmqsetup.py
wagner-certat: Good idea, but the mode needs to be more explicit to get the expected result:
```python
>>> stat.filemode(0o755)
'?rwxr-xr-x'
>>> stat.filemode(0o40755)
'drwxr-xr-x'
```

intelmq/bin/intelmqsetup.py
```diff
@@ -17,59 +22,134 @@
...
+
+
+def intelmqsetup_core(ownership=True, state_file=STATE_FILE_PATH):
+    create_directory(FILE_OUTPUT_PATH, 0o40755)
+    create_directory(VAR_RUN_PATH, 0o40755)
+    create_directory(DEFAULT_LOGGING_PATH, 0o40755)
+    create_directory(CONFIG_DIR, 0o40775)
 
     example_confs = glob.glob(pkg_resources.resource_filename('intelmq', 'etc/*.conf'))
     for example_conf in example_confs:
```
schacht-certat, line 108: You could also use the `glob` method of the pathlib module here:
```
pathlib.Path(pkg_resources.resource_filename('intelmq', 'etc')).glob('*.conf')
```
which gives you `PosixPath` objects to work with- then you don't have to use `Path()` around the name.

intelmq/bin/intelmqsetup.py
```diff
@@ -17,59 +22,134 @@
...
-        fname = os.path.split(example_conf)[-1]
-        if os.path.exists(os.path.join(CONFIG_DIR, fname)):
-            print('Not overwriting existing %r with example.' % fname)
+        fname = Path(example_conf).parts[-1]
```
schacht-certat, line 109: see [`PurePath.name`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.name)

intelmq/bin/intelmqsetup.py
```diff
@@ -17,59 +22,134 @@
...
+        getpwnam('intelmq')
+    except KeyError:
+        sys.exit(red("User 'intelmq' does not exist. Please create it and then re-run this program."))
+    try:
+        getgrnam('intelmq')
+    except KeyError:
+        sys.exit(red("Group 'intelmq' does not exist. Please create it and then re-run this program."))
+
+
+def create_directory(directory: str, octal_mode: int):
```
schacht-certat, line 67: we could also make this `directory: pathlib.Path` then we don't have to use `Path(directory)` four times :smile:

schacht-certat (approved):

wagner-certat:
Found some issues:
The API uses `/etc/intelmq` irrespective of the `CONFIG_DIR` variable, which is `/opt/intelmq/etc` on pip-installations. So the steps need to be performed for that as well. And installing the apache configurations is missing too.

wagner-certat:
> Found some issues:
> The API uses `/etc/intelmq` irrespective of the `CONFIG_DIR` variable, which is `/opt/intelmq/etc` on pip-installations. So the steps need to be performed for that as well. And installing the apache configurations is missing too.

I'll do that afterwards in a separate PR.

PR #1789, opened 2021-03-09
ENH: add http status code expert

schacht-certat:
This adds the module
intelmq.bots.experts.http_status.expert.HttpStatusExpertBot
It also adds a description in the documentation and a changelog entry
introducing the new bot.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ Update allowed classification fields to 2020-01-28 version (#1409, #1476). Old n
 
 #### Experts
 - `intelmq.bots.experts.domain_suffix.expert`: Added `--update-database` option to update domain suffix database.
+- Added `intelmq.bots.experts.http.expert_status`: A bot that fetches the HTTP Status for a given URI and adds it to the message.
 
 #### Outputs
 - Remove `intelmq.bots.outputs.xmpp`: one of the dependencies of the bot was deprecated and according to a short survey on the IntelMQ
--- a/docs/user/bots.rst
+++ b/docs/user/bots.rst
@@ -2041,6 +2041,21 @@ Other errors result in an exception if not ignored by the parameter `gaierrors_t
 All gaierrors can be found here: http://www.castaglia.org/proftpd/doc/devel-guide/src/lib/glibc-gai_strerror.c.html
 
 
+HTTP Status
+^^^^^^^^^^^
+
+Fetches the HTTP Status for a given URI
+
+**Information**
+
+* `name:` intelmq.bots.experts.http.expert_status
+* `description:` The bot fetches the HTTP status for a given URL and saves it in the event.
+
+**Configuration Parameters**
+
+* `field:` The name of the field containing the URL to be checked (required).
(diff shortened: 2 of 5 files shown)


intelmq/bots/experts/http_status/expert.py
```diff
@@ -0,0 +1,59 @@
...
+
+    Parameters
+    ----------
+    field: str
+        The name of the field containing the URL to be checked (required).
+    success_status_codes: List
+        A list of success status codes. If this parameter is omitted or the list is empty,
+        successful status codes are the ones between 200 and 400.
+    """
+    field: str = None  # The field containing the URI
```
waldbauer-certat, line 25: Isn't it better if this field would be `url` instead of `field` as its a httpStatusExpert?

waldbauer-certat (changes requested):

intelmq/bots/experts/http_status/expert.py
waldbauer-certat: Oh, now I got it. Was a bit confused. Thanks for clarification.

waldbauer-certat (approved):

intelmq/bots/experts/http_status/expert.py
wagner-certat, line 25:
```suggestion
    field: str = "source.url"  # The field containing the URI
```
`source.url` would be a good example and meaningful default value

intelmq/bots/experts/http_status/expert.py
```diff
@@ -0,0 +1,59 @@
...
+
+            event.add('status', 'offline')
+
+            try:
+                response = session.get(event[self.field])
+                # If success_status_codes are configured, we use those
+                # to check the success of the request, otherwise we
+                # rely on the boolean value of the response
+                if (self.success_status_codes and response.status_code in self.success_status_codes) or (response):
+                    event.add('status', "online", overwrite=True)
```
wagner-certat, line 47:
```suggestion
                    event.add('status', "online", overwrite=self.overwrite)
```
And add an overwrite parameter

wagner-certat (changes requested):
An `__init__.py` is missing, otherwise `list_all_bots` won't detect it.

Suggestion: As we will have multiple http-related experts, we can group them under `intelmq/bots/experts/http`, so this one could be names `intelmq/bots/experts/http/expert_status.py`.

intelmq/bots/experts/http_status/expert.py
schacht-certat: good idea, fixed!

intelmq/bots/experts/http_status/expert.py
schacht-certat: done.

schacht-certat:
> An `__init__.py` is missing, otherwise `list_all_bots` won't detect it.

also fixed.

> Suggestion: As we will have multiple http-related experts, we can group them under `intelmq/bots/experts/http`, so this one could be names `intelmq/bots/experts/http/expert_status.py`.

still not sure about the best approach, but lets do it like you suggested for now ;)

wagner-certat (approved):

PR #1795, opened 2021-03-10
Cymru CAP Parser: Adapt to new format of 'bruteforce' category

### LINKED ISSUE #1794: Incorrect parsing of Team Cymru message
Some lines we receive from Team Cymru are:
```
bruteforce|x.x.x.x|41390|2021-03-09 00:11:21|destination_port_numbers: 22;port: 16794;protocol: 6;|RN-DATA-, LV
```
The result is `"classification.identifier": "destination_port_numbers: 22"` which is not an identifier. Report type is new as it contains `#generated on 20210310`

wagner-certat:
Previously, the bruteforce events only had the protocol in the comment, while most other categories had a (more or less) clear type-value syntax. Now, the bruteforce categories also uses the type-value syntax. That makes the parser a bit more bloated, as it needs to support both formats, the old and the new. Whereas I don't even know if this is a real switch, or just an additional format.

Interestingly the data no longer contains the application protocol (SSH in the case of the example) :/

fixes #1794

@aleksejsv

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ CHANGELOG
 #### Collectors
 
 #### Parsers
+- `intelmq.bots.parsers.cymru.parser_cap_program`:
+  - Adapt parser to new upstream format for events of category "bruteforce" (PR#1795 by Sebastian Wagner, fixes 1794).
 
 #### Experts
 
--- a/intelmq/bots/parsers/cymru/parser_cap_program.py
+++ b/intelmq/bots/parsers/cymru/parser_cap_program.py
@@ -6,8 +6,7 @@ from intelmq.lib.bot import ParserBot
 
 MAPPING_STATIC = {'bot': {
     'classification.type': 'infected-system'},
-    'bruteforce': {
-    'classification.type': 'brute-force'},
+    'bruteforce': {'classification.type': 'brute-force'},
     'controller': {
     'classification.type': 'c2server'},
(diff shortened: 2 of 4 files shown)


schacht-certat (approved):

waldbauer-certat (approved):

wagner-certat:
> Works better now,

Thanks for testing!
> `"classification.identifier": "generic"`. But what is the idea behind `generic`? What about `"classification.identifier": "brute-force"` like you do with `open-*` identifiers?

IntelMQ's internal data format ("harmonization") has three classification fields. They are (in hierarchy, from top to down): taxonomy, type and identifier. Using this triple it is - in most cases - recognizable what the whole event is about. E.g. malicious code + infected system + zeus means: The event is about a machine infected with zeus.

Previously, Cymru included the application protocol in its reports, many other feeds do so as well (dataplane, shadowserver, spamhaus cert). If the application protocol is available a triple could look like: intrusion-attempts + brute-force + ssh. But now, we don't have a protocol available. Using 'brute-force' does not add value, and 'generic' doesn't either. So we can actually just leave it out then.

aleksejsv:
One more thing, the line:
```
controller|x.x.x.x|12525|2021-03-11 00:02:14|family: http_post;hostname: domain.lv;;port: 80;|INTERNETLTD, LV
```
produces `"malware.name": "http_post"` - is this expected?

PR #1801, opened 2021-03-12
DOC: bots: link ELK-Stack in bots/ES output

wagner-certat:


wagner-certat:
@gethvi

wagner-certat:
No feedback, so probably ok

PR #1819, opened 2021-03-17
ENH: Using msgpack instead of json

waldbauer-certat:
**NOTE**: This is a proof of concept. Being heavily tested!

### Introducation
Msgpack ( MessagePack ) is a (de)serialization format, which is similar to json, but more optimized for m2m ( Machine-to-Machine ) communication. For sure there are better protocols like protobuf, flatbuffers, capnproto, SBE and so on, but this doenst fit in intelmq very well. Msgpack uses a key-value pattern ( like in json ), so there wont be any major change. The real "magic" happens how the data is being stored, JSON is very human-readable due to its serialization, but msgpack packs data into binary format, which results in smaller size & faster processing - see the benchmark below.
If you want to know some specs, check it out [here](https://github.com/msgpack/msgpack/blob/master/spec.md).

Msgpack itself is available for multiple languages like golang, python, javascript, php and so on.

In addition, Redis - our internal message queue - is also capable of using msgpack within its lua api.

### Whats the goal?
- [x] Faster process time for (de)serialization.
- [x] less memory footprint
- [x] no breaking change

### Benchmark
For the benchmark, data was extracted from spamhaus-drop-collector, parsed by spamhaus-drop-parser and measured in deduplicator-expert. 460 events were processed in total.

I've tested the bots above, they worked fine with that change, it might break other bots ( which I havent tested yet )

| Type | Median data size |
|---|---|
| JSON | 387 bytes
| MSGPACK | 329 bytes
| Diff | 58 bytes ( 16,20% )
#### Serialize
| Type | Execution time in ns |
|---|---|
| JSON | 39286
| MSGPACK | 23483
| Diff | 15803 ( 50,35% )
#### Deserialize
| Type | Execution time in ns |
|---|---|
| JSON | 23483
| MSGPACK | 12602
| Diff | 10881 ( 80,62% )

To sum up, changing from json to msgpack will result in a faster (de)serialization and a lower memory footprint.


setup.py
```diff
@@ -15,6 +15,7 @@
     'pytz>=2012c',
     'redis>=2.10',
     'requests>=2.2.0',
+    'msgpack==0.6.2',
```
wagner-certat, line 18: Please don't pin version numbers and use `>=` if possible.

(btw, on intelmq-dev we have 0.5.6 installed)

intelmq/tests/lib/test_message.py
```diff
@@ -154,12 +155,12 @@ def test_event_ne_different_config(self):
     def test_invalid_type(self):
         """ Test if Message raises InvalidArgument for invalid type. """
         with self.assertRaises(exceptions.InvalidArgument):
-            message.MessageFactory.unserialize('{"__type": "Message"}', harmonization=HARM)
+            message.MessageFactory.unserialize(msgpack.dumps({"__type": "Message"}), harmonization=HARM)
```
wagner-certat, line 158: these changes effectively remove the tests for the JSON packer

intelmq/tests/lib/test_message.py
```diff
@@ -514,22 +513,32 @@ def test_event_json_hierarchical(self):
                     '{"observation": "2015-01-01T13:37:00+00:00"}}')
         self.assertDictEqual(json.loads(expected), json.loads(actual))
 
+    def test_event_msgpack(self):
+        """ Test event to_msgpack """
+        event = self.new_event()
+        event = self.add_event_examples(event)
+        actual = event.to_msgpack()
+        print(actual)
```
wagner-certat, line 521: print

intelmq/tests/lib/test_pipeline.py
```diff
@@ -66,7 +67,7 @@ def setUp(self):
 
     def test_receive(self):
         self.pipe.state['test-bot-input'] = [SAMPLES['normal'][0]]
-        self.assertEqual(SAMPLES['normal'][1], self.pipe.receive())
+        self.assertEqual(SAMPLES['normal'][1], utils.decode(self.pipe.receive()))
```
wagner-certat, line 70: instead of decoding, why not just compare with `SAMPLES['normal'][0]` instead?

intelmq/tests/lib/test_pipeline.py
```diff
@@ -75,7 +76,7 @@ def test_send(self):
 
     def test_receive_unicode(self):
         self.pipe.state['test-bot-input'] = [SAMPLES['unicode'][0]]
-        self.assertEqual(SAMPLES['unicode'][1], self.pipe.receive())
+        self.assertEqual(SAMPLES['unicode'][1], utils.decode(self.pipe.receive()))
```
wagner-certat, line 79: I'm not sure if that testcase still makes sense. Is unicode data in the pipeline valid at all now?

intelmq/tests/lib/test_pipeline.py
```diff
@@ -106,7 +107,7 @@ def test_reject(self):
         self.pipe.state['test-bot-input'] = [SAMPLES['normal'][0]]
         self.pipe.receive()
         self.pipe.reject_message()
-        self.assertEqual(SAMPLES['normal'][1], self.pipe.receive())
+        self.assertEqual(SAMPLES['normal'][1], utils.decode(self.pipe.receive()))
```
wagner-certat, line 110: same as above, why not compare with `SAMPLES['normal'][0]`?

intelmq/tests/bots/experts/idea/test_expert.py
```diff
@@ -82,10 +83,10 @@ def test_conversion(self):
         # The ID in the generated Idea event is random, so we have to extract
         # the data from the "output" field and compare after removing ID's
         event = self.get_output_queue()[0]
-        self.assertIsInstance(event, str)
-        event_dict = json.loads(event)
+        self.assertIsInstance(event, bytes)
+        event_dict = msgpack.loads(event)
```
wagner-certat, line 87: Could we use `MessageFactory.unserialize` here? Then we are flexible with the default packer and we don't need `b"output"` below.

PR #1821, opened 2021-03-19
ENH: Improvements to the Shodan parser

monoidic:
This is a set of improvements to improve the amount of data parsed by the Shodan parser, based on a set of collected Shodan banners.


wagner-certat (changes requested):
Thank you for this very extensive addition!

I only have one plea: Would it be possible to add simple tests cases for the added code (*not* all the mappings, that's not worth the effort), just that we have example input/output which covers the changed and added code (only the functions `_keys_conversion` and `_maybe_single_to_list`)

wagner-certat:
the codespell output 
```
./intelmq/bots/parsers/shodan/parser.py:288: dont' ==> don't
```
is a false-positive, we'll need to add it to the whitelist

wagner-certat (approved):

wagner-certat:
Thank you again for this pull request!

I cherry-picked the commits in branch maintenance, so the additions will be part of IntelMQ 2.3.1.

PR #1833, opened 2021-03-29
ENH: Convert extra.end_time to timestamp in Shadowserver feed

monoidic:
This should ensure that the `extra.end_time` field from the Shadowserver parser is treated as a timestamp rather than as a string.


monoidic:
The remaining test failures appear to be there in `HEAD` as well and weren't introduced by this PR.

PR #1835, opened 2021-03-29
WIP: ENH: tweaks to Shodan parser bot

monoidic:
This PR introduces a new expert bot, with the purpose of filtering out notable Shodan events from Shodan Stream. As it is right now, it can include many unnotable events (public web servers, video game servers etc.), and this is an attempt at making it more usable.
The filter includes a set of methods that take the event an argument and return a boolean indicating whether or not to consider the banner noteworthy. All configured filters are checked and if at least one filter returns true, the event is passed on and `extra.shodan_filter.matched_filters` is populated with all matched filter method names (hence there's extra reason for the method names to be descriptive). Any events that do not match any filters will be discarded.

Current filters:
* The event notes that the host has known vulnerabilities (configurable between only vulnerabilities verified by Shodan or all vulnerabilities)
* There root object has one of a set of user-specified keys (able to detect most kinds of services)
* The event has some associated tags (defaults: compromised, ICS and database)
* The event is for a recursive DNS server operating over UDP, representing a potential DDoS amplification threat (toggleable)
* The event is for a host with an expired TLS certificate (toggleable)

It also includes some updates to the Shodan parser.

Any feedback is welcome.


monoidic:
Very well, I will look into using the sieve expert bot to handle the filtering instead.
I would, however, like to keep the first commit from this PR, which updates the Shodan parser, in particular because I accidentally used `opts.vulns` when I meant to use `vulns`. I will modify this PR tomorrow.

wagner-certat:
> Very well, I will look into using the sieve expert bot to handle the filtering instead.

I haven't verified that it works and I'm pretty sure that the sieve expert would need extensions as well to handle lists (btw, I handle how lists are currently used in the sieve expert).

> I would, however, like to keep the first commit from this PR

Sure.

wagner-certat:
Also cherry-picked for maintenance and thus for the next bugfix release.

Added a small change in e64d15dbf9da8a25a029681539b101764827a176

PR #1843, opened 2021-04-07
PKG: remove BOTS file from Debian package

schacht-certat:
This removes any traces of the BOTS file from the Debian package rules
file, the intelmq.install file and the d/conffiles listing. It also adds
a intelmq.maintscript to remove the BOTS file on upgrade to 3.0.0.


wagner-certat (approved):

wagner-certat:
merged (squashed) in 4383b0c1cf594089879c8af0607c1b66f09034c2

Issue #1862, opened 2021-04-14
Cymru CAP parser misses destination ip and port because of missing whitespace


kapalex:
Some of the option field use the `key: value` formatting which is supported, but it seems that in some cases (like dest_ip) it uses `key:value`

This seems to be an easy fix, will add a pull request soon

PR #1864, opened 2021-04-14
Fix issue with some fields not being parsed #1862

kapalex:
# Please read the notes below and replace them with the description of you pull request

Make sure you follow the instructions in the [Developer Guide](https://intelmq.readthedocs.io/en/latest/dev/guide.html) - it describes how to run the test suite and which coding rules to follow.

# Commits

Please review your commits before creating a pull request. We try to keep our commit history clean.
If you had to fix something in your code and added another commit on top of your existing work, please [squash](https://blog.simontimms.com/2016/02/18/i-squash-my-pull-requests-and-you-should-too/) your commits.
This makes it a lot easier to find out why things work the way they do later on.

But please also keep in mind that one commit should only contain changes that belong together.
> The cardinal rule for creating good commits is to ensure there is only one "logical change" per commit
> - [OpenStack Wiki on GitCommitMessages](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

Thus, if your changes affect multiple files and also contains tests and documentation (which it should), it might make sense to have the tests and the documentation in a separate commit.

# Commit Messages

Please explain your changes also in the git commit message.
The commit message should contain a subject that gives an overview of the changes in the commit and is limited to 72 characters.
It should start with a capital letter and it should not end with a period.
Below the subject should be, separated by an empty line, the body of the commit message. The body should explain what the commit changes and why it changes thing the way it does.
Explain your modification and also explain why you didn't chose a different approach.
See also [How To Write a Git Commit Message](https://chris.beams.io/posts/git-commit/).

# Description

Parsing issue with the whitespace in ': ' described in #1862

### DIFF
--- a/intelmq/bots/parsers/cymru/parser_cap_program.py
+++ b/intelmq/bots/parsers/cymru/parser_cap_program.py
@@ -296,7 +296,7 @@ class CymruCAPProgramParserBot(ParserBot):
                         else:
                             break
                     raise ValueError('Unable to parse comment %r of category %r. Please report this.' % (comment, category))
-            key, value = comment.split(': ')
+            key, value = comment.split(':', 1)
             key = key.strip()
             value = value.strip()
             if key == 'family':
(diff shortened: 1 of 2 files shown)


kapalex:
This seems to be ready to go @wagner-certat 
A very minor fix

Please note line 279 which has a ': ' as well is ignored on purpose to avoid 'http://...' in phishing category being split.

wagner-certat (changes requested):
Thanks for the bug report and the fix!

However, could you please extend the tests to this specific case? Or modify an existing line in the test file. Thereby we can prevent that future changes to the code, including refactoring, reverts the change unintentionally.

kapalex:
@wagner-certat tried to add the test case with a minor addition but sadly it seems to fail the tests.
Honestly entirely unsure as to why the tests are failing (I cannot see the full comparison, and do not have a proper testing environment set up).
If you manage to figure out why it is failing (or maybe can provide a full output of the failing assertion event) I would gladly fix the commits.

If not I might have time to set it up correctly next week.

wagner-certat:
Looks like the fix doesn't work as expected. The comment of the affected line is not parsed correctly anymore.

PR #1865, opened 2021-04-15
ENH: impmlement handling of environment variables

schacht-certat:
See also IEP01
(https://lists.cert.at/pipermail/intelmq-dev/2020-December/000344.html)
which states that "Every bot also consults the environment and the
values that are set there overwrite the values in any configuration
file"

### DIFF
--- a/intelmq/lib/bot.py
+++ b/intelmq/lib/bot.py
@@ -747,18 +747,6 @@ class Bot(object):
 
         self.log_processed_messages_seconds = timedelta(seconds=self.log_processed_messages_seconds)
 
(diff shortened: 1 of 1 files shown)


waldbauer-certat (approved):

wagner-certat:
Will the documentation changes be as part of #1867 ?

Should we add a check if the parameter exists and is a variable? E.g.:
```python
if not callable(getattr(self, option, None)):
    setattr(self, option, value)
```
because otherwise this would be possible:
```bash
INTELMQ_SHUTDOWN="...malicious code.."
```
I don't know if it's possible to inject code this way though

PR #1873, opened 2021-04-19
FIX: recreate the legacy pipeline_configuration

schacht-certat:
This commit introduces a method _pipeline_configuration() that provides
the same dict the legacy `pipeline_configuration` attribute used to
store.

### DIFF
--- a/intelmq/bin/intelmqctl.py
+++ b/intelmq/bin/intelmqctl.py
@@ -1255,6 +1255,17 @@ Get some debugging output on the settings and the environment (to be extended):
             val = utils.list_all_bots()
             return 0, val
 
+    def _pipeline_configuration(self):
+        pipeline_configuration = {}
+        for botid, botconfig in self.runtime_configuration.items():
+            if botid != 'global':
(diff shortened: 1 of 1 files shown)


waldbauer-certat (approved):

intelmq/bin/intelmqctl.py
```diff
@@ -1255,6 +1255,16 @@ def list_bots(self, non_zero=False, configured=False):
             val = utils.list_all_bots()
             return 0, val
 
+    def _pipeline_configuration(self):
+        pipeline_configuration = {}
+        for botid, botconfig in self.runtime_configuration.items():
+            pipeline_configuration[botid] = {"source_queue": f"{botid}-queue", "destination_queues": []}
+            if 'source_queue' in botconfig['parameters']:
```
wagner-certat, line 1262: "global" needs to be ignored here, otherwise `intelmqctl list queues` etc. raises a KeyError

wagner-certat (changes requested):

intelmq/bin/intelmqctl.py
schacht-certat: fixed

wagner-certat (approved):

PR #1877, opened 2021-04-20
FIX: use a private attribute for log_processed_messages_seconds

schacht-certat:


waldbauer-certat (approved):

wagner-certat (changes requested):
This ignores the parameter `log_processed_messages_seconds` and therefore removes the feature. If this is inevitable (also temporarily), we need to *at least* document it, eventually also raise a warning when the parameter is actually set, stating that the parameter is ignored (to circumvent other errors).

wagner-certat (approved):

Issue #1886, opened 2021-04-22
gethostbyname can't handle null for parameter gaierrors_to_ignore


wagner-certat:
If parameter `gaierrors_to_ignore` is null:
```
openphish-gethostbyname-expert: Bot initialization failed.
Traceback (most recent call last):
  File "/home/aaron/intelmq/intelmq/lib/bot.py", line 168, in __init__
    self.init()
  File "/home/aaron/intelmq/intelmq/bots/experts/gethostbyname/expert.py", line 36, in init
    ignore = ignore.split(',')
AttributeError: 'NoneType' object has no attribute 'split'
```
Should be handled gracefully

PR #1890, opened 2021-04-22
BUG: gethostbyname: fix handling null gaierrors_to_ignore

wagner-certat:
the value null for the parameter gaierrors_to_ignore raised an error

this commit fixes this by handling it gracefully. actually was already
there, just in the wrong order

fixes certtools/intelmq#1886


waldbauer-certat (approved):
LGTM

Issue #1681, opened 2020-12-14
Sieve expert nested if-statements


wagner-certat:
Currently it is not possible to nest multiple if-statements or use parentheses:
```
if source.asn == 64496 {
    if classification.taxonomy == 'vulnerable' or classification.type == 'infected-system' {
        ...
    }
}

if source.asn == 64496 && (classification.taxonomy == 'vulnerable' || classification.type == 'infected-system') {
    ....
}
```
Splitting up the conditions into multiple separate if-clauses is possible in some cases, but not in all.

It's not clear which of the two options is easier to implement.

PR #1895, opened 2021-04-22
ENH: Refactor some of the Sieve expert, extend the Sieve DSL

monoidic:
This PR refactors some of the Sieve expert and extends the Sieve DSL. In particular, it adds:
* for the below items, a generic `TypedValue` type was added, which can be any of a float, int, bool or str, as well as `ListValue`, which can contain any of the above
* some methods of operating on list objects, like `:equals` for exact matching of lists, and `:setequals`, `:overlaps`, `:subsetof` and `:supersetof` for set-based operations, along with negated counterparts if `!` is specified before the operator (i.e `! :overlaps` would match if the two sets are disjoint)
* matching on bool values with == or !=
* appending to a list value, or force-appending, which appends to lists normally, coerces non-lists into lists with one element and initializes nonexisting lists
* unit tests for all of the above

Feedback is welcome.

### DIFF
--- a/intelmq/bots/experts/sieve/expert.py
+++ b/intelmq/bots/experts/sieve/expert.py
@@ -10,6 +10,11 @@ import os
 import re
 import traceback
 import datetime
+import operator
+
+from typing import Optional
+from enum import Enum, auto
+from functools import partial
 
 import intelmq.lib.exceptions as exceptions
 from intelmq import HARMONIZATION_CONF_FILE
@@ -27,10 +32,10 @@ except ImportError:
     metamodel_from_file = None
 
 
-class Procedure:
-    CONTINUE = 1  # continue processing subsequent rules (default)
-    KEEP = 2      # stop processing and keep event
-    DROP = 3      # stop processing and drop event
+class Procedure(Enum):
+    CONTINUE = auto()  # continue processing subsequent rules (default)
+    KEEP = auto()      # stop processing and keep event
+    DROP = auto()      # stop processing and drop event
 
 
 class SieveExpertBot(Bot):
@@ -40,7 +45,38 @@ class SieveExpertBot(Bot):
     _harmonization = None
     file: str = "/opt/intelmq/var/lib/bots/sieve/filter.sieve"  # TODO: should be pathlib.Path
 
-    def init(self):
+    _string_op_map = {
+        '==': operator.eq,
+        '!=': operator.ne,
+        ':contains': lambda lhs, rhs: lhs.find(rhs) >= 0,
+        ':notcontains': lambda lhs, rhs: lhs.find(rhs) == -1,
+        '=~': lambda lhs, rhs: re.search(rhs, lhs) is not None,
+        '!~': lambda lhs, rhs: re.search(rhs, lhs) is None,
+    }
+
+    _numeric_op_map = {
+        '==': operator.eq,
+        '!=': operator.ne,
+        '<=': operator.le,
+        '>=': operator.ge,
+        '<': operator.lt,
+        '>': operator.gt,
+    }
+
+    _basic_math_op_map = {
+        '+=': operator.add,
+        '-=': operator.sub,
+    }
+
+    _cond_map = {
+        'ExistMatch': lambda self, match, event: self.process_exist_match(match.key, match.op, event),
+        'StringMatch': lambda self, match, event: self.process_string_match(match.key, match.op, match.value, event),
+        'NumericMatch': lambda self, match, event: self.process_numeric_match(match.key, match.op, match.value, event),
+        'IpRangeMatch': lambda self, match, event: self.process_ip_range_match(match.key, match.range, event),
+        'Expression': lambda self, match, event: self.match_expression(match, event),
+    }
+
+    def init(self) -> None:
         if not SieveExpertBot._harmonization:
             harmonization_config = utils.load_configuration(HARMONIZATION_CONF_FILE)
             SieveExpertBot._harmonization = harmonization_config['event']
@@ -106,7 +142,7 @@ class SieveExpertBot(Bot):
         except Exception:
             return [['error', 'Validation of Sieve file failed with the following traceback: %r' % traceback.format_exc()]]
 
-    def process(self):
+    def process(self) -> None:
         event = self.receive_message()
         procedure = Procedure.CONTINUE
         if self.sieve:  # empty rules file results in empty string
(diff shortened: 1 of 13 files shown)


monoidic:
The remaining CI/CD failures appear in `HEAD` as well.

intelmq/bots/experts/sieve/expert.py
```diff
@@ -256,64 +279,91 @@ def process_ip_range_match(self, key, ip_range, event):
...
                 if action.operator != '=':
-                    value = SieveExpertBot.compute_basic_math(action, event)
+                    value = self.compute_basic_math(action, event)
                 event.change(action.key, value)
-        elif action.__class__.__name__ == 'RemoveAction':
+        elif name == 'RemoveAction':
             if action.key in event:
                 del event[action.key]
+        elif name == 'AppendAction':
+            if action.key in event and isinstance(event[action.key], list):
```
wagner-certat, line 341: if `event[action.key]` does not exist, nothing happens, is this intentionally?

intelmq/bots/experts/sieve/expert.py
```diff
@@ -256,64 +279,91 @@ def process_ip_range_match(self, key, ip_range, event):
...
+                list_ = event[action.key]
+                list_.append(action.value)
+                event.add(action.key, list_, overwrite=True)
```
wagner-certat, line 344: Is the `overwrite=True` intentioned, even though this is not the force-flavor of the Append-Action?

wagner-certat (changes requested):
Thank you so much for this tremendous and immaculate contribution!
The current handling of lists (`==` for "contains"/subsetof) has annoyed myself as well for some time and you can image that I am relieved you accepted the challenge to work on the sieve expert, which is, so I believe, the most complex but also most powerful, bot that we have in IntelMQ. The quality of your work is - again - superb.

I hope that we can get rid of the unintuitive behaviour of `==` with lists on the right side soon (after 3.0).

Could you please describe the newly added operators in the documentation in `docs/user/bots.rst`? Then they'll be available directly in the [user-documentation for the sieve bot](https://intelmq.readthedocs.io/en/latest/user/bots.html#sieve)

The only build fail of this PR is unrelated to your work and @schacht-certat was able to fix it in the meanwhile on the develop branch.

intelmq/bots/experts/sieve/expert.py
monoidic: It is. Nonexistent keys and non-list values are intentionally silently ignored. I'm somewhat confident in the latter option, but ambivalent on whether nonexistent keys should be created or not. If you think it would be a better design to create new fields on append, then I will change this.

intelmq/bots/experts/sieve/expert.py
monoidic: Now that I look at it, the `event.add` line in the non-force variant is currently redundant as `event[action.key]` and `list_` are the same list at the same memory locations, and updating `_list` implicitly updates the list at `event[action.key]`.
The same line is not redundant in `AppendForceAction` as a new list may be created. Though for this same reason, if `AppendAction` were to be made to also optionally create a new list if the key is not already included in the event, this line would be there to handle the case of a new list being created.
An alternative would be to simply only have the `event.add` in the branch of code where a new list is created.

monoidic:
Rebased on the newest develop branch `HEAD`, plus two changes in `expert.py`:
* the `AppendAction` case in `process_action` now creates a list in case the key does not yet exist
* updated tests to reflect this
* `_cond_map` extracted from `process_condition`
* added documentation on the changes in the DSL

monoidic:
Perhaps the current overloaded use of `==` for checking both for equality and inclusion in a list of values could be removed? For instance, I believe that for conveying the meaning of "if classification.type is either c2-server or malware-configuration, then..." is explained better by
```
if classification.type :in ['c2-server','malware-configuration'] { ... }
```
than by
```
if classification.type == ['c2-server','malware-configuration'] { ... }
```

monoidic:
Perhaps the current overloaded use of `==` for checking both for equality and inclusion in a list of values could be removed? For instance, I believe that for conveying the meaning of "if classification.type is either c2-server or malware-configuration, then..." is explained better by
```
if classification.type :in ['c2-server','malware-configuration'] { ... }
```
than by
```
if classification.type == ['c2-server','malware-configuration'] { ... }
```

This would, of course, be a breaking change for anyone relying on the list behaviour, and if the `:equals` were to be replaced with `==`, this would end up silently breaking a lot of people's sieve bots.

Some consideration could also be had for introducing `!` in match conditions to invert arbitrary conditions. This would allow `if ! :contains`, for instance, to replace `if :notcontains` and allow conditions to be more expressive, like having `if (x && y) || (y && z)`. True-matching operators like `:contains` wouldn't need inverted variants like `:notcontains` either.
The current behaviour of matching on nonexistant fields returning false regardless of negation in `ListMatch` would change, but that hasn't been added yet either.

Maybe something for allowing for nested if's would be neat as well?

I'm not sure if this is the best place to discuss potential major (breaking) changes and the future of the Sieve bot though.

wagner-certat:
> Perhaps the current overloaded use of `==` for checking both for equality and inclusion in a list of values could be removed? For instance, I believe that for conveying the meaning of "if classification.type is either c2-server or malware-configuration, then..." is explained better by
> 
> ```
> if classification.type :in ['c2-server','malware-configuration'] { ... }
> ```
> 
> than by
> 
> ```
> if classification.type == ['c2-server','malware-configuration'] { ... }
> ```

I fully agree!

> This would, of course, be a breaking change for anyone relying on the list behaviour, and if the `:equals` were to be replaced with `==`, this would end up silently breaking a lot of people's sieve bots.

Breaking bad behavior is less of an issue from my POV, but if the meaning changes without warning, that's bad. As alternative, we can disallow `foo == ['foo', 'bar']` (in favor of `foo :equals ['foo', 'bar']`?) in IntelMQ 3.0, and re-allow it in a later version with a different meaning. That way, we can ensure all users fix their scripts - as they throw errors. What do you think about this approach?
 
> Some consideration could also be had for introducing `!` in match conditions to invert arbitrary conditions, rather than just list matches. This would allow `if ! :contains`, for instance, to replace `if :notcontains` and allow conditions to be more expressive, like having `if ! (w && x) || (y && z)`. True-matching operators like `:contains` wouldn't need inverted variants like `:notcontains`.

That would be another great addition :)

> The current behaviour of matching on nonexistant fields returning false regardless of negation in `ListMatch` would change, but nobody relies on those methods yet anyway.

I'm not adept in that issue, but sounds reasonable.

> Maybe something for allowing for nested if's would be neat as well?

Definitely, this was something we were looking for already in the past (#1681) but the - until then undocumented - possibility for using parenthesis was enough for us.

> I'm not sure if this is the best place to discuss potential major (breaking) changes and the future of the Sieve bot though. And I would be interested in implementing these ideas at some point, if they don't seem too far-fetched.

Thank you very much for all your contributions. I really appreciate what you are doing!

Let's finalize the list-topic here first, and tackle the other issues and ideas (negation, nesting ifs, nonexistant fields) afterwards

monoidic:
> Let's finalize the list-topic here first, and tackle the other issues and ideas (negation, nesting ifs, nonexistant fields) afterwards

I believe I have implemented the requested changes.

wagner-certat (approved):

wagner-certat:
Could you please verify if I got the list of changes right?
https://github.com/certtools/intelmq/commit/872282f6f002409782f59a2503866479f39f7a4d

monoidic:
At the moment, only the list/set operators have `!` negation, and I should actually change the behaviour of i.e `! :equals` to match on nonexistant fields, just as `process_string_match` does. Though I will have a more generic solution for negation in the next PR and remove the special-case.
LGTM otherwise.

wagner-certat:
Thanks! I slightly adapted the changelog entry with your feedback incorporated (a7ecf626a)

PR #1901, opened 2021-04-26
Add templated SMTP output bot

creideiki:
The current SMTP output bot only supports formatting events as CSV attachments, which is not flexible enough for our needs, which include sending JSON-formatted attachments. Hence, this bot, which dumps the responsibility for formatting onto the runtime configuration, allowing you to create things like:
```yaml
attachments:
  - content-type: application/json
    text: |
      {{
        "host": "{event[source.fqdn]}",
        "malware": "{event[malware.name]}"
      }}
    name: report.json
```
Available formatting is limited to string substitution. This was sufficient for our needs, and avoids introducing control structures in the templates. If you feel there is a need for even more complicated output formatting, I could look into integrating a full template engine, e.g. Jinja.

The code and tests are heavily based on the original SMTP output bot. If it had included a copyright line, I would have reproduced it in my comments, but it didn't.

The bot was rewritten to the new configuration system for this pull request. The one we are actually running is still based on IntelMQ v2.3.1, with the old configuration system. This means that while the tests pass, I have not tried to run this exact code live.

### DIFF
--- a/intelmq/bots/outputs/templated_smtp/output.py
+++ b/intelmq/bots/outputs/templated_smtp/output.py
@@ -1,0 +1,201 @@
+# -*- coding: utf-8 -*-
+"""Templated SMTP output bot
+
+SPDX-FileCopyrightText: 2021 Linköping University <https://liu.se/>
+SPDX-License-Identifier: AGPL-3.0-or-later
+
+Sends a MIME Multipart message built from an event and static text
+using Jinja2 templates.
+
+Templates are in Jinja2 format with the event provided in the variable
+"event". E.g.:
+
+   mail_to: "{{ event['source.abuse_contact'] }}"
+
+See the Jinja2 documentation at https://jinja.palletsprojects.com/ .
+
+Attachments are template strings, especially useful for sending
+structured data. E.g. to send a JSON document including "malware.name"
+and all other fields starting with "source.":
+
+   attachments:
+     - content-type: application/json
+       text: |
+         {
+           "malware": "{{ event['malware.name'] }}",
+           {%- set comma = joiner(", ") %}
+           {%- for key in event %}
+              {%- if key.startswith('source.') %}
+           {{ comma() }}"{{ key }}": "{{ event[key] }}"
+              {%- endif %}
+           {%- endfor %}
+         }
+       name: report.json
+
+You are responsible for making sure that the text produced by the
+template is valid according to the content-type.
+
+SMTP authentication is attempted if both "smtp_username" and
+"smtp_password" are provided.
+
+Parameters:
+
+attachments: list of objects with structure:
+              - content-type: string, templated, content-type to use.
+                text: string, templated, attachment text.
+                name: string, templated, filename of attachment.
+
+body: string, optional, default see below, templated, body text.
+      The default body template prints every field in the event except
+      'raw', in undefined order, one field per line, as "field:
+      value".
+
+mail_from: string, templated, sender address.
+
+mail_to: string, templated, recipient addresses, comma-separated.
+
+smtp_host: string, optional, default "localhost", hostname of SMTP
+           server.
+
+smtp_password: string, default null, password (if any) for
+               authenticated SMTP.
+
+smtp_port: integer, default 25, TCP port to connect to.
+
+smtp_username: string, default null, username (if any) for
+               authenticated SMTP.
+
+tls: boolean, default false, whether to use use SMTPS. If true, also
+     set smtp_port to the SMTPS port.
+
+starttls: boolean, default true, whether to use opportunistic STARTTLS
+          over SMTP.
+
+subject: string, optional, default "IntelMQ event", templated, e-mail
+         subject line.
+
+verify_cert: boolean, default true, whether to verify the server
+             certificate in STARTTLS or SMTPS.
+
+"""
+
+import io
+import smtplib
+import ssl
+from typing import List, Optional
+from email.message import EmailMessage
+try:
+    from jinja2 import Template
+except:
+    Template = None
+
+from intelmq.lib.bot import Bot
+from intelmq.lib.exceptions import ConfigurationError
+from intelmq.lib.exceptions import MissingDependencyError
+
+
+class TemplatedSMTPOutputBot(Bot):
+    smtp_host: str = "localhost"
+    smtp_port: int = 25
+    ssl: bool = False
+    starttls: bool = False
+    username: Optional[str] = None
+    password: Optional[str] = None
+    verify_cert: bool = True
+    attachments: List[str] = []
+    mail_from: Optional[str] = None
+    mail_to: Optional[str] = None
+    subject: str = "IntelMQ event"
+    body: str = """{%- for field in event %}
+    {%- if field != 'raw' %}
+{{ field }}: {{ event[field] }}
+    {%- endif %}
+{%- endfor %}
+"""
+
+    def init(self):
+        if not Template:
+            raise MissingDependencyError("jinja2")
+
+        if self.ssl:
+            self.smtp_class = smtplib.SMTP_SSL
+        else:
+            self.smtp_class = smtplib.SMTP
+        if not self.mail_from:
+            raise ConfigurationError("Mail", "No sender specified")
+        if not self.mail_to:
+            raise ConfigurationError("Mail", "No recipient specified")
+
+        if self.username and not self.password:
+            raise ConfigurationError("Server", "SMTP username provided, but not password")
+        if self.password and not self.username:
+            raise ConfigurationError("Server", "SMTP password provided, but not username")
+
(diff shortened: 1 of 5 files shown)


creideiki:
I took a look at Jinja2, and it turns out using it is really easy, so I pushed an updated version for that instead.

At least the test pass, but this is even less tested than the `str.format()` version, since I haven't made the switch to Jinja2 in our running code. And I don't think I will bother, since writing Jinja2 templates in the JSON configuration where strings cannot contain newlines seems horrible.

wagner-certat:
I very much like the new feature to set texts and attachment / MIME parts more flexible, but on the other hand, doubling the code also means doubling the maintenance effort required. Does it sound sensible from your POV to replace the simple SMTP output by your one?

creideiki:
> Does it sound sensible from your POV to replace the simple SMTP output by your one?

Sure. I didn't replace the simple SMTP bot in the pull request because I wasn't sure what the policy for breaking changes was, but if you're OK with it and it happens in a major version bump, I see no problems. I'll write some documentation on how to convert from the old bot for those who still want CSV format.

wagner-certat:
> Sure. I didn't replace the simple SMTP bot in the pull request because I wasn't sure what the policy for breaking changes was, but if you're OK with it and it happens in a major version bump, I see no problems. I'll write some documentation on how to convert from the old bot for those who still want CSV format.

Most (basic) parameters don't need I change, e.g. all the SMTP-ones. For the others, I can then create a upgrade-function so that `intelmqctl upgrade-config` automatically upgrades the parameters. Then we are on the safe side, and still we have all the new features.

intelmq/bots/outputs/templated_smtp/output.py
```diff
@@ -0,0 +1,187 @@
...
+        if self.ssl:
+            self.smtp_class = smtplib.SMTP_SSL
+        else:
+            self.smtp_class = smtplib.SMTP
+        if not self.mail_from:
+            raise ConfigurationError("Mail", "No sender specified")
+        if not self.mail_to:
+            raise ConfigurationError("Mail", "No recipient specified")
+        if not self.subject:
+            raise ConfigurationError("Mail", "No subject specified")
```
schacht-certat, line 118: It might make sense to not enforce the subject and the body and simply use sensible default values

intelmq/bots/outputs/templated_smtp/output.py
```diff
@@ -0,0 +1,187 @@
...
+                                   disposition='attachment',
+                                   filename=att["name"].render(event=event))
+
+            recipients = [recipient.strip()
+                          for recipient
+                          in msg['To'].split(',')]
+            self.logger.info("Sending mail to %s.", recipients)
+            smtp.send_message(msg, from_addr=msg['From'],
+                              to_addrs=recipients)
+
```
schacht-certat, line 183: Can we add a try catch block around the `send_message` call and add some debug logging?

intelmq/bots/outputs/templated_smtp/output.py
```diff
@@ -0,0 +1,187 @@
...
+        with self.smtp_class(self.smtp_host, self.smtp_port,
+                             **kwargs) as smtp:
+            if self.starttls:
+                if self.verify_cert:
+                    smtp.starttls(context=ssl.create_default_context())
+                else:
+                    smtp.starttls()
+            if self.username and self.password:
+                smtp.login(user=self.username, password=self.password)
+
```
schacht-certat, line 163: The `smtp.login` call could use some exception handling and debug logging.
And maybe we could check in the `init` if both username and password are set

schacht-certat (changes requested):
Great bot, thank, I also think it could replace the existing SMTP output bot. (On the other hand, if there are more bots using SMTP connections, it might make sense to think about an SMTP Mixin at some point...)

Could you squash your 12 commits into one?

wagner-certat:
> On the other hand, if there are more bots using SMTP connections, it might make sense to think about an SMTP Mixin at some point...

:+1: 

I now think, that completely replacing the existing SMTP output also raises the bar to use it, as the templated bot is more complex to configure and requires jinja2.

intelmq/bots/outputs/templated_smtp/output.py
creideiki: I'm not sure what that would accomplish. Any errors from smtplib will already cause an exception to propagate upwards where the framework will log it, dump the event to disk so it isn't lost and restart the bot. I don't think the bot can do anything more intelligent by itself.

intelmq/bots/outputs/templated_smtp/output.py
creideiki: Same as above regarding exception handling.

intelmq/bots/outputs/templated_smtp/output.py
creideiki: I'm not sure they're sensible, but there are now default values.

intelmq/bots/outputs/templated_smtp/output.py
```diff
@@ -0,0 +1,202 @@
...
+except:
+    Template = None
+
+from intelmq.lib.bot import Bot
+from intelmq.lib.exceptions import ConfigurationError
+from intelmq.lib.exceptions import MissingDependencyError
+
+
+class TemplatedSMTPOutputBot(Bot):
+    smtp_host: Optional[str] = None
```
wagner-certat, line 97:
```suggestion
    smtp_host: str = 'localhost'
```

intelmq/bots/outputs/templated_smtp/output.py
```diff
@@ -0,0 +1,202 @@
...
+    {%- endif %}
+{%- endfor %}
+"""
+
+    def init(self):
+        if not Template:
+            raise MissingDependencyError("jinja2")
+
+        if not self.smtp_host:
+            raise ConfigurationError("Server", "No SMTP host specified")
```
wagner-certat, line 120:
```suggestion
```

wagner-certat (changes requested):
I think we can get rid of the check for the host as well, as localhost is a sensible default value. For from and to I have no good idea.

creideiki:
If that takes care of all comments, I'll squash and sign the finished commits as requested. Do you prefer squashing to one commit for everything, a few individual commits for code/documentation/tests, one commit per file, or something else?

Also, while I've been working on this, a new major version of Jinja has been released (according to PyPi, confusingly called Jinja2 version 3). But that's not even packaged for Fedora yet, so I haven't looked at how big the differences are, and this code is still using Jinja2 version 2.x.

wagner-certat:
> If that takes care of all comments, I'll squash and sign the finished commits as requested.

Thanks!
> Do you prefer squashing to one commit for everything, a few individual commits for code/documentation/tests, one commit per file, or something else?

Whatever is easiest (or: easy enough) for you. I can also squashing all in one commit myself (even github has that function now)

> Also, while I've been working on this, a new major version of Jinja has been released (according to PyPi, confusingly called Jinja2 version 3). But that's not even packaged for Fedora yet, so I haven't looked at how big the differences are, and this code is still using Jinja2 version 2.x.

I think we shouldn't wait for that :) Let's just merge it now.

wagner-certat (approved):

wagner-certat:
> FAIL: test_invalid_ptr2 (intelmq.tests.bots.experts.reverse_dns.test_expert.TestReverseDnsExpertBot)

I'll investigate that fail, won't block this PR :)

creideiki:
> > Do you prefer squashing to one commit for everything, a few individual commits for code/documentation/tests, one commit per file, or something else?
> 
> Whatever is easiest (or: easy enough) for you. I can also squashing all in one commit myself

There you go, one each for code, tests, documentation, and authors list.

I got a bit worried when GitHub said the tests failed after rebasing and squashing, but luckily that's in the reverse DNS expert, so not my fault :)

> (even github has that function now)

True, but that leaves the new commits unsigned.

> > Also, while I've been working on this, a new major version of Jinja has been released (according to PyPi, confusingly called Jinja2 version 3). But that's not even packaged for Fedora yet, so I haven't looked at how big the differences are, and this code is still using Jinja2 version 2.x.
> 
> I think we shouldn't wait for that :) Let's just merge it now.

Agreed.

That only leaves the question of maybe deprecating the original SMTP output bot, but that seems out of scope for this pull request (and also doesn't need my input).

wagner-certat:
> I got a bit worried when GitHub said the tests failed after rebasing and squashing, but luckily that's in the reverse DNS expert, so not my fault :)

It was even just a temporary issue. Re-running the jobs solved the "problem"

> That only leaves the question of maybe deprecating the original SMTP output bot, but that seems out of scope for this pull request (and also doesn't need my input).

Definitely a priority right now for me. Also I'm not sure if deprecating the other one is a good idea, as it solves the simplest use-case with easier configuration and without additional dependency.

PR #1908, opened 2021-04-27
FIX: create directory for state file

schacht-certat:


schacht-certat:
This will be fixed by #1912

PR #1909, opened 2021-04-27
ENH: add util functions to provide legacy interface

schacht-certat:
This replaces the _pipeline_configuration method from the IntelMQCTL class
with a function get_legacy_pipeline in the utils module. It also adds a
get_legacy_runtime function to provide an interface to the runtime
without the 'global' key and without the pipeline configuration objects.


wagner-certat:
We can use these functions in `intelmq.bin.intelmqctl.IntelMQController.upgrade_config` to provide the parameters, but merging runtime and pipeline afterwards can be tricky.

schacht-certat:
thought about that, we won't need that after all

PR #1910, opened 2021-04-27
Add Microsoft Defender bots

creideiki:
Fetches security alerts from Microsoft Defender ATP.

This bot is quite a mess in its present state - it horribly abuses the `extra.` namespace in messages, it hard-codes processing decisions, the error handling is not great, and it cannot do OAuth2 token refresh, so it needs to re-authenticate for each call. It's also only written for IntelMQ v2.3.1 currently.

Still, it does what it's supposed to do, and we think that's useful. We're currently using it together with a custom output bot to create tickets in our helpdesk system whenever Defender finds malware on a client.

Do you think it would be worthwhile to try to beat this code into shape for upstream inclusion? Or is this a non-starter because of e.g. our need to send lots of information that doesn't fit in the default harmonisation, and thus stuffing the `extra.` namespace full of JSON?

### DIFF
--- a/docs/user/bots.rst
+++ b/docs/user/bots.rst
@@ -2118,6 +2118,126 @@ Requires `jinja2`::
    pip3 install -r intelmq/bots/experts/defender_advanced_hunting/REQUIREMENTS.txt
 
 
+.. _intelmq.bots.experts.defender_file.expert:
+
+Defender file
+^^^^^^^^^^^^^
+
+**Information**
+
+* `name:` intelmq.bots.experts.defender_file.expert
+* `lookup:` Microsoft Defender ATP
+* `public:` yes
+* `description:` Fetches file information from Microsoft Defender ATP.
+
+
+**Configuration Parameters**
+* `api_region`: Optional, string. Default: None. Cloud region for API
+  calls. Either None (for worldwide) or one of "us", "eu", or "uk".
+* `tenant_id`: String, your Office 365 tenant ID.
+* `client_id`: String, the client ID you created for this application.
+* `client_secret`: String, the secret you created for this application.
+* `retries`: int, default 5, number of times to retry after receiving
+  a "file not found" error.
+* `min_wait`: int, default 30, minimum number of seconds to wait
+  between retry attempts.
+* `max_wait`: int, default 60, maximum number of seconds to wait
+  between retry attempts.
+
+**Description**
+
+Defender wants to include quite a lot of information that doesn't fit
+in IntelMQ's default harmonisation, so it abuses the "extra" namespace
+to store its information.
+
+There is a race condition in the Defender cloud service, where a file
+information structure may not be ready for retrieval even though an
+alert has been fired recently. To guard against this, any file not
+found errors result in retries a maximum of `retries` times, with a
+random delay of between `min_wait` and `max_wait` seconds between each
+attempt.
+
+Input structure::
+
+   "extra.evidence": [
+      List of "evidence" structures. The format is fixed, but contains
+      the union of all fields ever used. Hence, most fields are null,
+      and which fields contain useful data depends on the type of
+      evidence, which is stored in the "entityType" field.
+
+      Structure:
+      {
+         "aadUserId": "",
+         "accountName": "",
+         "detectionStatus": "",
+         "domainName": "",
+         "entityType": "",
+         "evidenceCreationTime": "Timestamp",
+         "fileName": "",
+         "filePath": "",
+         "ipAddress": "",
+         "parentProcessCreationTime": "",
+         "parentProcessFileName": "",
+         "parentProcessFilePath": "",
+         "parentProcessId": "",
+         "processCommandLine": "",
+         "processCreationTime": "",
+         "processId": "",
+         "registryHive": "",
+         "registryKey": "",
+         "registryValue": "",
+         "registryValueType": "",
+         "sha1": "",
+         "sha256": "",
+         "url": "",
+         "userPrincipalName": "",
+         "userSid": ""
+      }
+   ]
+
+Output structure::
+
+   "extra.fileinfo": [
+
+      List of "fileinfo" structures, one for each evidence object of
+      type "File". These are linked to the evidence structures through
+      the sha1 and sha256 hash values.
+
+      Structure:
+      {
+         "@odata.context": "https://api-eu.securitycenter.windows.com/api/$metadata#Files/$entity",
+         "determinationType": "",
+         "determinationValue": "Malware name, if known",
+         "fileProductName": "",
+         "filePublisher": "",
+         "fileType": "",
+         "globalFirstObserved": "Timestamp",
+         "globalLastObserved": "Timestamp",
+         "globalPrevalence": Integer,
+         "isPeFile": Boolean,
+         "isValidCertificate": "",
+         "issuer": "",
+         "md5": "Hash",
+         "sha1": "Hash",
+         "sha256": "Hash",
+         "signer": "",
+         "signerHash": "",
+         "size": Integer
+      }
+   ]
+
+**Requirements**
+
+Requires credentials as described in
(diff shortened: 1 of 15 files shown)


wagner-certat:
Cool addition!

It may not be the primary use-case of IntelMQ, but if it's useful and it does not *conflict* with IntelMQs principles, I see no obstacle to include it.

On the oauth issue: Well, if it works, it's already better than having nothing^^

creideiki:
Great, thanks! I'll make a note of updating this pull request once we've migrated this bot to 3.0 and cleaned it up a bit.

wagner-certat:
> Great, thanks! I'll make a note of updating this pull request once we've migrated this bot to 3.0 and cleaned it up a bit.

Cool, thanks in advance!

creideiki:
Pushed a new and improved design, now sort of conforming a bit more to the rest of IntelMQ.

Split the horrible hardcoded monolithic mess of a collector that basically only worked for our use case into four bots:
1) collector
2) parser
3) file expert (asks the Defender cloud for information on a file, searched by SHA1 hash)
4) advanced hunting expert (sends arbitary queries to the Defender cloud and updates the event with the results, somewhat based on my earlier Splunk saved search expert).

With simple unit tests for all. No documentation written yet, though.

creideiki:
I have force-pushed a new version of this set of bots, which is what we have been running internally for a while. They should be ready for review now.

This pull request contains six different bots, all having to do with communicating with Microsoft Defender's API:
1. Collector, for retrieving malware alerts from the Defender cloud service.
2. Parser.
3. Text formatting expert. In case the alert wasn't an exact match for something we know how to automate, this bot formats the alert to text in order to send it via e-mail (using the templated SMTP output bot) for manual processing.
4. Advanced hunting expert, which can run arbitrary queries against the Defender API. We use it to enrich the initial alert by querying for logged in users at the time of the alert.
5. File information expert. This takes the SHA1 or SHA256 hash of the file included in the initial alert, and fetches all information about it, e.g. signature (if it's a signed executable) and earliest detection time.

After that, our botnet has a bot (which is not included here, because it is very specific to us) for creating or updating an issue in our ticketing system. The final bot is:

6. Comment output. This updates the initial Defender alert with a comment; we add our internal ticket number.

ghost:
Thanks for the thorough pull request! I will have a look at it next week.

docs/user/bots.rst
```diff
@@ -168,6 +168,47 @@ The API is available at `/intelmq/push` if the HTTP interface is used (default).
 The `tornado` library is required.
 
 
+.. _intelmq.bots.collectors.defender.collector_defender:
+
+Defender
```
sebix, line 173:
```suggestion
Microsoft Defender
```

docs/user/bots.rst
```diff
@@ -1234,6 +1275,104 @@ http://www.team-cymru.com/bogon-reference.html
 * `description:` Parses data from full bogons feed.
 
 
+.. _intelmq.bots.parsers.defender.parser:
+
+Defender
```
sebix, line 1280:
```suggestion
Microsoft Defender
```

docs/user/bots.rst
```diff
@@ -1899,6 +2038,228 @@ Public documentation: https://www.team-cymru.com/IP-ASN-mapping.html#dns
 * ``: Overwrite existing fields. Default: `True` if not given (for backwards compatibility, will change in version 3.0.0)
 
 
+.. _intelmq.bots.experts.defender_advanced_hunting.expert:
+
+Defender advanced hunting
```
sebix, line 2043:
```suggestion
Microsoft Defender advanced hunting
```

docs/user/bots.rst
```diff
@@ -1899,6 +2038,228 @@ Public documentation: https://www.team-cymru.com/IP-ASN-mapping.html#dns
...
+advanced queries.
+
+Requires `jinja2`::
+
+   pip3 install -r intelmq/bots/experts/defender_advanced_hunting/REQUIREMENTS.txt
+
+
+.. _intelmq.bots.experts.defender_file.expert:
+
+Defender file
```
sebix, line 2123:
```suggestion
Microsoft Defender file
```

docs/user/bots.rst
```diff
@@ -1899,6 +2038,228 @@ Public documentation: https://www.team-cymru.com/IP-ASN-mapping.html#dns
...
+advanced queries.
+
+Requires `tenacity`, tested with version 7::
+
+   pip3 install -r intelmq/bots/experts/defender_file/REQUIREMENTS.txt
+
+
+.. _intelmq.bots.experts.defender_to_text.expert:
+
+Defender to text
```
sebix, line 2243:
```suggestion
Microsoft Defender to text
```

docs/user/bots.rst
```diff
@@ -3551,6 +3912,49 @@ xxx.xxx.xxx.xxx    Intel::ADDR    phishing    100    MISP XXX
 www.testdomain.com    Intel::DOMAIN    apt    85    CERT
 ```
 
+
+.. _intelmq.bots.outputs.defender_comment.output:
+
+Defender comment
```
sebix, line 3918:
```suggestion
Microsoft Defender comment
```

sebix (commented):
For consistency I propose to move the bots to "microsoft" directories, grouping them by vendor/feed provider.

Please also add a changelog.

sebix:
Someone needs to review the code, see https://lists.cert.at/pipermail/intelmq-dev/2021-November/000558.html

creideiki:
Thanks for taking a look at this. Unfortunately, there are some weird problems with the API right now which in the worst case may force a redesign.

When we started using the Defender API, it behaved as documented, and we ran the collector every 600 seconds asking for all new alerts from the last 605 seconds. This worked with the code in this pull request.

A few days ago, the collector stopped returning results. After experimentation, we discovered that the API endpoint that searches for the newest alerts only returned alerts that were more than 6 hours old. We added code to the comment output (not yet submitted here) to add a static string as a comment, and changed our botnet to always run the comment output bot as soon as an alert came in. We also added code to the parser (also not submitted yet) to ignore alerts with this comment set. This meant that we could increase the lookback time in the collector to more than 6 hours, thus getting alerts again.

Yesterday, the API's behaviour changed again. The search endpoint now only has a delay of 30 minutes, but the delay now applies not only to new alerts, but also to new comments. That means that even when we add comments to alerts saying the botnet has already seen them, when we search for alerts we don't see that comment until 30 minutes later. This meant that alerts we had already handled looked new and went through the botnet again. To compensate, we increased the collector rate limit to 3600 seconds (1 hour).

While wrestling with these problems, we found out that while the API we are using is still documented and not deprecated, there is a new API that is now recommended instead.

We are currently waiting to hear back from Microsoft support regarding the strange delays on the old API, and whether that API will be deprecated and we will have to move to the new API.

sebix:
Phew, that sounds annoying.

sebix:
Dear @creideiki,

Did the product's API stabilize and have you been able to get it working?

creideiki:
We have not heard anything more from Microsoft. The old API used here is still documented, the new API doesn't supply all the information the old one does (and that we need), and we are still running these bots in production, with the horrible hacks described above to work around the fact that the old API only returns events after several hours' delay.

sebix:
OK, that the code is already as stable as it can currently be, right? And additionally, it is thoroughly tested in production by yours?

creideiki:
That does seem to be the case, at least when the botnet is configured to work around the deficiencies. That configuration would need to be documented. And we have a few changes in production to support that workaround that are not included in this pull request, because I had hoped they would not be required long-term.

But if you think this functionality would be useful, I could incorporate those code changes and write that documentation.

creideiki:
There is now a third API, which does seem to contain all the information we want to use, but is in beta and explicitly not for production use: https://learn.microsoft.com/en-us/graph/api/security-list-alerts_v2

I'm starting to think that it's impossible to write a generic module to talk to Microsoft's APIs. They change so often that every API consumer has to be a developer to keep up.

sebix:
Oh dear that's complex.

PR #1913, opened 2021-04-28
ENH: add some small helper functions

schacht-certat:
those are used in the API

### DIFF
--- a/intelmq/lib/utils.py
+++ b/intelmq/lib/utils.py
@@ -876,6 +876,15 @@ def list_all_bots() -> dict:
     return bots
 
 
+def get_runtime() -> dict:
+    return load_configuration(RUNTIME_CONF_FILE)
+
+
 def get_global_settings() -> dict:
-    runtime_conf = load_configuration(RUNTIME_CONF_FILE)
+    runtime_conf = get_runtime()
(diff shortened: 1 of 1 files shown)


intelmq/lib/utils.py
```diff
@@ -876,6 +876,15 @@ def list_all_bots() -> dict:
     return bots
 
 
+def get_runtime() -> dict:
+    return load_configuration(RUNTIME_CONF_FILE)
```
wagner-certat, line 880: shouldn't global be removed here?

schacht-certat: I would not say so. The `get_runtime` and `set_runtime` functions read and write the `runtime.yaml` as a whole

wagner-certat: yeah, we can make a `get_effective_runtime(bot_id)` or so in the future, if we need it.

schacht-certat: If you're saying that the term `runtime` only refers to the individual configuration of the bots but not to the global settings which are now also part of `runtime.conf`, then it might make sense to rename the configuration file altogether (`intelmq.yaml`) and add one key `runtime` and one key `global`. That way we could use functions with meaningful names.

PR #1918, opened 2021-04-29
ENH: New expert bot for uWhoisd

Rafiot:
This bot will query an instance of uWhoisd (https://github.com/Lookyloo/uwhoisd), and add the whois entry into `extra.whois`, I don't think it is the proper way to do so, if it's not, let me know and I'll change it.

### DIFF
--- a/intelmq/bots/experts/uwhoisd/expert.py
+++ b/intelmq/bots/experts/uwhoisd/expert.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+
+import socket
+
+from intelmq.lib.bot import Bot
+
+
+class UniversalWhoisExpertBot(Bot):
+    """ Universal Whois expert bot get the whois entry related an a domain,
+    hostname, IP address, or ASN from a centralised uWhoisd instance """
+    server: str = 'localhost'
+    port: int = 4243
+
+    def _whois(self, query: str) -> str:
+        print(query)
+        bytes_whois = b''
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+            sock.connect((self.server, self.port))
+            sock.sendall('{}\n'.format(query).encode())
+            while True:
+                data = sock.recv(2048)
+                if not data:
+                    break
+                bytes_whois += data
+        to_return = bytes_whois.decode()
+        return to_return
+
+    def process(self):
+        event = self.receive_message()
+        if 'source.url' in event:
+            query = event.get('source.url')
+        elif 'source.asn' in event:
+            query = event.get('source.asn')
+        elif 'source.fqdn' in event:
+            query = event.get('source.fqdn')
+        elif 'source.ip' in event:
+            query = event.get('source.ip ')
+        else:
+            query = None
+
+        if query:
+            whois_entry = self._whois(query)
+            event.add('extra', {'whois': whois_entry})
+
+        self.send_message(event)
+        self.acknowledge_message()
+
+
+BOT = UniversalWhoisExpertBot
--- a/intelmq/tests/bots/experts/uwhoisd/test_expert.py
+++ b/intelmq/tests/bots/experts/uwhoisd/test_expert.py
@@ -1,0 +1,41 @@
+# -*- coding: utf-8 -*-
+import os
+import unittest
+import json
+
+import intelmq.lib.test as test
+from intelmq.bots.experts.uwhoisd.expert import UniversalWhoisExpertBot
+
+
+EXAMPLE_INPUT = {"__type": "Event",
+                 "source.fqdn": "www.cert.at",
+                 "time.observation": "2015-01-01T00:00:00+00:00"
+                 }
+EXAMPLE_OUTPUT = {"__type": "Event",
+                  "source.fqdn": "www.cert.at",
+                  "extra.whois": "Whois record",  # will be replaced in the code below
+                  "time.observation": "2015-01-01T00:00:00+00:00"
+                  }
+
+
+@test.skip_exotic()
+@unittest.skipIf(not os.environ.get('INTELMQ_UNIVERSALWHOIS_TEST'), "UniversalWhois test skipped, as we dont wanna overload public instance")
+class TestUniversalWhoisExpertBot(test.BotTestCase, unittest.TestCase):
+
+    @classmethod
+    def set_bot(cls):
+        cls.bot_reference = UniversalWhoisExpertBot
+
+    def test(self):
+        self.input_message = EXAMPLE_INPUT
+        self.run_bot()
+        event = self.get_output_queue()[0]
+        self.assertIsInstance(event, str)
+        event = json.loads(event)
+        if 'extra.whois' in event and 'NG8867695-NICAT' in event['extra.whois']:
+            EXAMPLE_OUTPUT['extra.whois'] = event['extra.whois']
+        self.assertMessageEqual(0, EXAMPLE_OUTPUT)
+
+
+if __name__ == '__main__':  # pragma: no cover
+    unittest.main()


intelmq/bots/experts/uwhoisd/expert.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class UniversalWhoisExpertBot(Bot):
+    """ Universal Whois expert bot get the whois entry related an a domain,
+    hostname, IP address, or ASN from a centralised uWhoisd instance """
+    server: str = 'localhost'
+    port: int = 4243
+
+    def _whois(self, query: str) -> str:
+        print(query)
```
wagner-certat, line 15: Please remove or replace by `self.logger.debug`

intelmq/bots/experts/uwhoisd/expert.py
```diff
@@ -0,0 +1,49 @@
...
+        bytes_whois = b''
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+            sock.connect((self.server, self.port))
+            sock.sendall('{}\n'.format(query).encode())
```
wagner-certat, line 19: I have exciting news for you: We dropped Python < 3.6 support, so we can use format strings =)

intelmq/tests/bots/experts/uwhoisd/test_expert.py
```diff
@@ -0,0 +1,41 @@
...
+        cls.bot_reference = UniversalWhoisExpertBot
+
+    def test(self):
+        self.input_message = EXAMPLE_INPUT
+        self.run_bot()
+        event = self.get_output_queue()[0]
+        self.assertIsInstance(event, str)
+        event = json.loads(event)
+        if 'extra.whois' in event and 'NG8867695-NICAT' in event['extra.whois']:
+            EXAMPLE_OUTPUT['extra.whois'] = event['extra.whois']
```
wagner-certat, line 36: we can leave EXAMPLE_OUTPUT out completely, if we only test `'NG8867695-NICAT' in event['extra.whois']`

intelmq/bots/experts/uwhoisd/expert.py
```diff
@@ -0,0 +1,49 @@
...
+        elif 'source.fqdn' in event:
+            query = event.get('source.fqdn')
+        elif 'source.ip' in event:
+            query = event.get('source.ip ')
+        else:
+            query = None
+
+        if query:
+            whois_entry = self._whois(query)
+            event.add('extra', {'whois': whois_entry})
```
wagner-certat, line 43:
```suggestion
            event.add('extra.whois': whois_entry)
```

intelmq/tests/bots/experts/uwhoisd/test_expert.py
```diff
@@ -0,0 +1,41 @@
...
+                 }
+EXAMPLE_OUTPUT = {"__type": "Event",
+                  "source.fqdn": "www.cert.at",
+                  "extra.whois": "Whois record",  # will be replaced in the code below
+                  "time.observation": "2015-01-01T00:00:00+00:00"
+                  }
+
+
+@test.skip_exotic()
+@unittest.skipIf(not os.environ.get('INTELMQ_UNIVERSALWHOIS_TEST'), "UniversalWhois test skipped, as we dont wanna overload public instance")
```
wagner-certat, line 22: Maybe I'm missing something, but doesn't this bot connect to localhost? (which won't work)

wagner-certat (changes requested):
Interesting project!

The bot code looks very good, I'm basically just missing documentation in `docs/user/bots.rst` covering a description of what this bot does and the parameters.

intelmq/bots/experts/uwhoisd/expert.py
Rafiot: woops, that was relic of the debug, I will remove it.

intelmq/bots/experts/uwhoisd/expert.py
Rafiot: nice!

intelmq/tests/bots/experts/uwhoisd/test_expert.py
Rafiot: True, it's probably enough.

intelmq/tests/bots/experts/uwhoisd/test_expert.py
Rafiot: We don't have a public instance of uWhoisd (yet?), so I ran it locally for testing and it worked when I did that:

```
INTELMQ_TEST_EXOTIC=1 INTELMQ_UNIVERSALWHOIS_TEST=1 nosetests-3.4 -v intelmq/tests/bots/experts/uwhoisd/test_expert.py
```

wagner-certat: Ah, ok. I'll check what options we have to test that bot.

Rafiot:
I'm wondering btw: right now, if we have an URL, an ASN, an FQDN, and an IP in the event, we will only query for the URL and ignore the other ones.

Would it make sense to add multiple entries in the extra key? Something like `extra.whois.url`, `extra.whois.ip`, ...?

wagner-certat:
> I'm wondering btw: right now, if we have an URL, an ASN, an FQDN, and an IP in the event, we will only query for the URL and ignore the other ones.
> 
> Would it make sense to add multiple entries in the extra key? Something like `extra.whois.url`, `extra.whois.ip`, ...?

I was also thinking if we should make the order configurable.

But actually, how does this bot differ from a generic whois expert? The differences that I recognize are, that this bot here always contacts *one* server and supports URLs (I guess the whoisd just extracts the domain from the URL? I haven't found upstream documentation on how it works).
To make the information actionable, the missing part is extracting information from the whois response, which is extremely tricky and can be done later in the processing pipeline as well (but then we would need two bots for getting the abuse-contact from whois, which sounds contra-intuitive to me).

There are endless opportunities to extend this (or any whois-related) expert. We can always expand the functionality of this bot. How much you want to do now basically only depends on how motivated you are right now :) Strictly speaking, the only blocker for merging this PR are the missing docs and the decision if we call the bot `uwhoisd` or `whois` (with the - currently required - paramer for setting a uwhoisd server).

Rafiot:
> > I'm wondering btw: right now, if we have an URL, an ASN, an FQDN, and an IP in the event, we will only query for the URL and ignore the other ones.
> > Would it make sense to add multiple entries in the extra key? Something like `extra.whois.url`, `extra.whois.ip`, ...?
> 
> I was also thinking if we should make the order configurable.

We can do that, but I'm wondering how often users will assume they get everything, even on a single IntelMQ instance, I can imagine having multiple usecases

> But actually, how does this bot differ from a generic whois expert? The differences that I recognize are, that this bot here always contacts _one_ server and supports URLs (I guess the whoisd just extracts the domain from the URL? I haven't found upstream documentation on how it works).

The data returned by uWhoisd is almost always the same as a normal whois call (it uses the `whois` command by default under the hoods). 
uWhoisd doesn't extract the domain from the url, I fixed it on the module side (https://github.com/certtools/intelmq/pull/1918/commits/08ea3ba813bde8c5f4586cfb42cb5e7af2e11e0d#diff-e0496854f70e04e61fc3a95e5ccb60e2ee1ed1e5a42786c80a82bab720179057L30), the only thing uwhoisd does with the query is stripping the hostname part (blah.example.co.uk -> example.co.uk).

The advantage of using it is that it acts as a proxy, caches the response, and allow to query historical information about whois entries (if the whowas module is enabled). Then, it also allows to have specific options for some whois servers, such as for `.lu`, where there is ratelimiting: https://github.com/Lookyloo/uwhoisd/blob/main/extra/uwhoisd.ini#L34
Or returning a message when a TLD doesn't have a whois server at all (`.es`): https://github.com/Lookyloo/uwhoisd/blob/main/extra/uwhoisd.ini#L64

> To make the information actionable, the missing part is extracting information from the whois response, which is extremely tricky and can be done later in the processing pipeline as well (but then we would need two bots for getting the abuse-contact from whois, which sounds contra-intuitive to me).

As you said extracting information from a whois entry is a complete nightmare, so it is not my priority right now: we use the whois entry as a last resort way to get a abuse contact and we grep for email addresses in the response, which is disgusting, but kinda works, sometimes.

> There are endless opportunities to extend this (or any whois-related) expert. We can always expand the functionality of this bot. How much you want to do now basically only depends on how motivated you are right now :) Strictly speaking, the only blocker for merging this PR are the missing docs and the decision if we call the bot `uwhoisd` or `whois` (with the - currently required - paramer for setting a uwhoisd server).

It would be nice to merge this bot now, and improve it down the line. And for the doc, is this one ok: https://github.com/certtools/intelmq/pull/1918/commits/0f1a9c3921b9ef27a561f49bec1227b13855ae49

I'll try to improve the uwhoisd doc soon.

wagner-certat:
> > > I'm wondering btw: right now, if we have an URL, an ASN, an FQDN, and an IP in the event, we will only query for the URL and ignore the other ones.
> > > Would it make sense to add multiple entries in the extra key? Something like `extra.whois.url`, `extra.whois.ip`, ...?
> > 
> > 
> > I was also thinking if we should make the order configurable.
> 
> We can do that, but I'm wondering how often users will assume they get everything, even on a single IntelMQ instance, I can imagine having multiple usecases

yes, definitely. flexibility and modularities are the buzzwords here :)

> > But actually, how does this bot differ from a generic whois expert? The differences that I recognize are, that this bot here always contacts _one_ server and supports URLs (I guess the whoisd just extracts the domain from the URL? I haven't found upstream documentation on how it works).
> 
> The data returned by uWhoisd is almost always the same as a normal whois call (it uses the `whois` command by default under the hoods).
> uWhoisd doesn't extract the domain from the url, I fixed it on the module side ([08ea3ba#diff-e0496854f70e04e61fc3a95e5ccb60e2ee1ed1e5a42786c80a82bab720179057L30](https://github.com/certtools/intelmq/commit/08ea3ba813bde8c5f4586cfb42cb5e7af2e11e0d#diff-e0496854f70e04e61fc3a95e5ccb60e2ee1ed1e5a42786c80a82bab720179057L30)),

ahh, ok. For that purpose we also have the url2fqdn expert, which also takes care of the special case of IP addresses in URLs.


> the only thing uwhoisd does with the query is stripping the hostname part (blah.example.co.uk -> example.co.uk).

I think that's an important bit of information, could you add that to the bot's docs as well?

> > To make the information actionable, the missing part is extracting information from the whois response, which is extremely tricky and can be done later in the processing pipeline as well (but then we would need two bots for getting the abuse-contact from whois, which sounds contra-intuitive to me).
> 
> As you said extracting information from a whois entry is a complete nightmare, so it is not my priority right now: we use the whois entry as a last resort way to get a abuse contact and we grep for email addresses in the response, which is disgusting, but kinda works, sometimes.

Nightmare is a good description. Even the popular Python whois-libraries can't do that - at least in my tests.

> > There are endless opportunities to extend this (or any whois-related) expert. We can always expand the functionality of this bot. How much you want to do now basically only depends on how motivated you are right now :) Strictly speaking, the only blocker for merging this PR are the missing docs and the decision if we call the bot `uwhoisd` or `whois` (with the - currently required - paramer for setting a uwhoisd server).
> 
> It would be nice to merge this bot now, and improve it down the line.

yep, fully agree

> And for the doc, is this one ok: [0f1a9c3](https://github.com/certtools/intelmq/commit/0f1a9c3921b9ef27a561f49bec1227b13855ae49)

looks good, except for some minor things

> 
> I'll try to improve the uwhoisd doc soon.

Thanks!

docs/user/bots.rst
```diff
@@ -2908,6 +2908,27 @@ writes it to `source.fqdn` or `destination.fqdn` if it is a hostname, or
 
 * `overwrite`: boolean, replace existing FQDN / IP address?
 
+uWhoisd
+^^^^^^^
+
+[uWhoisd](https://github.com/Lookyloo/uwhoisd) is a universal Whois server that supports
+caching and stores whois entries for historical purposes.
+
+The bot sends a request for `source.url`, `source.asn`, `source.fqdn`, or `source.ip`
```
wagner-certat, line 2917:
```suggestion
The bot sends a request for `source.url`, `source.asn`, `source.fqdn` or `source.ip` (the first existing field takes precedence)
```
or something like that

docs/user/bots.rst
```diff
@@ -2908,6 +2908,27 @@ writes it to `source.fqdn` or `destination.fqdn` if it is a hostname, or
...
+
+**Information**
+
+* `name:` intelmq.bots.experts.uwhoisd.expert
+* `description:` uWhoisd is a universal Whois server
+
+**Configuration Parameters**
+
+* `server`: IP or hostname to connect to
+* `port`: Port to connect to
```
wagner-certat, line 2930:
```suggestion
* `server`: IP or hostname to connect to  (default: localhost)
* `port`: Port to connect to (default: 4243)
```

wagner-certat (changes requested):

docs/user/bots.rst
Rafiot: I'll reorder them for now: url -> fqdn -> ip -> asn: it is closer to what I would expect in term of getting the most relevant whois entry for an intelmq event

Rafiot:
> > We can do that, but I'm wondering how often users will assume they get everything, even on a single IntelMQ instance, I can imagine having multiple usecases
> 
> yes, definitely. flexibility and modularities are the buzzwords here :)

Sounds good, should we add all the whois entries now? Or wait a bit?

> > The data returned by uWhoisd is almost always the same as a normal whois call (it uses the `whois` command by default under the hoods).
> > uWhoisd doesn't extract the domain from the url, I fixed it on the module side ([08ea3ba#diff-e0496854f70e04e61fc3a95e5ccb60e2ee1ed1e5a42786c80a82bab720179057L30](https://github.com/certtools/intelmq/commit/08ea3ba813bde8c5f4586cfb42cb5e7af2e11e0d#diff-e0496854f70e04e61fc3a95e5ccb60e2ee1ed1e5a42786c80a82bab720179057L30)),
> 
> ahh, ok. For that purpose we also have the url2fqdn expert, which also takes care of the special case of IP addresses in URLs.

In the whois case, it doesn't matter, whois will take care of figuring out if the request is about an IP or a domain. 

> > the only thing uwhoisd does with the query is stripping the hostname part (blah.example.co.uk -> example.co.uk).
> 
> I think that's an important bit of information, could you add that to the bot's docs as well?

Just did it, let me know what you think.

> > As you said extracting information from a whois entry is a complete nightmare, so it is not my priority right now: we use the whois entry as a last resort way to get a abuse contact and we grep for email addresses in the response, which is disgusting, but kinda works, sometimes.
> 
> Nightmare is a good description. Even the popular Python whois-libraries can't do that - at least in my tests.

Yeah, and most of the whois libraries try to be pure python, which means they do the requests themselves which is fragile, often doesn't query the most appropriate server (debian whois is a pretty big piece of work I which not to reimplement: https://github.com/rfc1036/whois), if you're running lots of queries, you end up being blacklisted by the whois servers, and the parsing is only limited to a small subset of responses anyway so it doesn't scale at all. uWhois tries to mitgate at least some of the issues with a reasonably limited amount of work.

I imagine it would make sense to maintain some sort whois parsing but someone will need to put the work on that, and I'm not sure it is worth the time on the long run.

wagner-certat:
> > > We can do that, but I'm wondering how often users will assume they get everything, even on a single IntelMQ instance, I can imagine having multiple usecases
> > 
> > 
> > yes, definitely. flexibility and modularities are the buzzwords here :)
> 
> Sounds good, should we add all the whois entries now? Or wait a bit?

If you want/need all four whois data entries anyway, let's do it now. Otherwise we'd be changing the data format (`extra.whois` evolving from a field to a dict).

> > > As you said extracting information from a whois entry is a complete nightmare, so it is not my priority right now: we use the whois entry as a last resort way to get a abuse contact and we grep for email addresses in the response, which is disgusting, but kinda works, sometimes.
> > 
> > 
> > Nightmare is a good description. Even the popular Python whois-libraries can't do that - at least in my tests.
> 
> Yeah, and most of the whois libraries try to be pure python, which means they do the requests themselves which is fragile, often doesn't query the most appropriate server (debian whois is a pretty big piece of work I which not to reimplement: https://github.com/rfc1036/whois), if you're running lots of queries, you end up being blacklisted by the whois servers, and the parsing is only limited to a small subset of responses anyway so it doesn't scale at all. uWhois tries to mitgate at least some of the issues with a reasonably limited amount of work.

Thanks for the context. I see that you are already preparing for the uwhoisd docs :)

> I imagine it would make sense to maintain some sort whois parsing but someone will need to put the work on that, and I'm not sure it is worth the time on the long run.

Yep, especially when RDAP is getting more and more widespread (and we already have an expert bot for that).

docs/user/bots.rst
```diff
@@ -2908,6 +2908,37 @@ writes it to `source.fqdn` or `destination.fqdn` if it is a hostname, or
 
 * `overwrite`: boolean, replace existing FQDN / IP address?
 
+uWhoisd
+^^^^^^^
+
+[uWhoisd](https://github.com/Lookyloo/uwhoisd) is a universal Whois server that supports
```
wagner-certat, line 2914:
```suggestion
`uWhoisd <https://github.com/Lookyloo/uwhoisd>`_ is a universal Whois server that supports
```

docs/user/bots.rst
```diff
@@ -2908,6 +2908,37 @@ writes it to `source.fqdn` or `destination.fqdn` if it is a hostname, or
...
+The bot sends a request for `source.url`, `source.fqdn`, `source.ip` or `source.asn` (the first existing field takes precedence)
+to the configured uWhoisd instance and saves the retrieved whois entry `extra.whois`.
+
+Events without `source.url`, `source.fqdn`, `source.ip`, or `source.asn`, are ignored.
+
+**Note**: requesting a whois entry for a fully qualified domain name (FQDN) only works if the request
+only contains the domain. uWhoisd will automatically strip the subdomain part if it is present in the request.
+
+Example: `https://www.theguardian.co.uk`
+* TLD: `co.uk` (uWhoisd uses the [Mozilla public suffix list](https://publicsuffix.org/list/) as a reference)
```
wagner-certat, line 2926:
```suggestion
* TLD: `co.uk` (uWhoisd uses the `Mozilla public suffix list <https://publicsuffix.org/list/>`_ as a reference)
```

wagner-certat (changes requested):
I overlooked the markdown-links syntax previously. Otherwise I'm satisfied and I consider the PR ready for merge (just waiting for the change concerning `extra.whois.*`)

Rafiot:
I changed the doc and the code accordingly to trigger multiple whois queries. I assume the hostname in a URL is the same as the FQDN if present... I think it is a good assumption, but please let me know if it's not.

wagner-certat (approved):

wagner-certat:
Thanks! I squashed the commits because we had 11 single commits, many of them small.

wagner-certat:
And added a changelog entry with credits to you in 16e2c939f

PR #1923, opened 2021-05-01
Sieve: expression negation, nested if / mixed conditionals and actions, support for non-strings in add/add!/update

monoidic:
This PR has three parts:
* generic negation of arbitrary expressions, along with the removal of the list match negation
* nested if statements, plus mixed actions and actions in the same scope
* the attribute manipulation actions `add`, `add!` and `update` support non-string (bool/int/float) values

There is also an RFC-ish last commit, with some somewhat breaking changes.
* Some operators which were made redundant by negation (`!~` and `:notcontains`) are removed
* `:in` is used for list inclusion checks instead of `==`
* there's a clear split between single-value and multi-value matches, with a couple of new operators due to that: `:containsany` and `:regexin` for string matching against multiple regexes or string inclusions)

The last commit is not expected to be accepted at the moment. It'll break a bunch of stuff. Though comments on it would be appreciated.

### DIFF
--- a/docs/user/bots.rst
+++ b/docs/user/bots.rst
@@ -2598,8 +2598,7 @@ The sieve file contains an arbitrary number of rules of the form:
    }
 
 
-*Please note* that nesting if-statements is currently not possible. `ACTIONS`
-must contain one or more actions of the actions listed below.
+Nested if-statements and mixed if statements and rules in the same scope are possible.
 
 *Expressions*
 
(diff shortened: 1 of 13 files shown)


wagner-certat:
Wow, great work! I'm just missing the docs :D

> The last commit is not expected to be accepted at the moment. It'll break a bunch of stuff. Though comments on it would be appreciated.

I have no problems with breaking changes
- as long as they are well documented
- in major release (3.0 is ahead, so that's OK)
- if behaviour doesn't change in unexpected ways, in this case the behaviour of `==` with lists is such a candidate. However, I'm open for discussions on this item =)

monoidic:
@wagner-certat Thoughts?

wagner-certat (approved):

wagner-certat:
> Some small adjustments, like dropping the negation tests on list/set actions. I will add the breaking changes in a seperate PR, this is enough for one PR as is.

Thank you very much - again!

> @wagner-certat Thoughts?

We had public holidays yesterday (and sometimes I may be busy :) )

wagner-certat:
changelog entry added in 7ca54fb76

PR #1745, opened 2021-02-16
New Bot for Fireeye Appliances

CysihZ:
Referring to my mail I created a Parser& Bot which is able to Collect and parse data from fireeye mail and file analysis appliances.
We are collection sha1 and MD5 hashed and if there is network communication we are also collecting urls and domains.

PR #1929, opened 2021-05-04
Revised Fireeye bots

wagner-certat:
Based on #1745 

Please review and test @CysihZ


wagner-certat:
The latest push fixed the docs, which contained totally unrelated and already applied changes

PR #1932, opened 2021-05-05
smtp output: handle empty fieldnames parameter

wagner-certat:
by sending no attachment
includes docs and test case


wagner-certat:
Apparently I overlooked my own PR #1896 (but that still helps for merging it into develop)

schacht-certat (approved):

PR #1933, opened 2021-05-05
FIX: GHA docker workflow publish

### LINKED ISSUE #1931: Docker image build action fails in forks
https://github.com/wagner-certat/intelmq/runs/2501171812?check_suite_focus=true#step:9:7

The docker login of the docker image build action fails in forks as the secrets are not available:

```
Run docker/login-action@v1
  with:
    logout: true
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.6.13/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.6.13/x64/lib
Error: Username and password required
```
This step should be skipped if the credentials are not provided

waldbauer-certat:
Publishing a new docker image to certat/* namespace on dockerhub
will only run if repository name is certtools/intelmq. So forks
dont fail as they dont execute the publish commands.

Fixes #1931

Signed-off-by: Sebastian Waldbauer <waldbauer@cert.at>

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -52,6 +52,7 @@ jobs:
 
       - name: Login to DockerHub
         uses: docker/login-action@v1
+        if: github.repository == 'certtools/intelmq'
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


wagner-certat (approved):
Works!
Example: https://github.com/wagner-certat/intelmq/runs/2508162454

PR #1934, opened 2021-05-05
DOCS: Updated docker instructions

waldbauer-certat:
Adapted docker instructions as they are now a bit different than
our first release. We are publishing intelmq with its current
version on dockerhub now. Tag 1.0 will remain but is deprecated.


wagner-certat:
Can you please either open the PR to maintenance or cherry-pick this commit for maintenance afterwards? Thanks.

schacht-certat (approved):

PR #1939, opened 2021-05-10
ENH: Cache mixin

waldbauer-certat:
The Cache mixin replaces the lib.cache implementation. Its easier to inherit & keeps the code more clean.

### DIFF
--- a/intelmq/lib/mixins/cache.py
+++ b/intelmq/lib/mixins/cache.py
@@ -10,6 +10,7 @@ from typing import Any, Optional
 import redis
 import intelmq.lib.utils as utils
 
+
 class CacheMixin:
     __redis: redis.Redis = None
     redis_cache_host: str = "127.0.0.1"
@@ -18,7 +19,7 @@ class CacheMixin:
     redis_cache_ttl: int = 15
     redis_cache_password: Optional[str] = None
 
-    def __init__(self):
+    def __init__(self, **kwargs):
         if self.redis_cache_host.startswith("/"):
             kwargs = {"unix_socket_path": self.redis_cache_host}
         elif self.redis_cache_host.startswith("unix://"):
@@ -31,6 +32,7 @@ class CacheMixin:
             }
 
         self.__redis = redis.Redis(db=self.redis_cache_db, password=self.redis_cache_password, **kwargs)
+        super().__init__()
 
     def cache_exists(self, key: str):
         return self.__redis.exists(key)
--- a/intelmq/lib/mixins/cache.py
+++ b/intelmq/lib/mixins/cache.py
@@ -1,0 +1,61 @@
+""" CacheMixin for IntelMQ
+
+SPDX-FileCopyrightText: 2021 Sebastian Waldbauer
+SPDX-License-Identifier: AGPL-3.0-or-later
+
+CacheMixin is used for caching/storing data in redis.
+"""
+
+from typing import Any, Optional
+import redis
+import intelmq.lib.utils as utils
+
+class CacheMixin:
+    __redis: redis.Redis = None
+    redis_cache_host: str = "127.0.0.1"
+    redis_cache_port: int = 6379
+    redis_cache_db: int = 9
+    redis_cache_ttl: int = 15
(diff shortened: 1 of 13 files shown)


wagner-certat:
That fixes the shadowserver test for me:
```
diff --git a/intelmq/lib/mixins/cache.py b/intelmq/lib/mixins/cache.py
index c51ca346f..8bdf243f7 100644
--- a/intelmq/lib/mixins/cache.py
+++ b/intelmq/lib/mixins/cache.py
@@ -32,6 +32,7 @@ def __init__(self):
             }
 
         self.__redis = redis.Redis(db=self.redis_cache_db, password=self.redis_cache_password, **kwargs)
+        super().__init__()
 
     def cache_exists(self, key: str):
         return self.__redis.exists(key)
diff --git a/intelmq/lib/mixins/http.py b/intelmq/lib/mixins/http.py
index d78e266c5..6a6befb96 100644
--- a/intelmq/lib/mixins/http.py
+++ b/intelmq/lib/mixins/http.py
@@ -51,6 +51,7 @@ class HttpMixin:
     def __init__(self, **kwargs):
         self.logger.debug("Running HTTP Mixin initialization.")
         self.http_header = {}
+        super().__init__()
 
     def setup(self):
         self.logger.debug("Setting up HTTP Mixin.")
```

intelmq/lib/mixins/cache.py
```diff
@@ -0,0 +1,63 @@
...
+from typing import Any, Optional
+import redis
+import intelmq.lib.utils as utils
+
+
+class CacheMixin:
+    __redis: redis.Redis = None
+    redis_cache_host: str = "127.0.0.1"
+    redis_cache_port: int = 6379
+    redis_cache_db: int = 9
```
wagner-certat, line 18: This will now be the default value for all bots using caches. Previously, (pre-parameters times, with BOTS), every bot had a distinct DB:
```
$ grep redis_cache_db intelmq/bots/BOTS 
                "redis_cache_db": "5",
                "redis_cache_db": "5",
                "redis_cache_db": 12,
                "redis_cache_db": "5",
                "redis_cache_db": "5",
                "redis_cache_db": "6",
                "redis_cache_db": "10",
                "redis_cache_db": "7",
                "redis_cache_db": "11",
```
Now, if the user does not watch out for setting unique DB indices, the user may end up with unclear behaviour, as the hashes as used by multiple bots are equal.

schacht-certat: Maybe it makes sense to use only **one** DB for the cache and have the `CacheMixin` uses a separate table (list, dict, whatever...) for every bot using the bot id as identifier

wagner-certat: > Maybe it makes sense to use only **one** DB for the cache and have the `CacheMixin` uses a separate table (list, dict, whatever...) for every bot using the bot id as identifier

AFAIK this is not possible. Redis has databases (numeric) and inside you can use the [`SET`](https://redis.io/commands/set) command, there is not other layer which we can use.

Using different `redis_cache_db` values in the bots on class-layer solves the issue IMO. E.g. setting `redis_cache_db` in DeduplicatorExpert to 6, the values which we had before. And so on.

schacht-certat: > AFAIK this is not possible. Redis has databases (numeric) and inside you can use the [`SET`](https://redis.io/commands/set) command, there is not other layer which we can use.

We could make the bot-id part of the hash

wagner-certat: By prepending (or similar) the botid to the key in `cache_exists`, `cache_get` and `cache_set`? If all bots use the same DB, then we can't flush the cache (and state) for the bots anymore, as *all* bots with state would be affected.

waldbauer-certat: If we would make the bot-id part of the hash, then we could flush the database. We couldnt use the "flushdb" command from redis, but we could write a routine to flush all keys which are part from the bot identified by the hash.

wagner-certat (approved):

PR #1945, opened 2021-05-11
FIX: RDAP checks if entity is valid entity-type

### LINKED ISSUE #1942: RDAP expert: Handle null value in entities
The RDAP expert should handle null values in the entities object like these by ignoring it:
```
 "entities": [
    {
      ...
    },
    null,
    {
      ...
    }
  ],
```
Plus a test case (can be a modified existing one instead of a new one)

waldbauer-certat:
As some RDAP implementation may vary, we check if the entity
is a valid entity-type. As of the RFC for RDAP only JSON dicts
are allowed.

Closes #1942

### DIFF
--- a/intelmq/tests/bots/experts/rdap/test_data/rdns.json
+++ b/intelmq/tests/bots/experts/rdap/test_data/rdns.json
@@ -1,0 +1,43 @@
+{
+    "description": "RDAP bootstrap file for Domain Name System registrations",
+    "publication": "2021-05-11T21:00:02Z",
+    "services": [
+      [
+        [
+          "nowruz",
+          "pars",
+          "shia",
+          "tci",
+          "xn--mgbt3dhd"
+        ],
+        [
+          "https://api.rdap.agitsys.net/"
+        ]
+      ],
+      [
+        [
+          "xn--p1acf"
+        ],
+        [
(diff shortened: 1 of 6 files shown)


intelmq/tests/bots/experts/rdap/test_data/rdns.json
```diff
@@ -0,0 +1,6841 @@
+{
+    "description": "RDAP bootstrap file for Domain Name System registrations",
+    "publication": "2021-05-11T21:00:02Z",
+    "services": [
+      [
+        [
+          "nowruz",
```
wagner-certat, line 7: Do we really need the complete file with nearly 7k lines in the test data?

wagner-certat (commented):
Code works:
```
rdap-expert: Received message {'source.fqdn': 'XYZ'}.
rdap-expert: Invalid type '<class 'NoneType'>' in entities of response for domain 'XYZ' found.
rdap-expert: Sending message to path '_default'.
```

intelmq/tests/bots/experts/rdap/test_data/rdns.json
waldbauer-certat: No, we dont need it. But I copied it 1 by 1, gonna crop it :)

wagner-certat (approved):

PR #1946, opened 2021-05-11
TST: simplify threshold tests

wagner-certat:
don't sleep, just flush the database
set the common parameters once
use the dedicated test database, works around a bug:
BotTestCase.setUpClass sets up the cache, but it doesn't know of the
redis database which the test will use. So let's just use the same.

fixes a test fail of #1939

### DIFF
--- a/intelmq/tests/bots/experts/threshold/test_expert.py
+++ b/intelmq/tests/bots/experts/threshold/test_expert.py
@@ -29,10 +29,10 @@ for msg in EVENTS_OUT:
 PARAMETERS = {
     'filter_keys': ['source.ip'],
     'filter_type': 'whitelist',
-    'redis_cache_db': '11',
+    'redis_cache_db': 4,
     'redis_cache_host': '127.0.0.1',
     'redis_cache_password': None,
-    'redis_cache_port': '6379',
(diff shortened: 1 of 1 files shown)


intelmq/tests/bots/experts/threshold/test_expert.py
```diff
@@ -48,61 +48,57 @@ class TestThresholdExpertBot(test.BotTestCase, unittest.TestCase):
...
+        self.run_bot(iterations=len(self.input_message))
         self.assertOutputQueueLen(0)
 
     def test_threshold_reached(self):
         """
         Ten messages, just enough to reach the threshold.
         """
-        time.sleep(2)  # Ensure cache has timed out
+        self.cache.flushdb()
+        print('keys', self.cache.keys())
```
waldbauer-certat, line 67: Debug?

waldbauer-certat (changes requested):

intelmq/tests/bots/experts/threshold/test_expert.py
wagner-certat: Thx

waldbauer-certat (approved):
LGTM

PR #1957, opened 2021-05-20
ENH: Sieve: Drop redundant operators, split string and numeric matches into single- and multivalued variants

monoidic:
This commit:
* drops the `:notcontains` and `!~` from Sieve, as they are made redundant by negation (`! foo :contains 'x'` == `foo :notcontains 'x'`)
* splits string and numeric matches into single- and multivalued variants, with the relevant new operators `:in`, `:containsany` and `:regexin` for string lists, and `:in` for numeric value lists
* adds documentation and tests for the above
* adds a note to the documentation regarding breaking changes

Should the breaking changes be referenced somewhere besides the documentation as well?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,20 @@ IntelMQ now uses YAML for the runtime configuration and therefore needs the `rua
 Both the XMPP collector bot and the XMPP output bot were removed. This [was evaluated on the mailinglist](https://lists.cert.at/pipermail/intelmq-users/2020-October/000177.html)
 and the XMPP bots were deprecated in 391d625.
 
+The Sieve expert bot has had major updates to its syntax. Breaking new changes:
+* the removal of the `:notcontains` operator, which can be replaced using the newly added
+ expression negation, e.g `! foo :contains ['.mx', '.zz']` rather than `foo :notcontains ['.mx', '.zz']`. 
+* changed operators for comparisons against lists of values, e.g `source.ip :in ['127.0.0.5', '192.168.1.2']` rather than `source.ip == ['127.0.0.5', '192.168.1.2']`
+
+New features:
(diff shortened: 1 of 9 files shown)


wagner-certat:
>     * drops the `:notcontains` and `!~` from Sieve, as they are made redundant by negation (`! foo :contains 'x'` == `foo :notcontains 'x'`)

Thinking a bit more about the `!~`, some programming / configuration have such an operator (ruby, perl, apache, lighttpd) so it is not so uncommon. Would that be a reason to keep it? (like the `!=` operator)
Input welcome, e.g. @schacht-certat @waldbauer-certat 

> Should the breaking changes be referenced somewhere besides the documentation as well?

Yes, could you please add a short summary in the NEWS.md file? That calls the attention of users (at least the fraction which reads the announcements and/or docs).

monoidic:
> > ```
> > * drops the `:notcontains` and `!~` from Sieve, as they are made redundant by negation (`! foo :contains 'x'` == `foo :notcontains 'x'`)
> > ```
> 
> Thinking a bit more about the `!~`, some programming / configuration languages have such an operator (ruby, perl, apache, lighttpd) so it is not so uncommon. Would that be a reason to keep it? (like the `!=` operator)
> Input welcome, e.g. schacht-certat waldbauer-certat

Hmm. Yeah, I guess `:notcontains` is the only odd one out ¯\\\_(ツ)\_/¯
I was familiar with `=~`, but not `!~`, though since the latter has an established meaning, I agree that it should be kept. I believe `:notcontains` is still a bit redundant though.

wagner-certat (approved):

wagner-certat:
Thank you for another super clean and perfect PR! I love reviewing your changes, all is done with perfection :)

PR #1962, opened 2021-05-26
Add RPZ file output bot

mariuskarotkis:
**Add new bot: RPZ file output**

Example output file:

```
$TTL 3600
@ SOA rpz.yourdomain.eu. hostmaster.rpz.yourdomain.eu. 2105260601 60 60 432000 60
NS localhost.
;
; yourdomain.eu. CERT.XX Response Policy Zones (RPZ)
; Last updated: 2021-05-26 06:01:41 (UTC)
;
; Terms Of Use: https://rpz.yourdomain.eu
; For questions please contact rpz [at] yourdomain.eu
;
*.maliciousdomain.com CNAME rpz.yourdomain.eu.
*.secondmaliciousdomain.com CNAME rpz.yourdomain.eu.
```

### DIFF
--- a/intelmq/bots/outputs/rpz_file/output.py
+++ b/intelmq/bots/outputs/rpz_file/output.py
@@ -1,4 +1,12 @@
 # -*- coding: utf-8 -*-
+
+"""
+RPZ file output
+
+SPDX-FileCopyrightText: 2021 Marius Karotkis <marius.karotkis@gmail.com>
+SPDX-License-Identifier: AGPL-3.0-or-later
+"""
+
 from datetime import datetime
 import os
 from collections import defaultdict
@@ -10,7 +18,7 @@ RPZ_INDICATOR_MAP = {
     "source.fqdn": "Intel::DOMAIN"
 }
 
-now = datetime.now() # for timestamp
+now = datetime.now()  # for timestamp
 
 
 class RpzFileOutputBot(OutputBot):
@@ -95,17 +103,17 @@ class RpzFileOutputBot(OutputBot):
             if 'time.observation' in ev:
                 try:
                     ev['time.observation'] = datetime.strptime(ev['time.observation'],
-                                                                        '%Y-%m-%dT%H:%M:%S+00:00')
+                                                               '%Y-%m-%dT%H:%M:%S+00:00')
                 except ValueError:
                     ev['time.observation'] = datetime.strptime(ev['time.observation'],
-                                                                        '%Y-%m-%dT%H:%M:%S.%f+00:00')
+                                                               '%Y-%m-%dT%H:%M:%S.%f+00:00')
             if 'time.source' in ev:
                 try:
                     ev['time.source'] = datetime.strptime(ev['time.source'],
-                                                                   '%Y-%m-%dT%H:%M:%S+00:00')
+                                                          '%Y-%m-%dT%H:%M:%S+00:00')
                 except ValueError:
                     ev['time.source'] = datetime.strptime(ev['time.source'],
-                                                                   '%Y-%m-%dT%H:%M:%S.%f+00:00')
+                                                          '%Y-%m-%dT%H:%M:%S.%f+00:00')
             filename = self.file.format(event=ev)
             if not self.file or filename != self._file.name:
                 self.open_file(filename)
--- a/intelmq/bots/outputs/rpz_file/output.py
+++ b/intelmq/bots/outputs/rpz_file/output.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-import datetime
 from datetime import datetime
 import os
 from collections import defaultdict
@@ -11,7 +10,7 @@ RPZ_INDICATOR_MAP = {
     "source.fqdn": "Intel::DOMAIN"
 }
 
-now = datetime.now()
+now = datetime.now() # for timestamp
 
 
 class RpzFileOutputBot(OutputBot):
@@ -49,7 +48,6 @@ class RpzFileOutputBot(OutputBot):
         self.set_rpz_header()
 
         self.logger.debug("Opening %r file.", self.file)
-        self.format_filename = getattr(self, 'format_filename', False)
         self.errors = getattr(self, 'encoding_errors_mode', 'strict')
         if not self.format_filename:
             self.open_file(self.file)
@@ -91,17 +89,17 @@ class RpzFileOutputBot(OutputBot):
             ev.update(event)
             if 'time.observation' in ev:
                 try:
-                    ev['time.observation'] = datetime.datetime.strptime(ev['time.observation'],
+                    ev['time.observation'] = datetime.strptime(ev['time.observation'],
                                                                         '%Y-%m-%dT%H:%M:%S+00:00')
                 except ValueError:
-                    ev['time.observation'] = datetime.datetime.strptime(ev['time.observation'],
+                    ev['time.observation'] = datetime.strptime(ev['time.observation'],
                                                                         '%Y-%m-%dT%H:%M:%S.%f+00:00')
             if 'time.source' in ev:
                 try:
-                    ev['time.source'] = datetime.datetime.strptime(ev['time.source'],
+                    ev['time.source'] = datetime.strptime(ev['time.source'],
                                                                    '%Y-%m-%dT%H:%M:%S+00:00')
                 except ValueError:
-                    ev['time.source'] = datetime.datetime.strptime(ev['time.source'],
+                    ev['time.source'] = datetime.strptime(ev['time.source'],
                                                                    '%Y-%m-%dT%H:%M:%S.%f+00:00')
             filename = self.file.format(event=ev)
             if not self.file or filename != self._file.name:
--- a/intelmq/bots/outputs/rpz_file/output.py
+++ b/intelmq/bots/outputs/rpz_file/output.py
@@ -57,16 +57,16 @@ class RpzFileOutputBot(OutputBot):
 
     def set_rpz_header(self):
         generate_time: str = now.strftime("%Y-%m-%d %H:%M:%S")
-        self._rpz_header = f"""$TTL " + str(self.ttl) + "\n
-                @ SOA " + self.rpz_domain + ". " + self.hostmaster_rpz_domain + ". " + str(self.serial) + " " + str(self.refresh) + " " + str(self.retry) + " " + str(self.expire) + " " + str(self.ncachttl) + "\n
-                NS localhost.\n
-                ;\n
-                ; " + self.organization_name + " Response Policy Zones (RPZ)\n
-                ; Last updated: " + str(self.generate_time) + " (UTC)\n
-                ;\n
-                ; Terms Of Use: https://" + self.rpz_domain + "\n
-                ; For questions please contact " + self.rpz_email + "\n
-                ;\n"""
+        self._rpz_header = f"""$TTL """ + str(self.ttl) + """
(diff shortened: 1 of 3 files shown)


wagner-certat:
Hi @mariuskarotkis thank you for contributing so much new stuff =) Looks all very good on the first glance.

Unfortunately I fear we won't have time for the reviews this week, so we'll need to keep you waiting a few days. But be assured that we don't ignore you :)

wagner-certat:
You can be sure that we haven't forgot your PRs, and we're currently working hard. Unfortunately we need to hold a deadline for IntelMQ 3.0 and as we're short-staffed currently, we've had to postpone some activities, like reviewing the PRs. If I won't have time for a review before 3.0, we'll put your additions into 3.1 of course. Sorry for these circumstances :/

intelmq/bots/outputs/rpz_file/output.py
```diff
@@ -0,0 +1,154 @@
+# -*- coding: utf-8 -*-
+import datetime
+from datetime import datetime
```
wagner-certat, line 3: That overwrites/conflicts

intelmq/bots/outputs/rpz_file/output.py
```diff
@@ -0,0 +1,154 @@
...
+from collections import defaultdict
+from pathlib import Path
+
+from intelmq.lib.bot import OutputBot
+
+RPZ_INDICATOR_MAP = {
+    "source.fqdn": "Intel::DOMAIN"
+}
+
+now = datetime.now()
```
wagner-certat, line 14: `now` will be the time when the bot was started. Is that intentional? if yes, please add an inline-comment why.

I'd have expected this in the init or process method.

intelmq/bots/outputs/rpz_file/output.py
```diff
@@ -0,0 +1,154 @@
...
+
+class RpzFileOutputBot(OutputBot):
+    _file = None
+    format_filename: bool = False
+    hierarchical_output: bool = False
+    keep_raw_field: bool = False
+    message_jsondict_as_string: bool = False
+    message_with_type: bool = False
+    single_key: bool = False
+    is_multithreadable = False
```
wagner-certat, line 25: That should work as intended:
```suggestion
    __Bot_is_multithreadable = False
```
(but I see that using a private member was not the best choice for this, as this requires the special naming in cases like this one - but that's a different discussion)

intelmq/bots/outputs/rpz_file/output.py
```diff
@@ -0,0 +1,154 @@
...
+    encoding_errors_mode = 'strict'
+    file: str = "/opt/intelmq/var/lib/bots/file-output/rpz"
+
+    cname: str = ""
+    organization_name: str = ''
+    rpz_domain: str = ''
+    hostmaster_rpz_domain: str = ''
+    rpz_email: str = ''
+    ttl: str = '3600'
+    generate_time: str = now.strftime("%Y-%m-%d %H:%M:%S")
```
wagner-certat, line 36: Is that a configuration parameter? If yes, why does its default value depend on the time the bot is created / configured?

intelmq/bots/outputs/rpz_file/output.py
```diff
@@ -0,0 +1,154 @@
...
+
+    encoding_errors_mode = 'strict'
+    file: str = "/opt/intelmq/var/lib/bots/file-output/rpz"
+
+    cname: str = ""
+    organization_name: str = ''
+    rpz_domain: str = ''
+    hostmaster_rpz_domain: str = ''
+    rpz_email: str = ''
+    ttl: str = '3600'
```
wagner-certat, line 35: Why as string?

intelmq/bots/outputs/rpz_file/output.py
```diff
@@ -0,0 +1,154 @@
...
+    rpz_header = ""
+
+    def init(self):
+        self._file = None
+
+        self.set_rpz_header()
+
+        self.logger.debug("Opening %r file.", self.file)
+        self.format_filename = getattr(self, 'format_filename', False)
+        self.errors = getattr(self, 'encoding_errors_mode', 'strict')
```
wagner-certat, line 53: the default values are already set above

wagner-certat: And `encoding_errors_mode` is not used

intelmq/bots/outputs/rpz_file/output.py
```diff
@@ -0,0 +1,154 @@
...
+        if not self.format_filename:
+            self.open_file(self.file)
+        self.logger.info("File %r is open.", self.file)
+
+    def set_rpz_header(self):
+        self.rpz_header = "$TTL " + str(self.ttl) + "\n@ SOA " + self.rpz_domain + ". " + self.hostmaster_rpz_domain + ". " + str(self.serial) + " " + str(self.refresh) + " " + str(self.retry) + " " + str(self.expire) + " " + str(self.ncachttl) + "\n NS localhost.\n;\n; " + self.organization_name + " Response Policy Zones (RPZ)\n; Last updated: " + str(self.generate_time) + " (UTC)\n;\n; Terms Of Use: https://" + self.rpz_domain + "\n; For questions please contact " + self.rpz_email + "\n;\n"
```
wagner-certat, line 59: Please format this string by other means to make it more readable, e.g. using format-strings, like
```python
self.rpz_header = f"""$TTL {self.ttl}
@ SOA {self.rpz_domain}. {self.hostmaster_rpz_domain}. {self.serial}"""
```
etc.

intelmq/bots/outputs/rpz_file/output.py
```diff
@@ -0,0 +1,154 @@
...
+    ttl: str = '3600'
+    generate_time: str = now.strftime("%Y-%m-%d %H:%M:%S")
+    ncachttl: str = '60'
+    serial: str = now.strftime("%y%m%d%H%M")
+    refresh: str = '60'
+    retry: str = '60'
+    expire: str = '432000'
+    test_domain: str = ''
+
+    rpz_header = ""
```
wagner-certat, line 44: is that a configuration parameter?

intelmq/bots/outputs/rpz_file/output.py
```diff
@@ -0,0 +1,154 @@
...
+            except IOError:
+                self.logger.exception('Directory %r could not be created.', path)
+                self.stop()
+            else:
+                self._file = open(filename, mode='a+t', encoding='utf-8', errors=self.errors)
+                self.add_rpz_header()
+
+    def add_rpz_header(self):
+        self._file.seek(0)
+        if self._file.read() != self.rpz_header and not os.stat(self._file.name).st_size:
```
wagner-certat, line 83: I don't understand the logic of this line. First, you check that the *complete* file is unequal to the previously generated RPZ header, then you stat by the filename to ensure it is empty. Wouldn't a size-check suffice then? (which can be done with the `Path` object as well, btw)

wagner-certat (changes requested):
What I am primarily missing is documentation for the bot which explains what it does, how it does it, how it can be configured and what can be achieved with it (linking to external sources is perfectly fine).

intelmq/bots/outputs/rpz_file/output.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+class RpzFileOutputBot(OutputBot):
+    _file = None
+    format_filename: bool = False
+    hierarchical_output: bool = False
+    keep_raw_field: bool = False
+    message_jsondict_as_string: bool = False
+    message_with_type: bool = False
+    single_key: bool = False
```
wagner-certat, line 24: They seem to be unused

intelmq/bots/outputs/rpz_file/output.py
mariuskarotkis: It's for output file timestamp (generate_time, serial).

intelmq/bots/outputs/rpz_file/output.py
mariuskarotkis: Yes, it is configuration parameter. generate_time it is like timestamp, if regenerate file this value need change.

intelmq/bots/outputs/rpz_file/output.py
mariuskarotkis: Because these parameters only for header.

intelmq/bots/outputs/rpz_file/output.py
mariuskarotkis: configuration parameter:

expire: str = '432000'
test_domain: str = ''

intelmq/bots/outputs/rpz_file/output.py
mariuskarotkis: Just check if the file is empty.

intelmq/bots/outputs/rpz_file/output.py
mariuskarotkis: I'm copy params from another bot.

intelmq/bots/outputs/rpz_file/output.py
wagner-certat: But the allowed values are supposedly only ints?

intelmq/bots/outputs/rpz_file/output.py
wagner-certat: If not, then let's mark it as private member
```suggestion
    _rpz_header = ""
```

intelmq/bots/outputs/rpz_file/output.py
wagner-certat: Sorry, I still don't understand.
You say that this is a configuration parameter. It's default value is "now". Let's assume, that a user configures an instance of the RPZ output. The default value is `2021-07-16 18:17:24`, which looks good to the user and sets that. As a result, the generated file always has the exact same timestamp. It could be interpreted as "the time when the user has configured this bot" or "initial creation date".
In the header, this value is used for "Last updated" which has - for me - a different meaning. "Last update" could be the time when the last entry was added to the file.

intelmq/bots/outputs/rpz_file/output.py
wagner-certat: Then let's remove them, if they are not used.
```suggestion
```

intelmq/bots/outputs/rpz_file/output.py
wagner-certat: Does this work as well?
```suggestion
        if not (len(self._file.read())):
```

intelmq/bots/outputs/rpz_file/output.py
mariuskarotkis: Changed to int.

intelmq/bots/outputs/rpz_file/output.py
mariuskarotkis: move to function, it is not configuration  parameter now.

intelmq/bots/outputs/rpz_file/output.py
```diff
@@ -0,0 +1,165 @@
...
+    "source.fqdn": "Intel::DOMAIN"
+}
+
+now = datetime.now()  # for timestamp
+
+
+class RpzFileOutputBot(OutputBot):
+    _file = None
+    format_filename: bool = False
+    __Bot_is_multithreadable = False
```
ghost, line 27:
```suggestion
    _is_multithreadable = False
```
That was recently fixed/changed in 5c91a8c9b59541c147774a727d000d18c5dc1d2e you could not have foreseen this

wagner-certat (approved):
Thanks. Apart from the minor change it looks good overall
I'll apply the change myself after merging, then we can finally conclude this as well. Sorry for the long waiting - with volatile priorities, it's not always easy to have enough time for the reviews. But I think that the current state is much better than what we had before.

PR #1965, opened 2021-05-26
Add new bot: cut string from string

mariuskarotkis:
Add new bot: cut string from string

### DIFF
--- a/intelmq/bots/experts/cut_from_string/expert.py
+++ b/intelmq/bots/experts/cut_from_string/expert.py
@@ -8,23 +8,20 @@ from intelmq.lib.bot import Bot
 class CutFromStringExpertBot(Bot):
     string_from_start: bool = True  # True - from start, False - from end
     string_for_cut: str = 'www.'
-    field_for_cut: str = 'source.fqdn'
-
-    def init(self):
-        pass
+    field: str = 'source.fqdn'
 
     def process(self):
         event = self.receive_message()
 
-        if self.field_for_cut in event:
-            field_string = event[self.field_for_cut]
+        if self.field in event:
+            field_string = event[self.field]
             if self.string_from_start == 1 and field_string.startswith(self.string_for_cut):
                 field_string = field_string[len(self.string_for_cut):]
-                event.change(self.field_for_cut, field_string)
+                event.change(self.field, field_string)
 
             if self.string_from_start == 0 and field_string.endswith(self.string_for_cut):
                 field_string = field_string[:-len(self.string_for_cut)]
-                event.change(self.field_for_cut, field_string)
+                event.change(self.field, field_string)
 
         self.send_message(event)
         self.acknowledge_message()
--- a/intelmq/bots/experts/cut_from_string/expert.py
+++ b/intelmq/bots/experts/cut_from_string/expert.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+"""
+Cut from string
+"""
+from intelmq.lib.bot import Bot
+
+
+class CutFromStringExpertBot(Bot):
+    string_from_start: int = 1  # 1 - from start, 0 - from end
+    string_for_cut: str = 'www.'
+    field_for_cut: str = 'source.fqdn'
+
+    def init(self):
+        pass
+
+    def process(self):
+        event = self.receive_message()
+
+        if self.field_for_cut in event:
+            field_string = event[self.field_for_cut]
+            if self.string_from_start == 1 and field_string.startswith(self.string_for_cut):
+                field_string = field_string[len(self.string_for_cut):]
+                event.change(self.field_for_cut, field_string)
+
+            if self.string_from_start == 0 and field_string.endswith(self.string_for_cut):
+                field_string = field_string[:-len(self.string_for_cut)]
+                event.change(self.field_for_cut, field_string)
+
+        self.send_message(event)
+        self.acknowledge_message()
+
+
+BOT = CutFromStringExpertBot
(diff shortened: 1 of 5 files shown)


intelmq/bots/experts/cut_from_string/expert.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
+"""
+Cut from string
+"""
+from intelmq.lib.bot import Bot
+
+
+class CutFromStringExpertBot(Bot):
+    string_from_start: int = 1  # 1 - from start, 0 - from end
```
wagner-certat, line 9: If the valid values for these parameters are just 0 and 1 why not use a boolean value? From a user-perspective, maybe string values like `cut_from = 'start'` are better? (just thinking aloud)

wagner-certat (changes requested):

wagner-certat (changes requested):
Please add documentation in `docs/user/bots.rst`

intelmq/bots/experts/cut_from_string/expert.py
mariuskarotkis: Good point.
It can be improved.

schacht-certat: Given that there are only two possible values, it should IMO be a boolean

intelmq/bots/experts/cut_from_string/expert.py
```diff
@@ -0,0 +1,33 @@
...
+    string_for_cut: str = 'www.'
+    field_for_cut: str = 'source.fqdn'
+
+    def init(self):
+        pass
+
```
schacht-certat, line 15: If this method doesn't do anything, we maybe simple can remove it?

intelmq/bots/experts/cut_from_string/expert.py
```diff
@@ -0,0 +1,33 @@
...
+                field_string = field_string[len(self.string_for_cut):]
+                event.change(self.field_for_cut, field_string)
+
+            if self.string_from_start == 0 and field_string.endswith(self.string_for_cut):
+                field_string = field_string[:-len(self.string_for_cut)]
+                event.change(self.field_for_cut, field_string)
+
+        self.send_message(event)
+        self.acknowledge_message()
+
```
schacht-certat, line 31: PEP616 lists example code for the `removeprefix` and `removesuffix` String methods- maybe we can use those here and add a comment to drop them once we only support Python versions that implement those methods:
```
def removeprefix(self: str, prefix: str, /) -> str:
    if self.startswith(prefix):
        return self[len(prefix):]
    else:
        return self[:]

def removesuffix(self: str, suffix: str, /) -> str:
    # suffix='' should not call self[:-0].
    if suffix and self.endswith(suffix):
        return self[:-len(suffix)]
    else:
        return self[:]
```

then the code would look like this:
```
if self.removeprefix:
    event.change(self.field, self.removeprefix(self.field, self.affix)
else:
    event.change(self.field, self.removesuffix(self.field, self.affix)
```

intelmq/bots/experts/cut_from_string/expert.py
```diff
@@ -0,0 +1,33 @@
...
+"""
+Cut from string
+"""
+from intelmq.lib.bot import Bot
+
+
+class CutFromStringExpertBot(Bot):
+    string_from_start: int = 1  # 1 - from start, 0 - from end
+    string_for_cut: str = 'www.'
+    field_for_cut: str = 'source.fqdn'
```
schacht-certat, line 11: I think `field_for_cut` should simply be called `field` - there are already some other bots that operate on fields, that use the variable name `field`

schacht-certat (changes requested):
If I understand correctly, the bot will remove a prefix or a suffix from a field. So I would call it RemoveAffix. There is also [PEP616](https://www.python.org/dev/peps/pep-0616/) which proposes `str` methods which do exactly that and which seems to have been [accepted](https://bugs.python.org/issue39939) for Python 3.9. So once we only support Python 3.9, we can use that (maybe add a comment so that we don't forget).

Please add a reuse compliant license and copyright header to the files, an example can be found in the [documentation](https://intelmq.readthedocs.io/en/latest/dev/guide.html#template)

intelmq/bots/experts/cut_from_string/expert.py
mariuskarotkis: I will remove it.

mariuskarotkis:
Updated

mariuskarotkis:
> If I understand correctly, the bot will remove a prefix or a suffix from a field. So I would call it RemoveAffix. There is also [PEP616](https://www.python.org/dev/peps/pep-0616/) which proposes `str` methods which do exactly that and which seems to have been [accepted](https://bugs.python.org/issue39939) for Python 3.9. So once we only support Python 3.9, we can use that (maybe add a comment so that we don't forget).
> 
> Please add a reuse compliant license and copyright header to the files, an example can be found in the [documentation](https://intelmq.readthedocs.io/en/latest/dev/guide.html#template)

Please check.

ghost:
I think the only discussion point left open is this comment/proposal by @schacht-certat:
> If I understand correctly, the bot will remove a prefix or a suffix from a field. So I would call it RemoveAffix.

@mariuskarotkis What do you think of this idea? Does something indicate against it?

mariuskarotkis:
> RemoveAffix

It's good for me. I think it can be renamed.

wagner-certat (approved):

ghost:
Thank you very much - again! - for this useful addition.

PR #1967, opened 2021-05-26
Add new bot: cut string by delimeter, max length

mariuskarotkis:
Add new bot: cut string by delimeter, max length


intelmq/bots/experts/string_cut_delimiter/expert.py
```diff
@@ -0,0 +1,29 @@
...
+"""
+Cut string if length is bigger than max
+"""
+from intelmq.lib.bot import Bot
+
+
+class StringCutDelimiterExpertBot(Bot):
+    string_delimiter: str = '.'
+    max_length: int = 200
+    field_for_cut: str = 'source.fqdn'
```
schacht-certat, line 11: Please rename this attribute to `field`

intelmq/bots/experts/string_cut_delimiter/expert.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def init(self):
+        pass
+
```
schacht-certat, line 15: Please remove the init, if it is not supposed to do anything

intelmq/bots/experts/string_cut_delimiter/expert.py
```diff
@@ -0,0 +1,29 @@
...
+    def process(self):
+        event = self.receive_message()
+
+        if self.field_for_cut in event:
+            long_string = event[self.field_for_cut]
+            while (long_string.find(self.string_delimiter) != -1) and (len(long_string) > self.max_length):
```
schacht-certat, line 21: `long_string.find(self.string_delimiter) != -1` can be written as `self.string_delimiter in long_string`

intelmq/bots/experts/string_cut_delimiter/expert.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
+"""
+Cut string if length is bigger than max
+"""
+from intelmq.lib.bot import Bot
+
+
+class StringCutDelimiterExpertBot(Bot):
+    string_delimiter: str = '.'
```
schacht-certat, line 9: I think it would be oke if we just name that attribute `delimiter`

intelmq/bots/experts/string_cut_delimiter/expert.py
```diff
@@ -0,0 +1,29 @@
...
+    def init(self):
+        pass
+
+    def process(self):
+        event = self.receive_message()
+
+        if self.field_for_cut in event:
+            long_string = event[self.field_for_cut]
+            while (long_string.find(self.string_delimiter) != -1) and (len(long_string) > self.max_length):
+                long_string = long_string.split(self.string_delimiter, 1)[1]
```
schacht-certat, line 22: I think it is more readable if you write:
`long_string, _ = long_string.rsplit(self.string_delimiter, 1)`

schacht-certat (changes requested):
Making a string shorter in the Python world is usually called `truncating` (see for example the [jinja2 truncate](https://jinja.palletsprojects.com/en/3.0.x/templates/#jinja-filters.truncate) filter), so calling the bot something like `TruncateByDelimiter` is maybe a more descriptive name.

Please also add a reuse compliant license and copyright header to the files.

As a long term solution it might make more sense to implement a separate bot for string field manipulation, like a JinjaBot which then can do all the jinja filters and which would make it easier to add custom filters. @wagner-certat what do you think?

wagner-certat (changes requested):
Could you please add a license header like
```
# SPDX-FileCopyrightText: 2018 <your name/mailaddress>
#
# SPDX-License-Identifier: AGPL-3.0-or-later
```
to
* intelmq/bots/experts/truncate_by_delimiter/expert.py
* intelmq/tests/bots/experts/truncate_by_delimiter/test_expert.py
Thanks

Please describe better, how the bot works so that it's not necessary to study the code before using it. It's not clear from the current docs how delimiter and max_length work together. Maybe add some usage examples which demonstrate it's capabilities, e.g. with max_length = 0 and max_length > 0, and a.b.c.example.com with max_length = 13 and = 15.

wagner-certat (approved):

wagner-certat:
Thanks again for your contribution! And sorry for the long waiting time - several issues at the same time forced us to set the focus elsewhere :/

PR #1968, opened 2021-05-26
ENH: Mixin & once behaviour

waldbauer-certat:
Mixins are initialized before the bot init, this is mandatory
for the mixins, because mixins have to be initialized before
the bot gets initialized. Otherwise the bot would raise errors
because mixins cant access data.

once() is a newly introduced function and will run once after bot
start. This function can be used for some after start cleanups or
pre-data collection from pipelines.


wagner-certat:
Obsoleted by #1982

Issue #569, opened 2016-07-14
Filter bot 


sebix:
During a workshop we collected some feedback on needed filtering possibilities. We now came up with this list of useful comparisons. Could be implemented by an abstract base class and derived specific classes.

I used pseudo code for easier reading.
## Generic filter

The filter bot shall support these generic filtering comparisons:

_Match any key in the event against a value with these comparison operators_:
- [x] `==`
- [ ] `<`,`>`
- [ ] `<=`,`>=`
- [ ] `!=`
- [ ] `in`, e.g. `source.asn in [1234, 1235]`
- [x] `~=`, matching: extended regex, with `re.search` (#676)
- exists/`is not None`
- not exists/`is None`

## Ip specific filtering

In particular, the filter bot shall have a derived class "IPAddressFilter" which can compare:
_source.ip, destination.ip, source.network, destination.network_ against a value with these comparators:
- [x] `ip = '192.0.2.1'`
- [ ] `ip in '192.0.2.1-192.0.2.10'`
- [ ] `ip in ['192.0.2.1', '192.0.2.2']`
- [ ] `ip in '192.0.2.0/24'`
- [ ] `ip in ['192.0.2.1-192.0.2.10', '192.0.2.100/28']`
- [ ] for `source.ip` and `destination.ip`
## Time specific filtering:

Again this is a derived class. 
Compare time.source, time.observation via:
- [x] `<`,`>`
- [ ] `<=`,`>=`
- [x] `==`
- [ ] `!=`
- [x] absolute
- [x] relative

comments appreciated

EDIT 2016-07-18: Added exists
EDIT 2016-09-07: ip-filtering: different entries in lists

ondj:
Because not every message has every field, "field exists" and "field does not exist" filters are necessary (or handle not exists fields as null).

And then at least `and`, `or` and `not` operators are useful for advanced filtering.

aaronkaplan:
> On 18 Jul 2016, at 14:28, ondj <notifications@github.com> wrote:
> 
> Because not every message has every field, "field exists" and "field does not exist" filters are necessary (or handle not exists fields as null).
> 
> And then at least and, or and not operators are useful for advanced filtering.
> 
Agreed.

dmth:
An enhancement of the filters would be great. Do these ideas also include a "choice of output queues", for example: `if ip in 192.0.2.1/28 then my-special-expert-queue else normal-expert-queue`?

Can this be realised?

sebix:
[...]

PR #1969, opened 2021-05-26
Add new bot: time filter

mariuskarotkis:
Add new bot: time filter

### DIFF
--- a/intelmq/bots/experts/time_filter/expert.py
+++ b/intelmq/bots/experts/time_filter/expert.py
@@ -1,0 +1,55 @@
+# -*- coding: utf-8 -*-
+from datetime import datetime, timedelta
+from dateutil import parser
+from intelmq.lib.utils import get_timedelta
+from intelmq.lib.bot import Bot
+from datetime import timezone
+
+
+class TimeFilterExpertBot(Bot):
+    search_field: str = 'time.source'
+    search_from: str  = '1d'
+
+    not_after = None
+
+    def init(self):
+        self.search_field = self.search_field
+        if self.search_from:
+            timedelta_params = get_timedelta(self.search_from)
+            self.not_after = datetime.now(tz=timezone.utc) - timedelta(**timedelta_params)
+
+    def process(self):
+        event = self.receive_message()
+        # time based filtering
+        if self.search_field in event:
+            try:
+                event_time = parser.parse(str(event.get(self.search_field)))
+            except ValueError:
+                event_time = self.not_after
+                self.process_message(event_time, event)
+                return
+            else:
+                self.process_message(event_time, event)
+                return
+        else:
+            # not found field
+            event_time = self.not_after
+            self.process_message(event_time, event)
+            return
+
+    def process_message(self, event_time, event):
+        event_time = event_time.replace(tzinfo=None)
+        self.not_after = self.not_after.replace(tzinfo=None)
+
+        if event_time < self.not_after:
+            self.acknowledge_message()
+            self.logger.debug(
+                f"Filtered out event with search field {self.search_field} and event time {event_time} .")
+            return
+        else:
+            self.send_message(event)
+            self.acknowledge_message()
+            return
+
+
+BOT = TimeFilterExpertBot
--- a/intelmq/tests/bots/experts/time_filter/test_expert.py
+++ b/intelmq/tests/bots/experts/time_filter/test_expert.py
@@ -1,0 +1,106 @@
+# -*- coding: utf-8 -*-
+
+import unittest
+from freezegun import freeze_time
+
+import intelmq.lib.test as test
+from intelmq.bots.experts.time_filter.expert import TimeFilterExpertBot
+
+EXAMPLE_INPUT_DROP = {
+    "__type": "Event",
+    "feed.accuracy": 90.0,
+    "feed.name": "Feodo Tracker IPs",
+    "feed.provider": "abuse.ch",
+    "feed.url": "https://feodotracker.abuse.ch/downloads/ipblocklist.csv",
+    "time.observation": "2020-10-13T06:14:49+00:00",
+    "raw": "dGVzdA==",
+    "extra.firstseen": "2020-10-11T02:10:59+00:00",
+    "source.port": 447,
+    "extra.lastonline": "2020-08-13T00:00:00+00:00",
+    "malware.name": "trickbot",
+    "time.source": "2020-10-13T00:00:00+00:00"
+}
+EXAMPLE_INPUT_PASS = {
+    "__type": "Event",
+    "feed.accuracy": 90.0,
+    "feed.name": "Feodo Tracker IPs",
+    "feed.provider": "abuse.ch",
+    "feed.url": "https://feodotracker.abuse.ch/downloads/ipblocklist.csv",
+    "time.observation": "2020-10-13T06:14:49+00:00",
+    "raw": "dGVzdA==",
+    "extra.firstseen": "2020-10-11T02:10:59+00:00",
+    "source.port": 447,
+    "extra.lastonline1": "2020-09-13T00:00:00+00:00",
+    "malware.name": "trickbot",
+    "time.source": "2020-10-13T00:00:00+00:00"
+}
+EXAMPLE_INPUT_PASS_2 = {
+    "__type": "Event",
+    "feed.accuracy": 90.0,
+    "feed.name": "Feodo Tracker IPs",
+    "feed.provider": "abuse.ch",
+    "feed.url": "https://feodotracker.abuse.ch/downloads/ipblocklist.csv",
+    "time.observation": "2020-10-13T06:14:49+00:00",
+    "raw": "dGVzdA==",
+    "extra.firstseen": "2020-10-11T02:10:59+00:00",
+    "source.port": 447,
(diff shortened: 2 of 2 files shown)


wagner-certat:
We already have the "filter" expert which can filter on (non-)equality, non-existence and regular expressions (match and non-match). We also have the sieve expert with filtering capabilities on all of the above plus existence, substrings, numeric comparisons, network ranges, set/list operations and all with typing support.

I'm reflecting what is the best user experience
1. having a dedicated bot for filtering only on time (con: causes the bot-list to grow with yet-another filter expert), or
2. expanding the existing filter expert to time-filtering (see also #569), (con: makes the configuration of the bot more complex), and/or
3. expanding the sieve expert to time filtering

1 and 2 contradict each other, while 3 is orthogonal. I'm currently tending towards 1 (which is, what this PR does), as 2 means a more complex configuration of the bot - but I want to sleep over it first and hear others' opinions. And 3, our swiss-army knife, would be cool as well =)

intelmq/bots/experts/time_filter/expert.py
```diff
@@ -0,0 +1,55 @@
+# -*- coding: utf-8 -*-
+from datetime import datetime, timedelta
+from dateutil import parser
+from intelmq.lib.utils import get_timedelta
```
schacht-certat, line 4: This function does not exist

intelmq/bots/experts/time_filter/expert.py
```diff
@@ -0,0 +1,55 @@
...
+from intelmq.lib.bot import Bot
+from datetime import timezone
+
+
+class TimeFilterExpertBot(Bot):
+    search_field: str = 'time.source'
```
schacht-certat, line 10: Please rename this attribute to `field`

intelmq/bots/experts/time_filter/expert.py
```diff
@@ -0,0 +1,55 @@
+# -*- coding: utf-8 -*-
+from datetime import datetime, timedelta
+from dateutil import parser
+from intelmq.lib.utils import get_timedelta
+from intelmq.lib.bot import Bot
+from datetime import timezone
+
+
+class TimeFilterExpertBot(Bot):
```
schacht-certat, line 9: Please add a docstring explaining what this bot does

intelmq/bots/experts/time_filter/expert.py
```diff
@@ -0,0 +1,55 @@
...
+    search_field: str = 'time.source'
+    search_from: str  = '1d'
```
schacht-certat, line 11: Why is this called `search_from`? Isn't this the time span in which events should be filtered? Then I would call the attribute `timespan`

intelmq/bots/experts/time_filter/expert.py
```diff
@@ -0,0 +1,55 @@
...
+
+    not_after = None
```
schacht-certat, line 13: This is not a user configurable variable, so please make it private (prepend `__` like `__not_after`). But IIRC a timedelta can also contain a negative value, so we could make call this attribute simply `__delta`, that way we can filter for events in the future and the past.

intelmq/bots/experts/time_filter/expert.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def init(self):
+        self.search_field = self.search_field
```
schacht-certat, line 16: I guess this can be removed

intelmq/bots/experts/time_filter/expert.py
```diff
@@ -0,0 +1,55 @@
...
+        if self.search_from:
+            timedelta_params = get_timedelta(self.search_from)
+            self.not_after = datetime.now(tz=timezone.utc) - timedelta(**timedelta_params)
+
+    def process(self):
+        event = self.receive_message()
+        # time based filtering
+        if self.search_field in event:
+            try:
+                event_time = parser.parse(str(event.get(self.search_field)))
```
schacht-certat, line 26: the timebased fields in the event already contain a `datetimevalue`, so you're converting those to string and back to `datetime` - is there a specific reason for this?

intelmq/bots/experts/time_filter/expert.py
```diff
@@ -0,0 +1,55 @@
...
+                self.process_message(event_time, event)
+                return
+            else:
+                self.process_message(event_time, event)
+                return
+        else:
+            # not found field
+            event_time = self.not_after
+            self.process_message(event_time, event)
+            return
```
schacht-certat, line 38: If you set `event_time = self.not_after` in the beginning of `process` and combine the three `process_message` calls to one and move it at the end you can remove both `else`-clauses

intelmq/bots/experts/time_filter/expert.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def process_message(self, event_time, event):
+        event_time = event_time.replace(tzinfo=None)
+        self.not_after = self.not_after.replace(tzinfo=None)
```
schacht-certat, line 42: Harmonization sanitizes all datetime objects to `UTC` timezone, is there a specific reason to remove the timezone from those datafields?

intelmq/bots/experts/time_filter/expert.py
```diff
@@ -0,0 +1,55 @@
...
+        if event_time < self.not_after:
+            self.acknowledge_message()
+            self.logger.debug(
+                f"Filtered out event with search field {self.search_field} and event time {event_time} .")
+            return
+        else:
+            self.send_message(event)
+            self.acknowledge_message()
+            return
+
```
schacht-certat, line 53: Please move the `acknowledge_message` call down after the if/else statement and remove the `return`s

schacht-certat (changes requested):
I think that a dedicated time filtering bot would be the best solution, but we have to think twice about which features this bot should have, to make it not too complicated to configure.

wagner-certat:
I not sure yet what the purpose of this bot is due to missing docs, however I want to point out named queued ("paths") in IntelMQ. Using them could extend the functionality of the bot, by keeping the configuration simple. Thay are used in the "normal" filter expert like this: https://intelmq.readthedocs.io/en/latest/user/bots.html#filter (section Possible paths) and https://github.com/certtools/intelmq/blob/develop/intelmq/bots/experts/filter/expert.py#L105-L143

For example, with a single parameter (timespan), multiple use-cases could be solved, all used in parallel by the IntelMQ user
- path `older_than` receives all data matching `event < now()-timespan`
- path `newer_than` receives all data matching `event > now()-timespan`
- path `older_than_future` receives all data matching `event < now()+timespan`
- path `newer_than_future` receives all data matching `event > now()+timespan`

or similar.

intelmq/bots/experts/time_filter/expert.py
mariuskarotkis: This is for a two-date check.

```
event_time = event_time.replace(tzinfo=None)
        self.__delta = self.__delta.replace(tzinfo=None)

        if event_time > self.__delta:
```

intelmq/bots/experts/time_filter/expert.py
mariuskarotkis: Some field can be string, not date

wagner-certat:
@mariuskarotkis The tests (or other Actions) of all your remaining PRs are failing currently. Could you please have a look at them? Please also apply the lessons learnt from the review of the merged PRs - if applicable to the new ones. If a PR is ready for review from your side, please give us a short ping.

mariuskarotkis:
Please check tests.

wagner-certat:
The pycodestyle check fails with
```
intelmq/lib/utils.py:925:13: E701 multiple statements on one line (colon)
```
For the nosetests, you introduced a new testing requirement `freeze_time`. But that does not exist: https://pypi.org/project/freeze_time/ and causes the check to fail. Maybe you meant `freezegun`? However, we already use `time-machine` is a similar test case: https://github.com/certtools/intelmq/blob/develop/intelmq/tests/bots/experts/aggregate/test_expert.py We evaluated various libraries for simulating the time and that was the best matching result. It would also support decorators: https://github.com/adamchainz/time-machine#readme I'd prefer if we can stick to one time travelling library to keep the maintenance low - if possible.

intelmq/tests/bots/experts/time_filter/test_expert.py
```diff
@@ -0,0 +1,121 @@
...
+    "raw": "dGVzdA==",
+    "extra.firstseen": "2020-10-11T02:10:59+00:00",
+    "source.port": 447,
+    "extra.lastonline": "2020-09-13",
+    "malware.name": "trickbot",
+    "time.source": "2020-10-13T00:00:00+00:00"
+}
+
+
+@test.skip_exotic()
```
wagner-certat, line 78: You could put all this in a if-clause to fix the error
```suggestion
if time_machine:
    @test.skip_exotic()
    ...
```

mariuskarotkis:
@sebix Please check reviewer is active or no.

sebix:
Can you answer this question in the docs?

> When I read the docs of this new bot I don't understand the difference to the existing filter expert using `not_before` = `2 days` (the value of the `timespan` parameter here). Can you please elaborate on docs more, so that the advantage / difference to the filter expert is recognizable?

PR #1970, opened 2021-05-27
ENH: add Shadowserver scan_exchange config & tests

schacht-certat:


wagner-certat:
Some follow-up changes in fab5ab0ba

PR #1971, opened 2021-05-27
Shadowserver: tweaks, add event-sinkhole-http-referer support

monoidic:
This PR tweaks the Shadowserver parser bot a bit, mainly by avoiding iterating over the entire mapping. It also adds support for the new `event-sinkhole-http-referer` feed.
The first change reveals that `scan_http` appears as a filename in both the `Accessible-HTTP` and `Vulnerable-HTTP` feeds. The way the bot is implemented before this PR, if searched by filename, it defaults to `Accessible-HTTP`, while after, it defaults to `Vulnerable-HTTP`. The mapping could be reordered, though it would just give a different incorrect answer than the current one. Is there a good way to solve this, such as by removing filename-based matching for `(Vulnera|Accessi)ble-HTTP`? Otherwise, people may end up with incorrectly marked `feed.name`.


wagner-certat:
Thanks! I cherry-picked f9f9554d18fea22be15afa3939e3607c2263adea for maintenance (without the change in `test_broken`) so that we can release the next bugfix version. I'll look at the other changes after the release.

intelmq/tests/bots/parsers/shadowserver/test_broken.py
```diff
@@ -49,7 +49,7 @@ def test_broken(self):
         self.assertLogMatches(pattern="Detected report's file name: 'scan_http'.",
                               levelname="DEBUG")
         self.assertLogMatches(pattern="Failed to parse line.")
-        self.assertLogMatches(pattern="ValueError: Required column 'timestamp' not found in feed 'Accessible-HTTP'. Possible change in data format or misconfiguration.")
+        self.assertLogMatches(pattern="ValueError: Required column 'timestamp' not found in feed 'Vulnerable-HTTP'. Possible change in data format or misconfiguration.")
```
wagner-certat, line 52: Why is that change necessary?

monoidic: This is what the second paragraph of the original post discusses.
>  The first change reveals that scan_http appears as a filename in both the Accessible-HTTP and Vulnerable-HTTP feeds. The way the bot is implemented before this PR, if searched by filename, it defaults to Accessible-HTTP, while after, it defaults to Vulnerable-HTTP. The mapping could be reordered, though it would just give a different incorrect answer than the current one. Is there a good way to solve this, such as by removing filename-based matching for (Vulnera|Accessi)ble-HTTP? Otherwise, people may end up with incorrectly marked feed.name.

tl;dr `scan_http` appears twice in `mapping`, and my solution matches the last `scan_http` while the previous solution matches the first one, neither of which really seem "correct" to me. This test assumes the previous behaviour of matching the first `scan_http`.

wagner-certat: Sorry, should have read or remembered that - fully understood. The circumstance itself is not very fortunate as it causes the feeds to be detected wrongly, but that's independent of this PR.

I'll open separate issue regarding this.

wagner-certat:
Thanks again for your work!

Due to the conflicts, merged "by hand" in 94c21f5e3
Created #1984 for the incomplete feed detection as discussed above

PR #1979, opened 2021-06-06
DOCS: Updated release procedure for docker images

waldbauer-certat:


schacht-certat:
I think the list should also describe how to update the submodules

wagner-certat (changes requested):

wagner-certat (approved):

PR #1980, opened 2021-06-07
ENH: replace copy2 with pathlib

schacht-certat:
copy2 tries to copy file attributes. This leads to problems when we try
to use the write_configuration function from code that runs under
another user, even when the user is in the intelmq group.
Simple copying the content of the file should be enough, though.

### DIFF
--- a/intelmq/lib/utils.py
+++ b/intelmq/lib/utils.py
@@ -241,7 +241,8 @@ def write_configuration(configuration_filepath: str,
         if content == old_content:
             return None
     if not new and backup:
-        shutil.copy2(configuration_filepath, configuration_filepath + '.bak')
+        config = pathlib.Path(configuration_filepath)
+        pathlib.Path(configuration_filepath + '.bak').write_text(config.read_text())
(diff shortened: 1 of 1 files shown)


wagner-certat:
I think I chose copy2 because the restore from backup is less troublesome, a simple `mv` is enough, as the file permissions are already correct. When creating a new file, the group write permissions probably is not set, causing differing file permissions for the original file and the backup file.

schacht-certat:
> I think I chose copy2 because the restore from backup is less troublesome, a simple `mv` is enough, as the file permissions are already correct. When creating a new file, the group write permissions probably is not set, causing differing file permissions for the original file and the backup file.

Oke, that would mean the configuration can/should only be written by the intelmq user itself and thus we would have to find a way to run the API as the intelmq user

wagner-certat:
Nah, that sounds even more troublesome. Haven't had in mind that the routine is called by the API backend, as webserver user. Let's go the way you proposed.

wagner-certat (approved):

PR #1959, opened 2021-05-25
ENH: Aggregate Expert

waldbauer-certat:
This expert is able to aggregate events with a defined threshold.

How does it work?
The expert will use specified fields from an event as identifier.
Then stores how often the event occur and if the threshold has
been reached within its set timespan, it will push the event to
the queue, else it will just get dropped.

PR #1982, opened 2021-06-07
ENH: lib/bot: Init pipeline before init call

wagner-certat:
allows bots accessing data directly on startup/initialization for cleanup or maintenance tasks
this required small adaptions in the test code, and shortens its code a
bit

related: certtools/intelmq#1968
related: certtools/intelmq#1959

@waldbauer-certat  please verify if this solves your issue

### DIFF
--- a/intelmq/lib/bot.py
+++ b/intelmq/lib/bot.py
@@ -199,6 +199,9 @@ class Bot(object):
             self.__load_harmonization_configuration()
 
             self._parse_common_parameters()
+
+            super().__init__()
+            self.__connect_pipelines()
             self.init()
 
             if not self.__instance_id:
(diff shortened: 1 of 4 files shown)


schacht-certat (approved):

intelmq/lib/bot.py
```diff
@@ -199,6 +199,8 @@ def handle_sighup_signal_threading(signum: int,
             self.__load_harmonization_configuration()
 
             self._parse_common_parameters()
+
```
waldbauer-certat, line 202: `super().__init__()` has to be initialized before the `self.init()`, because Mixins have to be available before the init is called.

waldbauer-certat (changes requested):

intelmq/lib/bot.py
wagner-certat: Thanks, sounds logical

wagner-certat: Fixed

waldbauer-certat (approved):

PR #1989, opened 2021-06-14
Aggregate expert Test and docs fixes

wagner-certat:
Please especially check if `counter_ignored` is implemented correctly

### DIFF
--- a/intelmq/bots/experts/aggregate/expert.py
+++ b/intelmq/bots/experts/aggregate/expert.py
@@ -31,13 +31,19 @@ class AggregateExpertBot(Bot, CacheMixin):
 
     def cleanup(self):
         if self.__next_cleanup <= time.time():
+            self.logger.debug('Started Cleanup.')
             delta = timedelta(minutes=self.__timespan)
+
+            counter_sent = 0
+            counter_ignored = 0
+            counter_dropped = 0
             for key in self.cache_get_redis_instance().keys(pattern="aggregate.*"):
                 keys = self.cache_get_redis_instance().hgetall(key)
                 data = {y.decode('utf-8'): keys.get(y).decode('utf-8')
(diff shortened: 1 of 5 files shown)


intelmq/bots/experts/aggregate/expert.py
```diff
@@ -46,8 +51,14 @@ def cleanup(self):
                     event.add("extra.count", int(data['c']))
                     event.add("extra.time_end", data['l'])
                     self.send_message(event)
+                    counter_sent += 1
+                else:
+                    counter_ignored += 1
```
wagner-certat, line 56: Maybe this case here is "message gets discarde because threshold was not reached"?

waldbauer-certat: Yes, if threshold is not reached, message gets dropped.

waldbauer-certat (approved):

intelmq/bots/experts/aggregate/expert.py
```diff
@@ -46,8 +51,14 @@ def cleanup(self):
...
                 self.cache_get_redis_instance().delete(key)
             self.__next_cleanup = int(time.time()) + 10
+            self.logger.debug('Completed Cleanup. Messages sent: %d, messages ignored: %d.', counter_sent, counter_ignored)
+        else:
+            self.logger.debug('Skipped Cleanup (%fs < 10s).', self.__next_cleanup - time.time())
```
waldbauer-certat, line 61: This might get spammy in the logs, because each event received will run the cleanup process.

wagner-certat: OTOH it's the debug log, which logs *at least* one line for every incoming message anyway

waldbauer-certat (approved):

PR #1992, opened 2021-06-16
harm: move dga domain classification to other

wagner-certat:
move "dga domain" to the taxonomy other, as it is not accepted in RSIT
rename to "dga-domain" to match the name scheme of other types
adapt bots depending on the name
add changelog and news entries, including SQL update statements

### DIFF
--- a/intelmq/lib/harmonization.py
+++ b/intelmq/lib/harmonization.py
@@ -282,7 +282,7 @@ class ClassificationType(String):
                       'ddos',
                       'ddos-amplifier',
                       'defacement',
-                      'dga domain',
+                      'dga-domain',
                       'dos',
                       'dropzone',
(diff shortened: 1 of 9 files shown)


intelmq/lib/harmonization.py
```diff
@@ -282,7 +282,7 @@ class ClassificationType(String):
                       'ddos',
                       'ddos-amplifier',
                       'defacement',
-                      'dga domain',
+                      'dga-domain',
```
wagner-certat, line 285: ￼￼forgot to add backwards compatibility on this section. Will add that tomorrow

wagner-certat: done

waldbauer-certat (approved):

PR #2010, opened 2021-06-30
BUG: utils/load_configuration: fix loading JSON with tabs

wagner-certat:
ruamel can't load JSON files with tabs as whitespace
if this specific excpetion is thrown, try loading the file with
json.load instead

this adds some tests for the load_configuration function
with JSON files and YAML file


waldbauer-certat (approved):
LGTM

PR #2020, opened 2021-07-20
FIX: s/self.parameters/self/ in Shodan collector

monoidic:
The `error_max_retries` attribute is no longer accessed via `self.parameters` and the Shodan collector appears to have been missed in the updates.


wagner-certat (approved):
Thanks!

wagner-certat:
Also cherry-picked for maintenance in 7f1eda7d3

PR #2033, opened 2021-08-10
Make update script less verbose by default

### LINKED ISSUE #2016: Bot database update scripts: Be quiet by default in cron scripts
Currently, the bot database update scripts print this on every call if nothing is done:

```
Database update skipped. No bots of type intelmq.bots.experts.asn_lookup.expert present in runtime.conf.
```

I think a good default would be to be quiet by default on the cron runs, e.g. by setting a `--quiet` flag. If the user wants to have verbose information, it can always be turned on.

Or: honour the logging level as defined for the bot. With the configuration overhaul, this should now be easier to do on the code-side.

schacht-certat:
This introduces a --verbose flag to the experts which use update
functions. The default behaviour now is to *not* print any messages,
which makes the update function less noisy when used with cron.

Fixes #2016

In the long run we should probably move the update logic to the bot class itself.

### DIFF
--- a/intelmq/bots/experts/asn_lookup/expert.py
+++ b/intelmq/bots/experts/asn_lookup/expert.py
@@ -78,7 +78,7 @@ class ASNLookupExpertBot(Bot):
             parsed_args = cls._create_argparser().parse_args()
 
         if parsed_args.update_database:
-            cls.update_database()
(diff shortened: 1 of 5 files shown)


wagner-certat (approved):

wagner-certat:
Also cherry-picked for maintenance

Issue #1875, opened 2021-04-19
AMQP pipeline: arguments are ignored


wagner-certat:
https://github.com/certtools/intelmq/blob/e54b47027a2100b37f7fe415433dfe2001171ce1/intelmq/lib/pipeline.py#L436-L457

The amqp class accesses `self` to get the parameters, while the redis class uses `self.pipeline_args`. The amqp pipeline does not receive any parameters because of that bug.

PR #2034, opened 2021-08-11
FIX: load pipeline configurations before connecting

monoidic:
This appears to fix an issue where the `host` attribute of the pipeline is not set before being accessed while connecting pipelines, causing a bot to fail to start, as shown in this stack trace:
```
  File "/usr/local/lib/python3.8/dist-packages/intelmq/lib/bot.py", line 208, in __init__
    self.__connect_pipelines()
  File "/usr/local/lib/python3.8/dist-packages/intelmq/lib/bot.py", line 579, in __connect_pipelines
    self.__destination_pipeline.connect()
  File "/usr/local/lib/python3.8/dist-packages/intelmq/lib/pipeline.py", line 204, in connect
    if self.host.startswith("/"):
AttributeError: 'Redis' object has no attribute 'host'
```

### DIFF
--- a/intelmq/lib/bot.py
+++ b/intelmq/lib/bot.py
@@ -563,6 +563,7 @@ class Bot(object):
                                                             load_balance=self.load_balance,
                                                             is_multithreaded=self.is_multithreaded)
 
+            self.__source_pipeline.load_configurations('source')
             self.__source_pipeline.connect()
             self.__current_message = None
             self.logger.debug("Connected to source queue.")
@@ -576,6 +577,7 @@ class Bot(object):
                                                                  load_balance=self.load_balance,
                                                                  is_multithreaded=self.is_multithreaded)
 
+            self.__destination_pipeline.load_configurations('destination')
             self.__destination_pipeline.connect()
             self.logger.debug("Connected to destination queues.")
         else:
--- a/intelmq/lib/pipeline.py
+++ b/intelmq/lib/pipeline.py
@@ -177,6 +177,9 @@ class Pipeline(object):
     def _reject_message(self):
         raise NotImplementedError
 
+    def load_configurations(self, queues_type):
+        raise NotImplementedError
+
 
 class Redis(Pipeline):
     has_internal_queues = True
@@ -405,6 +408,11 @@ class Pythonlist(Pipeline):
         """
         No-op because of the internal queue
         """
+
+    def load_configurations(self, queues_type):
+        '''
+        Nothing to load
+        '''
 
 
 class Amqp(Pipeline):


wagner-certat:
Under which circumstances did you encounter this bug?

monoidic:
I'm not fully certain at the moment, actually.
It was while running IntelMQ 3.0 with the Redis backend, and seemingly only manifested itself in output bots. Manually setting the `Host` parameter in the manager or setting it to a different value (such as swapping between 127.0.0.1 and 127.0.0.2) was a temporary workaround, which lasted for some indefinite period of time.

wagner-certat:
`PipelineFactory.create` calls `set_queues`, which then calls `load_configurations`.

A comment in `intelmq.tests.lib.test_pipeline_TestRedis` class reminded me of #1875

wagner-certat:
Still debugging...

In https://github.com/certtools/intelmq/commit/56df092a578cc4c69a8c6a565364cdb2c1612c50 I rewrote the redis test to better match the behavior in Bot, using the same methods and arguments. If I set a different host (e.g. with `INTELMQ_PIPELINE_HOST` the test times out), so that *seems* to work in principle.

monoidic:
In case you are unable to reproduce it, I can test out whatever you want.
After applying this patch:
```patch
diff --git a/intelmq/lib/pipeline.py b/intelmq/lib/pipeline.py
index 6c276b0f..e0933856 100644
--- a/intelmq/lib/pipeline.py
+++ b/intelmq/lib/pipeline.py
@@ -57,6 +57,8 @@ def create(logger, broker=None, direction=None, queues=None, pipeline_args=None,
                 broker = broker.title()
             else:
                 broker = "Redis"
+
+        logger.info(f'{queues=} {direction=}')
         pipe = getattr(intelmq.lib.pipeline, broker)(logger=logger, pipeline_args=pipeline_args, load_balance=load_balance, is_multithreaded=is_multithreaded)
         if queues and not direction:
             raise ValueError("Parameter 'direction' must be given when using "
```
I got the following lines in a bot's log file, when attempting to start it:

```
2021-08-16 16:47:37,029 - elasticsearch-output - INFO - ElasticsearchOutputBot initialized with id elasticsearch-output and intelmq 3.0.0 and python 3.8.10 (default, Jun  2 2021, 10:49:15) as process 551040.
2021-08-16 16:47:37,029 - elasticsearch-output - INFO - Bot is starting.
2021-08-16 16:47:37,586 - elasticsearch-output - INFO - queues='elasticsearch-output-queue' direction='source'
2021-08-16 16:47:37,586 - elasticsearch-output - INFO - queues={} direction='destination'
2021-08-16 16:47:37,586 - elasticsearch-output - ERROR - Bot initialization failed.
  File "/usr/local/lib/python3.8/dist-packages/intelmq/lib/bot.py", line 208, in __init__
    self.__connect_pipelines()
  File "/usr/local/lib/python3.8/dist-packages/intelmq/lib/bot.py", line 581, in __connect_pipelines
    self.__destination_pipeline.connect()
  File "/usr/local/lib/python3.8/dist-packages/intelmq/lib/pipeline.py", line 208, in connect
    if self.host.startswith("/"):
AttributeError: 'Redis' object has no attribute 'host'
2021-08-16 16:47:37,587 - elasticsearch-output - INFO - Bot stopped.
```

So it appears to happen with the destination queue specifically, and retrying it a couple of times gives identical log lines.

wagner-certat:
I just got it reproduced locally, with an output bot, as you said.

wagner-certat:
`set_queues` and `load_configurations` is correctly called for the source pipeline. I think the primary problem is, that the code also loads the destination pipeline which doesn't make sense and obviously fails.

In `__connect_pipelines`:
https://github.com/certtools/intelmq/blob/fb03cccaa32f7cf40082b910569574e38628d118/intelmq/lib/bot.py#L570
the check is `is not None`. The empty default value is:
https://github.com/certtools/intelmq/blob/fb03cccaa32f7cf40082b910569574e38628d118/intelmq/lib/bot.py#L67
also `None`. However, something between sets the variable to an empty dict, which `is not None` and then the bot tries to connect to nothing.
Haven't found that code yet, but maybe an empty dict is a better default value anyway, also because it is assumed in the code, e.g. here:
https://github.com/certtools/intelmq/blob/fb03cccaa32f7cf40082b910569574e38628d118/intelmq/lib/bot.py#L392

wagner-certat:
> Haven't found that code yet,

Ah, that's the loading of the runtime configuration. That makes sense. If the variable is not defined, the error does not occur, but if the configuration has for example:
```
    destination_queues: {}
```
Then it occurs.

monoidic:
Sounds like it's related to some of the changes I've made in the manager then.

monoidic:
And, indeed, I have `destination_queues: {}` in the `runtime.yaml` on those bots, due to some of my new code in intelmq-manager setting the field to an empty object by default in intelmq-manager so its existence does not have to be checked for when adding new edges.
Now the question seems to be what the solution should be.
* intelmq-manager should clear empty `destination_queues` from the configuration sent to the backend,
* `destination_queues: {}` should be handled as a valid option indicating no output queues,
* both?

wagner-certat:
I favour the second option, which is also simple fixable. The core should handle any valid configuration, independent of the tool which created the config. Such a config could also be written by a human.

@schacht-certat What's your opinion?

schacht-certat:
> I favour the second option, which is also simple fixable. The core should handle any valid configuration, independent of the tool which created the config. Such a config could also be written by a human.
> 
> @schacht-certat What's your opinion?

Yes, handling this in intelmq itself is the better option, `destination_queues: {}` should be handled the same as if it was `None` (or we simply set it to an empty dict by default and remove the checks for `None`)

wagner-certat:
based on what we found out here in this thread, i created https://github.com/certtools/intelmq/pull/2051 with a smaller change

monoidic:
Fixed properly by #2051

Issue #2003, opened 2021-06-25
ruamel.yaml loader and dumper: human readability / support for comments


ghost:
277f72d019eab234501a3e91176186e673b880ba fixed the issue that comments where not preserved.
However when the runtime configuration is loaded with the default loader, as it was with the comment, and then dumped afterwards, the resulting file is formatted uglily:
```
{file-input: {bod_id: file-input, description: Reads file /assets/*.txt, group: Collectors,
name: File Input, enabled: true, module: intelmq.bots.collectors.file.collector_file,
run_mode: scheduled, parameters: {path: /assets, postfix: .txt, delete_file: false,...
```
Therefore I reverted the commit with 9a6946120c16505c1440081e850196415bb23ade, which brings back a human readable and better structured format for the `runtime.yaml`. But OTOH it breaks comments again. We need a solution to preserve both, comments and readability.

wagner-certat:
Maybe we use the dump-method wrongly?

sebix:
Comments are still not preserved, even though we use `YAML(typ='unsafe', pure=True)`

PR #2041, opened 2021-08-16
Use the YAML unsafe loader instead of the safe loader

schacht-certat:
This lets us be backwards compatible to JSON and still dump YAML in a
readable format.

Another approach would be to first try to load JSON and only if that fails use yaml with the default loader.


wagner-certat:
fixes #2003

wagner-certat (approved):

PR #2042, opened 2021-08-16
Implement a clear class hierarchy

schacht-certat:


wagner-certat (changes requested):
I'd appreciate changelog entries for the changes (new file `intelmq/lib/datatypes.py`, new mixin, new class `ExpertBot`, parent classes for all bots changed, ...). Otherwise: :tada:

intelmq/lib/datatypes.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class BotType(str, Enum):
+    COLLECTOR   = "Collector"
+    PARSER      = "Parser"
+    EXPERT      = "Expert"
+    OUTPUT      = "Output"
+
+    def toJson(self):
+        return self.value
```
wagner-certat, line 16: Doesn't that miss the quotes necessary for making it a JSON-string?

wagner-certat:
Could you please also update the dev guide (sections Template & Mixins are affected as I see with a quick search)

intelmq/lib/datatypes.py
schacht-certat: `toJson` is used by the JSON dump method to serialize individual values. In the case of the `BotType` class the dump method does not know how to represent the class (i.e. when dumping a bots attributes). Given that `self.value` is a string, the JSON dump method know how to handle the rest

wagner-certat: Thanks for the explanation!

wagner-certat (approved):

PR #2044, opened 2021-08-16
DOCS: Added better description for trusted introducer

### LINKED ISSUE #2017: Trusted introducer expert: Missing doc on processed fields
https://github.com/certtools/intelmq/pull/1978#pullrequestreview-678110190
> Please add to the docs which information is looked for and which is information is saved in which field. And that the first match wins.

Still missing. Please submit to maintenance.

waldbauer-certat:
Fixes #2017

### DIFF
--- a/docs/user/bots.rst
+++ b/docs/user/bots.rst
@@ -3218,7 +3218,12 @@ Trusted Introducer Lookup Expert
 
 **Configuration Parameters**
 
-* **order**: Possible values are 'domain', 'asn'
+* **order**: Possible values are 'domain', 'asn'. You can set multiple values, so first match wins.
+* If 'domain' is set, it will lookup the `source.fqdn` field. It will go from high-order to low-order, i.e. 1337.super.example.com -> super.example.com -> example.com -> `.com`
+* If 'asn' is set, it will lookup `source.asn`.
+
+After a match, the abuse contact will be fetched from the trusted introducer teams list and will be stored in the event as `source.abuse_contact`.
+If there is no match, the event will not be enriched and will be sent to the next configured step.
 
 
 .. _intelmq.bots.experts.tuency.expert:


wagner-certat (changes requested):
The information, what is stored where in the message is still missing.

docs/user/bots.rst
```diff
@@ -3218,7 +3218,12 @@ Trusted Introducer Lookup Expert
 
 **Configuration Parameters**
 
-* **order**: Possible values are 'domain', 'asn'
+* **order**: Possible values are 'domain', 'asn'. You can set multiple values, so first match wins.
+  * If 'domain' is set, it will lookup the `source.fqdn` field. It will go from high-order to low-order, i.e. 1337.super.example.com -> super.example.com -> example.com -> *.com
+  * If 'asn' is set, it will lookup `source.asn`.
+
+After a match, then the abuse contact will be fetched from the trusted introducer teams list and will be stored in the event as `source.abuse_contact`.
```
schacht-certat, line 3225: *After a match, the abuse contact will ...* (without the `then`)

docs/user/bots.rst
```diff
@@ -3218,7 +3218,12 @@ Trusted Introducer Lookup Expert
...
+If there is a no match, the event will not be enriched and will be sent to the next configured step.
```
schacht-certat, line 3226: *If there is no match...* (without the `a`)

schacht-certat (changes requested):

wagner-certat (approved):
what @schacht-certat said

schacht-certat:
Screenshot_2021-08-24 Bots — intelmq 3 0 1 beta1 documentation
This doesn´t render very readable. Not sure what you were going for with the `*` symbol in the beginning of the line, if the goal was to make a listing, you should remove the whitespaces in front of the `*`.
The single tick isn´t a special character in RST, maybe you wanted to use backticks instead, like you did with `source.fqdn`?
I would remove the `*` before the `.com` because it is interpreted as an anchor by RST.
Also you could use → instead of `->` to make it more readable.

schacht-certat (approved):

PR #2047, opened 2021-08-17
Add Honeypot HTTP Scan feed to Shadowserver parser

monoidic:
This PR adds the Honeypot HTTP Scan feed to the Shadowserver parser, along with the related sample and test.


intelmq/bots/parsers/shadowserver/_config.py
```diff
@@ -204,6 +205,18 @@ def convert_date_utc(value: str) -> Optional[str]:
...
+_base64_alphabet = set('ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=')
+
+
+def maybe_base64(value: Optional[str]) -> Optional[str]:
+    if not value:
+        return None
+    elif set(value).issubset(_base64_alphabet):
+        return value
+
+    return base64.b64encode(value.encode()).decode()
```
wagner-certat, line 217: Do you want to *decode from* base64 or *encode to* base64? Currently the code does encode, but that doesn't really make sense for me. If you want to decode, the latter should work:

```suggestion
def decode_base64(value: Optional[str]) -> Optional[str]:
    """
    Lazy decoding of base64-encoded data.
    If the character set indicates a base64-encoded content, decoding is attempted and the result returned.
    Otherwise, or if the decoding fails, the original string is returned.
    """
    if not value:
        return None
    elif set(value).issubset(_base64_alphabet):
        try:
            return base64.b64encode(value.encode()).decode()
        except binascii.Error:
            return Value
    else:
        return value
```

monoidic: Err. It's meant to be encode, actually, so as to keep the data consistently base64-encoded, whereas currently it's sometimes base64-encoded, sometimes not. Decoding arbitrary binary values from POST queries as UTF8 woudln't work very well.

wagner-certat: Ah, okay. Then I propose to chose a different function name, plus docstring and don't use the issubset for checking if the value is base64, but a try-except-block decoding the value.

intelmq/bots/parsers/shadowserver/_config.py
```diff
@@ -204,6 +206,22 @@ def convert_date_utc(value: str) -> Optional[str]:
...
+    uniformly converting the data in the field to be base64-encoded
+    """
+    if not value:
+        return None
+
+    try:
+        base64.b64decode(value)  # return value intentionally ignored
+        return value
+    except binascii.Error:
+        return base64.b64encode(value.encode()).decode()
```
wagner-certat, line 222: let's keep the try-block minimal to avoid unexpected behaviour (although the return statement shouldn't cause any direct troubles, having return inside try is not a good practice)

```suggestion
    try:
        base64.b64decode(value)  # return value intentionally ignored
    except binascii.Error:
        return base64.b64encode(value.encode()).decode()
    else:
        return value
```

wagner-certat (changes requested):
otherwise ready for merge (tests are appreciated but not required)

monoidic:
Ah, right. I'll add some tests on monday (friday's a national holiday).

monoidic:
I added `validate=True` to the decode step, as by default, the base64 module just silently ignores invalid characters, which is the opposite of the intended effect, and some tests.

wagner-certat (approved):

wagner-certat:
Thanks! also cherry-picked for maintenance (coming 3.0.1 release)

PR #2048, opened 2021-08-18
Remove workarounds for XMPP bots

schacht-certat:
XMPP based bots were removed in 3.0.0


wagner-certat (approved):
Some similar workaround is maybe needed for the n6stomp collector and output, but we'd need to validate that first.

PR #2049, opened 2021-08-18
Make the harmonization attribute of the bot private

schacht-certat:
that way we can exclude it from attribute listings of the bots
attributes.


intelmq/lib/bot.py
```diff
@@ -241,6 +243,9 @@ def catch_shutdown():
         if start:
             self.start()
 
+    def harmonization(self):
+        return self.__harmonization
+
```
wagner-certat, line 248: Just an idea: Why not use [`@property`](https://docs.python.org/3/library/functions.html#property)

wagner-certat (approved):

PR #2051, opened 2021-08-18
BUG: lib/bot: handle empty dict in destination_queues

wagner-certat:
the manager (or any user in the file directly) sets destination_queues
to an empty dict. the check `is not None` did not catch empty dicts,
initializing the destination pipeline even when no queues were defined.
The check is now more lax and the code handles empty dicts normally,
making them also the default value instead of None.

fixes certtools/intelmq#2034

@monoidic please check if that solves your observed error


monoidic:
> @monoidic please check if that solves your observed error

It does indeed, thanks!

schacht-certat (approved):

wagner-certat:
rebased to resolve conflict

PR #2054, opened 2021-08-19
ENH: lib/bot: fix behavior for unconfigured bots

wagner-certat:
Add `enabled` and `run_mode` as members of the `Bot` class with default values.
Log the system parameters in the same way as default and runtime parameters.
Warn if disallowed system parameters are encountered.
Warn if no configuration could be found in the runtime configuration file for the bot's ID.
Re-initialize logging if logging parameters have been changed by environment variables.

### DIFF
--- a/intelmq/lib/bot.py
+++ b/intelmq/lib/bot.py
@@ -42,6 +42,9 @@ from intelmq.lib.utils import RewindableFileHandle, base64_decode
 from intelmq.lib.datatypes import *
 
 __all__ = ['Bot', 'CollectorBot', 'ParserBot', 'OutputBot', 'ExpertBot']
+ALLOWED_SYSTEM_PARAMETERS = {'enabled', 'run_mode', 'group', 'description', 'module', 'name'}
+# The first two keys are only used by the IntelMQ Manager and can be ignored, the last just contains the runtime parameters and is handled separately
+IGNORED_SYSTEM_PARAMETERS = {'groupname', 'bot_id', 'parameters'}
 
 
 class Bot(object):
@@ -53,6 +56,8 @@ class Bot(object):
     # Bot is capable of SIGHUP delaying
     _sighup_delay: bool = True
     # From the runtime configuration
+    enabled: bool = True
+    run_mode: str = "continuous"
     description: Optional[str] = None
     group: Optional[str] = None
     module = None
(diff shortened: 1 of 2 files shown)


intelmq/lib/bot.py
```diff
@@ -752,16 +757,21 @@ def __load_runtime_configuration(self):
...
-            self.description = params.get('description')
-            self.group = params.get('group')
-            self.module = params.get('module')
+            for option, value in params.items():
+                if option in ALLOWED_SYSTEM_PARAMETERS and value:
+                    self.__log_configuration_parameter("system", option, value)
+                    setattr(self, option, value)
+                elif option not in IGNORED_SYSTEM_PARAMETERS:
+                    self.logger.warning('Ignoring disallowed system parameter %r.',
+                                        option)
```
schacht-certat, line 766: Whats the difference between an `option` that is not in `ALLOWED_SYSTEM_PARAMETERS` and an option that is in `IGNORED_SYSTEM_PARAMETERS` (besides the latter is logged)? Why not simply drop `IGNORED_SYSTEM_PARAMETERS` and log all parameters that are not in `ALLOWED_SYSTEM_PARAMETERS`?
Nitpicking: I prefer iterating through dicts using `key, value` instead of `option, value`. If you don't want to use `key` I'd prefer the term `parameter` instead of `option` to make it more consistent.

intelmq/lib/bot.py
```diff
@@ -42,6 +42,9 @@
 from intelmq.lib.utils import RewindableFileHandle, base64_decode
 
 __all__ = ['Bot', 'CollectorBot', 'ParserBot', 'SQLBot', 'OutputBot']
+ALLOWED_SYSTEM_PARAMETERS = {'enabled', 'run_mode', 'group', 'description', 'module', 'name'}
```
schacht-certat, line 45: why dict and not list?

schacht-certat (changes requested):

intelmq/lib/bot.py
schacht-certat: ah, its a set...

intelmq/lib/bot.py
wagner-certat: Ad naming: I just used the same naming as in the iteration below

wagner-certat: `IGNORED_SYSTEM_PARAMETERS` are silently ignored, as they are legit, but irrelevant here.
Everything which is not in `ALLOWED_SYSTEM_PARAMETERS` or `IGNORED_SYSTEM_PARAMETERS` is not allowed and likely a user configuration error (e.g. things that should have gone into `parameters`)

wagner-certat: Fixed the naming

schacht-certat (approved):

PR #2059, opened 2021-08-20
IDF: Fix and enforce field name restrictions

wagner-certat:
lib/message: fix and pre-compile data format key regex
the regular expression, which is used to check the validity of
harmonization key names, was never effective as the dot wasn't escaped.
Additionally, the characters of the first element were not allowed to
repeat.

add github action for regexploit
Pre-compile the regular expression on module loading instead of every instantiation.
check also extra field names
update two affected bots
plus documentation


.github/workflows/regexploit.yml
```diff
@@ -0,0 +1,32 @@
...
+    branches: [develop, maintenance, master]
+    paths-ignore:
+      - '.github/**'
+  pull_request:
+    branches: [develop, maintenance]
+    paths-ignore:
+      - '.github/**'
+
+jobs:
+  codespell:
```
schacht-certat, line 20: s/codespell/regexploit

schacht-certat (approved):

PR #2076, opened 2021-08-27
parser openphish : source.url => destination.url

certxlm:
Hello,  

The URLs on the OpenPhish feed are phishing URLs and should probably be considered as destination URLs.

Regards,


certxlm:
Indeed, I was not aware of this definition matrix. So I will close the  pull request.  
Thank you for the clarification.

PR #2079, opened 2021-08-30
ENH: make json output indented to make it better readable by humans

schacht-certat:


wagner-certat:
I did a quick test in the manager if nothing breaks and it seems to work as expected.

PR #2081, opened 2021-08-30
Adapt ansible workflow to ansible tests

schacht-certat:


wagner-certat:
Also cherry-picked for maintenance, as that failed as well

PR #2089, opened 2021-08-31
ENH: use the upstream ansible tasks playbook

schacht-certat:
This makes the ansible workflow follow the upstream playbook more
closely, which reduces duplicate code.

### DIFF
--- a/.github/workflows/ansible-tests.yml
+++ b/.github/workflows/ansible-tests.yml
@@ -33,4 +33,4 @@ jobs:
         path: ./src/intelmq-vagrant
 
     - name: Run ansible tests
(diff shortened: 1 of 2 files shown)


wagner-certat (approved):

wagner-certat:
Not sure why the tests are not starting and the ansible is missing there?

b1rger:
> Not sure why the tests are not starting and the ansible is missing there?

because most of the workflows have a `paths-ignore` setting set to `.github/**` so they don't run on any change to the workflows (i.e. it doesn't make sense to run the nosetest testsuite when modifying the codespell workflow)

PR #2104, opened 2021-09-09
FIX: statistics_host returns if no host value is given

### LINKED ISSUE #2103: Bot statistics enabled by default cause performance drop, disabling them is not possible
If the value of parameter `statistics_host` is `null`, the following exception is thrown:
```
    self.instance = bot(bot_id, disable_multithreading=True)
  File "/usr/lib/python3/dist-packages/intelmq/lib/bot.py", line 1164, in __init__
    super().__init__(bot_id=bot_id)
  File "/usr/lib/python3/dist-packages/intelmq/lib/bot.py", line 239, in __init__
    ttl=None,
  File "/usr/lib/python3/dist-packages/intelmq/lib/cache.py", line 27, in __init__
    if host.startswith("/"):
AttributeError: 'NoneType' object has no attribute 'startswith'
```

null should be handled gracefully (i.e. no statistics be written)

waldbauer-certat:
Fixes #2103

### DIFF
--- a/intelmq/lib/cache.py
+++ b/intelmq/lib/cache.py
@@ -24,6 +24,9 @@ class Cache():
 
     def __init__(self, host: str, port: int, db: str, ttl: int,
                  password: Optional[str] = None):
+        if host is None:
+            return
+
         if host.startswith("/"):
             kwargs = {"unix_socket_path": host}
 


intelmq/lib/cache.py
```diff
@@ -24,6 +24,9 @@ class Cache():
 
     def __init__(self, host: str, port: int, db: str, ttl: int,
                  password: Optional[str] = None):
+        if host is None:
+            return
```
ghost, line 28: That is only applicable because the only code calling this cache is the bots.py for the statistics. If any other code would use this class, this fix silently ignores misconfiguration

ghost (changes requested):
please add a changelog entry

PR #2106, opened 2021-09-09
FIX: performance issue because of re-loading harmonization file

waldbauer-certat:
The harmonization file had been loaded within each new report, this nearly kills the CPU, because it has to open the file, parse the file & load it in the memory.


ghost (changes requested):
Please add a changelog entry

ghost (approved):

PR #2117, opened 2021-09-30
ENH: Shodan parser: handle invalid hostnames, move _common_keys out of class, adjust apply_mapping

monoidic:
This PR makes some changes to the Shodan parser, namely:
* Instead of keeping track of `extra.ftp.<something>.parameters`, FTP parameters are collected together into `extra.ftp.features` as a list of said features, reducing field count
* `rsync.modules` is collected
* conversion functions can raise `NoValueException` with a string argument to signify that the conversion would not succeed, such as in the case of a single IP address being given in `hostnames`, which would then be passed into `source.reverse_dns` and fail to validate as a FQDN
* `_common_keys` is moved out of the class
* `_dict_dict_to_obj_list` is introduced, for converting a string-to-dict mapping into a list of dicts with the previous key as an attribute of the dict; this can be useful for preventing issues where, when feeding the data into aggregating tools, you'd end up with many more fields than necessary, e.g `vulns.CVE-2010-0001.cvss`, `CVE-2010-0002.cvss` etc.
* `_get_first` to get the first item from a list, with `NoValueException` raised on empty lists
* `_get_first_hostname` to handle the first valid FQDN from a list of hostnames for `hostnames` in the Shodan banner, if there is one, and gives `NoValueException` otherwise
* `ssl.cert.serial` and `ssl.dhparams.generator`, which may return both integers and strings, are converted to strings
* Changes to `apply_mapping`, such as reducing needless loop iterations, removing a big try-except, and adding the `NoValueException` handling described above
* Stops falsy values (`False`, 0) besides `None` from being filtered out


waldbauer-certat (approved):
Looks good for me

ghost:
Thanks!

I add a changelog entry in 58b84237e, heavily based on your PR description above.

PR #2120, opened 2021-10-21
Add JSON parser function to Jinja2 templates

creideiki:
In our use cases, we sometimes have some JSON data that we want to format nicely for sending to another system via e-mail. We currently do that by having entire expert bots just to format the data into a string to go in the `output` field of a message, which seems wasteful.

With this patch, the Jinja2 template environment gains a function `from_json()` (which just calls `json.loads()` in the standard Python environment), meaning the Templated SMTP output bot can take strings containing JSON documents and do the formatting itself.


waldbauer-certat (approved):
Looks good for me

ghost (approved):

PR #2126, opened 2021-11-28
DOC: feeds.yaml: fix HIBP nginx config

sebix:
the HIBP Enterprise feed snippet's nginx example configuration was
missing three additional whitespaces of indentation to be rendered as
code

Before:
image
Now:
image


waldbauer-certat (approved):
LGTM

PR #2137, opened 2022-01-12
Http mixin default http header overwrites bot configuration

Narzhan:
# Description

The default value for http headers was present also in the constructor of the mixin class effectively overwriting the bot http header configuration if any was supplied. This should be aligned with the 3.x design pattern.

Update: I found issue with the user agent not being replaced in the http header since the section was unreachable. I've added this commit to this pr since I think that creating a new one wouldn't be needed.


sebix (commented):
Thanks for the PR, but unfortunately this is a bit a can of worms. While most of the HTTP Mixin works perfectly fine, there's an unclear behaviour regarding the `http_header` variable which is the reason why it's re-set in the constructor (see the test fail https://github.com/certtools/intelmq/runs/4790117560?check_suite_focus=true#step:9:118).
A few months ago (still at cert.at), we tried to debug that but came to no conclusion, only this workaround. But as you now have found, that creates another bug :/
I fear there's a flaw elsewhere causing that.

The ~~`is not None`~~ changes look good at first sight.

Narzhan:
Understood. However there should either be some reliable way to construct a http requests or this should be kept the generic http collector (with the proper handling of all configuration). I think I'd be okay with both of those options.

Anyway these changes fixed the two issues I was facing so I'll leave it up to you whether you want this to be merged or not.

sebix:
> However there should either be some reliable way to construct a http requests or this should be kept the generic http collector (with the proper handling of all configuration).

I completely agree. However, as I am no longer working at CERT.at my work on IntelMQ is also not funded any more. I fear this is left unsolved unless someone steps up with either time or budget to fix this bug.

sebix:
@waldbauer-certat I think you are more experienced with the HTTP Mixin than me (#2151), can you take charge of this PR?

PR #2142, opened 2022-01-25
TST: also test on python 3.10, switch to pytest

sebix:


sebix:
nosetest doesn't support Python 3.10 -> We can use unittest itself for running the test or pytest

sebix:
pytest by default captures all warnings, but we also want to test in our unittest if the capturing of warnings by the logging works. For a simple solution and to keep this PR focused, I chose to disable the warnings capturing in pytest.

@aaronkaplan Can you please set the nosetests as optional (not required) in the GitHub settings? Thanks.

aaronkaplan:
looks good. IMHO we can merge.

PR #2144, opened 2022-01-27
MS CTIP Azure feed invalid hostname filtering

monoidic:
In some cases, the CTIP Azure feed payload JSON section contains an IP address in `Payload.domain`, which gets mapped into `destination.fqdn`, but IP addresses are not valid FQDNs and hence said events are dumped. This PR adds explicit FQDN validation for `Payload.domain`, discarding invalid values.


intelmq/tests/bots/parsers/microsoft/ctip_azure.txt
```diff
@@ -1,4 +1,5 @@
...
...P.Infected","SourcedFrom":"Microsoft.DCU.CTIP.Sinkhole","DateTimeReceivedUtc":132651352622420000,"DateTimeReceivedUtcTxt":"Monday May 10 2021 15:47:42.2420","Malware":"Avalanche","ThreatCode":"B67-SS-Gamarue","ThreatConfidence":"Low","TotalEncounters":2,"TLP":"Green","SourceIp":"224.0.5.8","SourcePort":28285,"DestinationIp":"10.0.0.1","DestinationPort":80,"SourceIpInfo":{"SourceIpAsnNumber":"64496","SourceIpAsnOrgName":"Example AS","SourceIpCountryCode":"AT","SourceIpRegion":"","SourceIpCity":"","SourceIpPostalCode":"","SourceIpLatitude":48.2,"SourceIpLongitude":16.3667,"SourceIpMetroCode":0,"SourceIpAreaCode":0,"SourceIpConnectionType":"Cellular","SourceIpv4Int":3758097672},"HttpInfo":{"HttpHost":"","HttpRequest":"","HttpMethod":"","HttpReferrer":"","HttpUserAgent":"","HttpVersion":""},"CustomInfo":{"CustomField1":"andromeda210","CustomField2":"","CustomField3":"","CustomField4":"","CustomField5":""},"Payload":"eyJ0cyI6MTYyMDY2MTY2Mi4yNDIzMTYsImlwIjoiMjI0LjAuNS44IiwicG9ydCI6MjgyODUsInNlcnZlcklwIjoiMTAuMC4wLjEiLCJzZXJ2ZXJQb3J0Ijo4MCwiZG9tYWluIjoiZXhhbXBsZS5jb20iLCJmYW1pbHkiOiJhbmRyb21lZGEiLCJtYWx3YXJlIjp7fSwicmVzcG9uc2UiOiJIdHRwT2siLCJoYW5kbGVyIjoiaGFuZGxlcjEiLCJ0eXBlIjoiSHR0cCJ9"}
```
sebix: Is that just a GitHub bug showing a line with no changes as changed line or are my eyes and diffing tools fooling me?

intelmq/bots/parsers/microsoft/parser_ctip.py
```diff
@@ -291,6 +291,9 @@ def parse_azure(self, line, report):
                 if payload_protocol:
                     # needs to overwrite a field previously parsed and written
                     event.add('protocol.application', payload_protocol, overwrite=True)  # "HTTP/1.1", save additionally
+            elif key == 'Payload.domain':
+                if not FQDN.is_valid(value):
+                    continue
```
sebix: Can we be more strict here to prevent accidental loss of data? Just in case that the data some day does not only put the IP address there, but actually invalid domains or so. Or different IP addresses, which you might not want to miss.
```suggestion
            elif key == 'Payload.domain':
                # Sometimes the destination address is also given as domain, ignore it here as we already save it as destination.ip (see https://github.com/certtools/intelmq/pull/2144)
                if not FQDN.is_valid(value) and value == line.get('Payload.serverIp'):
                    continue
```

sebix (changes requested):

intelmq/tests/bots/parsers/microsoft/ctip_azure.txt
monoidic: The file didn't have a newline at the end before.

sebix: Ah. I thought GitHub would display that :/

sebix (approved):

waldbauer-certat (approved):

PR #2149, opened 2022-02-03
FIX: Use redis BLMOVE instead of BRPOPLPUSH (deprecated)

### LINKED ISSUE #1827: redis pipeline: brpoplpush is deprecated
[BRPOPLPUSH](https://redis.io/commands/brpoplpush) is deprecated in favour of [BLMOVE](https://redis.io/commands/blmove).

The command is used here:
https://github.com/certtools/intelmq/blob/979af96050175ec52f15f31a5041aab8112c7df3/intelmq/lib/pipeline.py#L271-L272
and mocked here:
https://github.com/certtools/intelmq/blob/188609bccc11338b995a9e5fde4418059febfeea/intelmq/lib/bot_debugger.py#L120

waldbauer-certat:
As BRPOPLPUSH has been marked as deprecated in favor of BLMOVE.

Fixes #1827


sebix (changes requested):
BLMOVE is 
> Available since: 6.2.0

Source: https://redis.io/commands/blmove/

For example, Debian Bullseye (11) has version 6.0. Just switching to BLMOVE will cause massive turbulences for the users, requiring them to switch to rolling release distros, PPA et. al. or compiling redis themselves.

I propose to support redis < 6.2.0 too.

sebix (approved):

PR #2151, opened 2022-02-04
FIX/ENH: HttpMixin refactored and various fixes

### LINKED ISSUE #2053: Move from `set_request_parameters` to HttpMixin
The following files still use th `set_request_parameters` method from `intelmq.lib.bot`:
```
intelmq/bots/collectors/shodan/collector_stream.py
intelmq/bots/experts/do_portal/expert.py
intelmq/bots/experts/rdap/expert.py
intelmq/bots/experts/ripe/expert.py
intelmq/bots/experts/tuency/expert.py
intelmq/bots/experts/national_cert_contact_certat/expert.py
intelmq/bots/experts/http/expert_status.py
intelmq/bots/experts/http/expert_content.py
intelmq/bots/experts/splunk_saved_search/expert.py
intelmq/bots/outputs/restapi/output.py
intelmq/bots/outputs/elasticsearch/output.py
```
If we update those to use the `intelmq.lib.mixins.HttpMixin` we could finally remove the `set_request_parameters` method from the bot class.

### LINKED ISSUE #2150: Custom headers ignored in HTTPCollectorBot
# The problem
When using the `HTTPCollectorBot`, it is possible to provide custom headers in the bot's configuration using the `http_header` variable. However, when running the bot, these custom headers are ignored, i.e. not passed on in the request.

# Probable cause
The `HTTPCollectorBot` uses the `HttpMixin` class. This class contains a [line](https://github.com/certtools/intelmq/blob/8d1c926ad0f886a937fda5742fa9969501f082cd/intelmq/lib/mixins/http.py#L53) that sets the instance's variable `self.http_header` to an empty dictionary. This overwrites possible custom headers set in the bot's configuration in the manager.

# Possible solutions
1. Simply commenting out [this line](https://github.com/certtools/intelmq/blob/8d1c926ad0f886a937fda5742fa9969501f082cd/intelmq/lib/mixins/http.py#L53) seems to fix the problem.
2. In `bot.py` there is a function [`set_request_parameters`](https://github.com/certtools/intelmq/blob/8d1c926ad0f886a937fda5742fa9969501f082cd/intelmq/lib/bot.py#L855) which looks like it is supposed to set the custom headers (among other things). It might be a better solution to move this function to the `HttpMixin` and calling it at the start of the `setup` function in the `HttpMixin`.

waldbauer-certat:
**General**
Removed 'requests' MissingDependencyError, because requests is a core lib from intelmq
Removed HTTP variables from Bot class in favor of HttpMixin
Removed trying to import requests in pipeline, its a core lib from intelmq
Added additional configuration variables to HttpMixin ( from Bot class )

**Bots**
GitHub API is now using HttpMixin
MS Azure Collector is now using HttpMixin
DO-Portal Expert is now using HttpMixin
GeoHash using MissingDependencyError instead of ValueError (consistency)
HttpContentExpert is now using HttpMixin
HttpStatusExpert is now using HttpMixin
NationalCERTContactCertATExpert is now using HttpMixin
RDAPExpert is now using HttpMixin
RIPEExpert is now using HttpMixin
SplunkSavedSearchExpert is now using HttpMixin
TuencyExpert is now using HttpMixin
RestAPIOutput is now using HttpMixin

**Bot tests**
GitHub API Collector is now using requests_mock instead of MagicMock (consistency)
RestAPI Output is now using correct headers

Fixes #2150
Fixes #2137

### DIFF
--- a/intelmq/bots/collectors/microsoft/collector_azure.py
+++ b/intelmq/bots/collectors/microsoft/collector_azure.py
@@ -11,7 +11,7 @@ import io
 
 from intelmq.lib.bot import CollectorBot
 from intelmq.lib.exceptions import MissingDependencyError
-from intelmq.lib.mixins import CacheMixin
+from intelmq.lib.mixins import CacheMixin, HttpMixin
 
 try:
     from azure.storage.blob import ContainerClient
@@ -23,7 +23,7 @@ except ImportError:
     create_configuration = None  # noqa
 
 
-class MicrosoftAzureCollectorBot(CollectorBot, CacheMixin):
+class MicrosoftAzureCollectorBot(CollectorBot, CacheMixin, HttpMixin):
     "Fetch data blobs from a Microsoft Azure container"
     connection_string: str = "<insert your connection string here>"
     container_name: str = "<insert the container name>"
--- a/intelmq/lib/mixins/http.py
+++ b/intelmq/lib/mixins/http.py
@@ -7,12 +7,8 @@ Based on `create_request_session` in intelmq.lib.utils and
 `set_request_parameters` in intelmq.lib.bot.Bot
 """
 
-from intelmq.lib.exceptions import MissingDependencyError
-
-try:
-    import requests
-except ImportError:
-    requests = None
+import requests
+from typing import Optional
 
 
 class TimeoutHTTPAdapter(requests.adapters.HTTPAdapter):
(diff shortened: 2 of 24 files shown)


intelmq/bots/collectors/microsoft/collector_azure.py
```diff
@@ -23,7 +23,7 @@
     create_configuration = None  # noqa
 
 
-class MicrosoftAzureCollectorBot(CollectorBot, CacheMixin):
+class MicrosoftAzureCollectorBot(CollectorBot, CacheMixin, HttpMixin):
```
wagner-intevation, line 26: Is it used here? The bot uses the azure library and the parameters `http_proxy` and `https_proxy` are used by direct access.

intelmq/lib/mixins/http.py
```diff
@@ -41,12 +37,12 @@ class HttpMixin:
...
-    http_proxy = None
-    https_proxy = None
+    http_proxy: Optional[str] = None
+    https_proxy: Optional[str] = None
     http_timeout_max_tries: int = 3
     http_timeout_sec: int = 30
-    http_username = None
-    http_password = None
+    http_username: str = None
+    http_password: str = None
```
wagner-intevation, line 45:
```suggestion
    http_username: Optional[str] = None
    http_password: Optional[str] = None
```

wagner-intevation (changes requested):

PR #2153, opened 2022-02-08
FIX: Added vscode to gitignore & automated docker build with ACT

waldbauer-certat:


aaronkaplan (commented):
thanks for your submission. merging.

PR #2155, opened 2022-02-16
FIX: threshold expert: configure timeout using Redis cache mixin

creideiki:
When the Redis cache functionality was moved to a mixin, the code that
set the cache timeout from the "timeout" parameter was removed, but
the error checking and documentation remained. Change documentation and
configuration checking code to refer to the standard redis_cache_ttl
parameter instead.

Also remove some redundant default values which are now set in the
cache mixin.

Fixes: 2a1b12640947 ("FIX: Use CacheMixins instead of lib.cache")


sebix (approved):

sebix:
For reasons I don't understand the tests are now failing :/

PR #2158, opened 2022-03-05
bug: relief jinja2 version restriction

sebix:
jinja2 version 3 did not introduce breaking changes, so we can remove
the version restriction
add the requirement also for the jinja expert
fix error message on the required library in the jinja expert


sebix:
Will be merged as part of #2155 as we can't merge the other PR otherwise

PR #2172, opened 2022-05-04
SQL: add MSSQL support and allow storing only a subset of event fields

creideiki:
This pull request enhances IntelMQ's SQL database support in three ways:
1. Add support for MSSQL servers in the `SQLMixin` class, requiring the package `pymssql`.
2. Add a parameter `fields` to `SQLOutputBot`, containing a list of which event fields to store in the database. If this is not specified, store all fields as before.
3. Add a parameter `reconnect_delay` to `SQLMixin`, making it wait that many seconds before attempting to reconnect after a connection error.
    * We found out the need for this the hard way, when the DB server rebooted and IntelMQ's tight reconnect loop filled the entire disk with error logs before the DB server came back.

### DIFF
--- a/docs/user/bots.rst
+++ b/docs/user/bots.rst
@@ -4006,7 +4006,7 @@ SQL
 * `lookup:` no
 * `public:` yes
 * `cache (redis db):` none
-* `description:` SQL is the bot responsible to send events to a PostgreSQL or SQLite Database, e.g. the IntelMQ :doc:`eventdb`
+* `description:` SQL is the bot responsible to send events to a PostgreSQL, SQLite, or MSSQL Database, e.g. the IntelMQ :doc:`eventdb`
 * `notes`: When activating autocommit, transactions are not used: http://initd.org/psycopg/docs/connection.html#connection.autocommit
 
 **Configuration Parameters**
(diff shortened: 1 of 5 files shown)


intelmq/lib/mixins/sql.py
```diff
@@ -18,16 +20,19 @@ class SQLMixin:
 
     POSTGRESQL = "postgresql"
     SQLITE = "sqlite"
+    MSSQL = "mssql"
     default_engine = "postgresql"
     engine = None
     # overwrite the default value from the OutputBot
     message_jsondict_as_string = True
+    reconnect_delay = 0
```
sebix, line 29:
```suggestion
    reconnect_delay: int = 0
```

docs/user/bots.rst
```diff
@@ -4015,15 +4015,17 @@ The parameters marked with 'PostgreSQL' will be sent to libpq via psycopg2. Chec
...
+* `host`: Database host
 * `jsondict_as_string`: save JSONDict fields as JSON string, boolean. Default: true (like in versions before 1.1)
-* `port`: PostgreSQL port
-* `user`: PostgreSQL user
-* `password`: PostgreSQL password
-* `sslmode`: PostgreSQL sslmode, can be `'disable'`, `'allow'`, `'prefer'` (default), `'require'`, `'verify-ca'` or `'verify-full'`. See postgresql docs: https://www.postgresql.org/docs/current/static/libpq-connect.html#libpq-connect-sslmode
+* `port`: Database port
+* `user`: Database user
+* `password`: Database password
+* `sslmode`: Database sslmode, can be `'disable'`, `'allow'`, `'prefer'` (default), `'require'`, `'verify-ca'` or `'verify-full'`. See postgresql docs: https://www.postgresql.org/docs/current/static/libpq-connect.html#libpq-connect-sslmode
```
sebix, line 4025:
```suggestion
* `sslmode`: PostgreSQL sslmode, can be `'disable'`, `'allow'`, `'prefer'` (default), `'require'`, `'verify-ca'` or `'verify-full'`. See postgresql docs: https://www.postgresql.org/docs/current/static/libpq-connect.html#libpq-connect-sslmode (PostgreSQL only)
```

docs/user/bots.rst
```diff
@@ -4015,15 +4015,17 @@ The parameters marked with 'PostgreSQL' will be sent to libpq via psycopg2. Chec
...
 * `table`: name of the database table into which events are to be inserted
+* `fields`: list of fields to read from the event. If None, read all fields
+* `reconnect_delay`: number of seconds to wait before reconnecting in case of an error
```
sebix, line 4028:
```suggestion
* `reconnect_delay`: number of seconds to wait before reconnecting in case of an error (default: 0)
```

intelmq/bots/outputs/sql/output.py
```diff
@@ -29,16 +35,21 @@ class SQLOutputBot(OutputBot, SQLMixin):
     sslmode = "require"
     table = 'events'
     user = "intelmq"
+    fields = None
```
sebix, line 38:
```suggestion
    fields: Optional[list] = None
```

intelmq/bots/outputs/sql/output.py
```diff
@@ -17,6 +17,12 @@
 from intelmq.lib.mixins import SQLMixin
```
sebix, line 17:
```suggestion
from intelmq.lib.mixins import SQLMixin
from typing import Optional
```

sebix (changes requested):
Can't test it, but looks good overall.

Entry in changelog is missing

PR #2180, opened 2022-05-23
ENH: new parameter autoupdate_cached_database

sebix:
New parameter `autoupdate_cached_database` to disable automatic updates (downloads) of cached databases


waldbauer-certat (approved):
We should maybe consider cronjobs instead of integrating update mechanisms into intelmq bots. but seems fine for me

sebix:
> We should maybe consider cronjobs instead of integrating update mechanisms into intelmq bots. but seems fine for me

Funny that you suggest that.
cron is a (deprecated) tool to schedule program starts. We use it for exactly this purpose (see https://github.com/certtools/intelmq/tree/develop/contrib/cron-jobs etc.)
And instead of having a set of shell scripts downloading the databases @gethvi migrated the update code to the bots which has major advantages, e.g. the update scripts are aware of the settings of the bots (license keys, database locations) and can also check if the bots are configured. The updates code is also located directly at the same place as the code using it and future improvements could include automated database-updates not relying on cron, but on internal timers (I image that a fully dockerized setup will need that due to the lack of third-party components in the minimalistic container)
See #1524 and https://github.com/certtools/intelmq/blob/version-3.0-ideas/docs/architecture-3.0.md#user-content-general-principles and you can find further reasoning in the CEF-2018-AT-IA-0111 reports and at other places.

PR #2186, opened 2022-05-30
DOC: installation: remove EOL Fedora versions

sebix:
Fedora 33 and 34 are both EOL, so remove them from the list.


waldbauer-certat (approved):
LGTM

PR #2187, opened 2022-05-30
BUG: mail collectors: fix IMAP abort handling

wagner-intevation:
the try-except block to handle aborted IMAP connections (timeout after a
long processing of the message in IntelMQ) used the wrong class
imaplib.abort does not exist, but the class is named imaplib.IMAP4.abort


waldbauer-certat (approved):
LGTM

PR #2190, opened 2022-05-31
doc: ms azure collector: more precise requirement name

wagner-intevation:
use the precise name of the dependency of the Microsoft CTIP requirement
https://pypi.org/project/azure-storage-blob/


waldbauer-certat (approved):
LGTM

PR #2143, opened 2022-01-25
Shadowserver 2022-01 update

elsif2:
This commit is to update the Shadowserver bots to support all current report types.

## Test Results

```
sudo -u intelmq INTELMQ_SKIP_INTERNET=1 python3 -m unittest
----------------------------------------------------------------------
Ran 1457 tests in 74.351s

OK (skipped=246)
```

PR #2192, opened 2022-06-07
ParserBot: Fix line recovery and message dumping

wagner-intevation:
- BUG: Fix line recovery and message dumping in ParserBot
  ParserBot.process first iterates over all message in a parsed report and
  catches the errors, saves the erroneous message and at the end dumps all
  of them in one row.
  The recover_line methods were not consistent in their use of the line
  given as parameter and the use of the _current_line class member.
  This lead to parsers always dumping the last message (_current_line
  which was active at the time of the dumping) leading to data loss for
  dumped messages for `recover_line_csv_dict` and
  `recover_line_json_stream`.
- DOC: type annotations and inline documentation
  mainly for ParserBot's recover_line methods and related
- ENH: generic csv parser: use current_line
  Use RewindableFileHandle to use the orignal current line for line recovery
- BUG: ctip parser and shadowserver parser: use current line for raw
  use the unparsed current line as string instead of the parsed line
- ENH: ctip parser: better error message for bogous reports
  reports must not be dropped silently if their format is unknown (can
  also happen because of misconfiguration)
- TST: add and update tests for correct line recovery

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,7 @@ CHANGELOG
 - `intelmq.bots.parsers.shadowserver._config`:
   - Added support for `Accessible AMQP`, `Device Identification Report` (IPv4 and IPv6) (PR#2134 by Mateo Durante).
   - Added file name mapping for `SSL-POODLE-Vulnerable-Servers IPv6` (file name `scan6_ssl_poodle`) (PR#2134 by Mateo Durante).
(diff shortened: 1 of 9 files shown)


sebix:
fyi @elsif2 there's also a small fix in the shadowserver parser

elsif2:
Fix to intelmq/bots/parsers/shadowserver/parser.py added to #2143.

sebix:
> Fix to intelmq/bots/parsers/shadowserver/parser.py added to #2143.

The fix is already in this PR #2192. What's the point of fixing it twice?

waldbauer-certat (approved):

PR #2209, opened 2022-07-05
tst: fix threshold expert redis host

### LINKED ISSUE #2207: Threshold Expert test uses hardcoded Redis host
https://github.com/certtools/intelmq/blob/79cae294a74ab77f09b1f371a19f04ad65abab12/intelmq/tests/bots/experts/threshold/test_expert.py#L38
Instead of using 127.0.0.1, `INTELMQ_PIPELINE_HOST` should be used, as done by other tests, to support running tests with redis on different hosts, e.g. in docker setups

wagner-intevation:
Use environment variable `INTELMQ_PIPELINE_HOST` as redis host, analogous to other tests
fixes certtools/intelmq#2207


wagner-intevation:
The unit test fail will be fixed by #2208

waldbauer-certat (approved):
LGTM

PR #2211, opened 2022-07-12
FIX: Use python recommended f-strings & some pep8 corrections [QOL]

waldbauer-certat:
Some Quality of Life updates. Mostly changed .format( & '%' string manipulation to python recommended f-strings.


intelmq/bots/experts/domain_suffix/expert.py
```diff
@@ -54,12 +54,12 @@ def process(self):
     @staticmethod
     def check(parameters):
         if not os.path.exists(parameters.get('suffix_file', '')):
-            return [["error", "File given as parameter 'suffix_file' does not exist."]]
+            return [['error', 'File given as parameter \'suffix_file\' does not exist.']]
```
sebix: What is the advantage of changing the quotes and then escaping it? Is the escaping easier to read/write than the double quotes?

intelmq/bots/experts/domain_valid/expert.py
```diff
@@ -80,35 +80,35 @@ def update_database(cls, verbose=False):
...
         if not bots:
             if verbose:
-                print(f"Database update skipped. No bots of type {__name__} present in runtime.conf.")
+                print(f'Database update skipped. No bots of type {__name__} present in runtime.conf.')
             sys.exit(0)
 
         try:
             session = create_request_session()
-            url = "https://data.iana.org/TLD/tlds-alpha-by-domain.txt"
+            url = 'https://data.iana.org/TLD/tlds-alpha-by-domain.txt'
```
sebix, line 92: What is the advantage of single quotes over double quotes?

intelmq/bots/experts/domain_suffix/expert.py
waldbauer-certat: There is no advantage, just played around and thought about a new standard in intelmq using single quotes outside and double quotes inside f-strings. But in order to avoid confusion, my second guess was using single quotes everytime and use double quotes inside. But as I was unsure if I will break some tests using regex, so I just escaped them for now.

sebix: https://github.com/kenshoo/python-style-guide#user-content-double-vs-single-quotes
says
> 
    Prefer double quotes when your string has multiple words or when formatting ("hi bob").
    If the string is a single word use single quotes ('bob').

https://peps.python.org/pep-0008/#string-quotes says

> In Python, single-quoted strings and double-quoted strings are the same. This PEP does not make a recommendation for this. Pick a rule and stick to it. When a string contains single or double quote characters, however, use the other one to avoid backslashes in the string. It improves readability.
> 
> For triple-quoted strings, always use double quote characters to be consistent with the docstring convention in [PEP 257](https://peps.python.org/pep-0257).

These are just two examples of multiple (and we are not obliged to stick to them), but I agree with them.

I appreciate your conversion to format-strings, this makes the code much cleaner, more readable and a bit more efficient. But I don't see a reason to change the quotes if we need to escape inner quotes.

waldbauer-certat: @sebix As mentioned in PEP8, i've removed all backslashes now.

waldbauer-certat: IntelMQ is now also PEP257 compliant.

waldbauer-certat:
@sebix weird bug appeared

```
DEBUG    test-bot:bot.py:668 Received message {'feed.url': 'http://www.example.com/', 'time.observation': '2015-08-11T13:03:40+00:00', 'raw': 'IyBpZ25vcmUgdGhpcwoyMDE1LzA2LzA0IDEzOjM3ICswMCxleGFtcGxlLm9yZywxOTIuMC4yLjMscmV2ZXJzZS5leGFtcGxlLm5ldCxleGFtcGxlIGRlc2NyaXB0aW9uLHJlcG9ydEBleGFtcGxlLm9yZywxCgoyMDE1LzA2LzA0IDEzOjM4ICswMCxleGFtcGxlLm9yZywxOWEyLjAuMi4zLHJldmVyc2UuZXhhbXBsZS5uZXQsZXhhbXBsZSBkZXNjcmlwdGlvbixyZXBvcnRAZXhhbXBsZS5vcmcsMQoyMDE1LzA2LzA0IDEzOjM4ICswMCxleGFtcGxlLm9yZywxOWIyLjAuMi4zLHJldmVyc2UuZXhhbXBsZS5uZXQsZXhhbXBsZSBkZ...', 'feed.name': 'Example'}.
INFO     test-bot:test_parser_bot.py:85 Lorem ipsum dolor sit amet.
DEBUG    test-bot:test_parser_bot.py:89 test!
DEBUG    test-bot:bot.py:610 Sending message to path '_default'.
DEBUG    test-bot:test_parser_bot.py:89 test!
ERROR    test-bot:bot.py:1056 Failed to parse line.
  File "/home/runner/work/intelmq/intelmq/intelmq/lib/bot.py", line 1052, in process
    events: list[libmessage.Event] = list(filter(bool, value))
  File "/home/runner/work/intelmq/intelmq/intelmq/tests/lib/test_parser_bot.py", line 93, in parse_line
    event['source.ip'] = line[2]
  File "/home/runner/work/intelmq/intelmq/intelmq/lib/message.py", line 136, in __setitem__
    self.add(key, value)
  File "/home/runner/work/intelmq/intelmq/intelmq/lib/message.py", line 255, in add
    raise exceptions.InvalidValue(key, old_value, object=bytes(json.dumps(self.iterable), 'utf-8'))
intelmq.lib.exceptions.InvalidValue: invalid value '19a2.0.2.3' (<class 'str'>) for key 'source.ip'
DEBUG    test-bot:test_parser_bot.py:89 test!
ERROR    test-bot:bot.py:1056 Failed to parse line.
  File "/home/runner/work/intelmq/intelmq/intelmq/lib/bot.py", line 1052, in process
    events: list[libmessage.Event] = list(filter(bool, value))
  File "/home/runner/work/intelmq/intelmq/intelmq/tests/lib/test_parser_bot.py", line 93, in parse_line
    event['source.ip'] = line[2]
  File "/home/runner/work/intelmq/intelmq/intelmq/lib/message.py", line 136, in __setitem__
    self.add(key, value)
  File "/home/runner/work/intelmq/intelmq/intelmq/lib/message.py", line 255, in add
    raise exceptions.InvalidValue(key, old_value, object=bytes(json.dumps(self.iterable), 'utf-8'))
intelmq.lib.exceptions.InvalidValue: invalid value '19b2.0.2.3' (<class 'str'>) for key 'source.ip'
INFO     test-bot:test_parser_bot.py:85 Lorem ipsum dolor sit amet.
ERROR    test-bot:bot.py:359 Bot has found a problem.
  File "/home/runner/work/intelmq/intelmq/intelmq/lib/bot.py", line 316, in start
    self.process()
  File "/home/runner/work/intelmq/intelmq/intelmq/lib/bot.py", line 1066, in process
    self._dump_message(exc, report_dump)
  File "/home/runner/work/intelmq/intelmq/intelmq/tests/lib/test_parser_bot.py", line 137, in dump_message
    self.assertDictEqual(EXPECTED_DUMP[self.call_counter], message)
KeyError: 0
```

sebix:
> @sebix weird bug appeared

I tried to figure it out, but wasn't easy to find. The exceptions and dumping look legitimate and intentional, so I suspect some changed behaviour in the utilities which test and handle this behaviour.

Maybe try bisect-like methods to find out which of your changes triggers this?

kamil-certat (changes requested):
The direction looks good, but:
1. Please rebase and verify which changes are already implemented.
2. Please do not use f-strings or any pre-created strings in log entries. The power of logging module is passing a template and variables - and the message will be rendered only when log should be generated (is in the configured logging level). It would be a better change than modifying to f-strings.
3. Please decide on consistency usage of `'` or `"`. There are changes in both directions, and some of them looks valid (e.g. when using string with apostrophe in text), but other reasons are unclear.

aaronkaplan:
So, what about this PR? It seems to be on outdated code. I agree with @kamil-certat 's opinion on no f-strings for loggers, but how to best proceed with this PR? 
Re-do it on the current code base?
Move to 3.2 ?
What's your opinion?

kamil-certat:
I'd suggest that it may need a recreating from scratch - the codebase has changed, and I'm not sure how much of the changes are still valid or missed. I'd suggest not to push it to the current release.

aaronkaplan:
okay, got it. Moving to 3.2

sebix:
@waldbauer-certat is this PR still active?

waldbauer-certat:
> @waldbauer-certat is this PR still active?

No, not working on it anymore, can be discarded.

sebix:
OK, Thanks for the feedback

PR #2215, opened 2022-07-14
FIX: Added check if cymru hostname is an ip address

### LINKED ISSUE #2169: Cymru parser_cap_program invalid hostname filtering
In some Team-Cymru payloads the `optional_information` entry includes `hostname: ` parameter with an IP address. This is parsed in `intelmq.bots.parsers.cymru.parser_cap_program` bot into `source.fqdn` field, which does not allow IP entries.  

There should be check to filter out IP hostnames, since we already have that data included in `source.ip` field.

The fix can be equivalent to PR #2144. Including `FQDN.is_valid` check seems overkill in this case, but I included it in the example below.

```python
elif key == 'hostname':
    # some hostnames are IP and not valid fqdn
    if not FQDN.is_valid(value) and value == ip:
        continue
    event['source.fqdn'] = value
```

waldbauer-certat:
Fixes #2169


sebix (approved):

waldbauer-certat:
Fixed the merge conflict in CHANGELOG.md - Github is sooo fancy :partying_face:

PR #2223, opened 2022-07-25
postgresql output: escape all null bytes

### LINKED ISSUE #2203: PostgreSQL Output: Issue with Nullbytes in JSON-Data
Nullbytes in JSON-Data (in the field `extra`) can cause troubles:
```
psycopg2.errors.UntranslatableCharacter: unsupported Unicode escape sequence
DETAIL:  \\u0000 cannot be converted to text.
CONTEXT:  JSON data, line 1: ....0\"}, \"malware\": \"Conficker\", \"payload\": {\"text\":...
PL/pgSQL function directives_from_extra(bigint,json) line 5 during statement block local variable initialization
SQL statement \"SELECT directives_from_extra(NEW.id, NEW.extra)\"
PL/pgSQL function events_insert_directives_for_row() line 3 at PERFORM"
```
A problematic extra-value can look like:
```
'{"payload": {"text": "M41\u0012)3U>\bxӾ6\u0000\u0013M6M6M4M4]4y]4ӭ4"}}'
```
Source: Microsoft CTIP

While the `json` type in postgres allows (encoded) Nullbytes in principle, some systems have a problem with that. I have not found out yet, what causes some Postgres-Systems to throw an error here, while others work.
Tested with PostgreSQL 14.4. Works fine on e.g. openSUSE Tumbleweed, but not on Ubuntu 20.04 with Postgres-APT-Repo. But the latter is a target plattform of IntelMQ, so IntelMQ should be compatible with it.

The only solution I see is to escape the Nullbytes in the SQL Output for Postgres (SQLite does not have an issue with these characters).
Working on it.

wagner-intevation:
while null bytes (`\0`, not SQL "NULL") in TEXT and JSON/JSONB fields are valid, data containing null bytes can cause troubles in some combinations of clients, servers and each settings.
To prevent unhandled errors, and data which can't be inserted into the database, all null bytes are escaped

fixes certtools/intelmq#2203

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,8 @@ CHANGELOG
 - `intelmq.bots.outputs.templated_smtp.output`:
   - Add new function `from_json()` (which just calls `json.loads()` in the standard Python environment), meaning the Templated SMTP output bot can take strings containing JSON documents and do the formatting itself (PR#2120 by Karl-Johan Karlsson).
   - Lift restriction on requirement jinja2 < 3 (PR#2158 by Sebastian Wagner).
+- `intelmq.bots.outputs.sql`:
+  - For PostgreSQL, escape Nullbytes in text to prevent "unsupported Unicode escape sequence" issues (PR#2223 by Sebastian Wagner, fixes #2203).
 
 ### Documentation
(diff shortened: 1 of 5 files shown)


waldbauer-certat (approved):
LGTM

PR #2236, opened 2022-08-15
Multithreading causes bot to crash or not start

DigitalTrustCenter:
We are testing the mulithreading capabilities of IntelMQ in combination with RabbitMQ. We have found two (regression) issues that completely break mulithreaded bots:

1. The new YAML configuration format uses a global YAML object in lib/utils.py to read/write configuration. When mulithreading is enabled for a bot, each thread calls \__init__ on the bot class. Since the init function loads the configuration of the bot, this means that each thread tries to load the configuration at the same time. This causes a race condition on the YAML loader, and usually crashes the threads, and thus the bot fails to start.

2. When starting new threads for a bot, the parameter start=True is passed to the \__init__ method of the Bot class. This causes the bot at the end of the \__init__ method to call the start function, and thus start processing messages.
The problem lies in the fact that the subclasses for Bot (CollectorBot, ParserBot, ExpertBot, OutputBot) do not pass the start parameter to the (super) Bot class. This defaults the start parameter to False, hence the threads never start, and quit without doing anything.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ CHANGELOG
   - Enhance behaviour if an unconfigured bot is started (PR#2054 by Sebastian Wagner).
   - Fix line recovery and message dumping of the `ParserBot` (PR#2192 by Sebastian Wagner).
     - Previously the dumped message was always the last message of a report if the report contained multiple lines leading to data-loss.
+  - Fix crashing at start in multithreaded bots (PR#2236 by DigitalTrustCenter).
 - `intelmq.lib.pipeline`:
   - Changed `BRPOPLPUSH` to `BLMOVE`, because `BRPOPLPUSH` has been marked as deprecated by redis in favor of `BLMOVE` (PR#2149 by Sebastian Waldbauer, fixes #1827)
 - `intelmq.lib.utils`:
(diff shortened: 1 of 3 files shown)


sebix (approved):
Thank you for identifying the issues and providing the fixes right away! The problem you experienced shows that we are obviously lacking tests for multi-threading.

The corrections come just in time to being included in the upcoming 3.1.0 release.

btw, did you find the multi-threading feature useful and suitable?

sebix:
Merging is blocked by #2237

DigitalTrustCenter:
Any update on this?

We just had the same problem with the YAML loader in the API and webinput CSV.
We are using Apache + mod WSGI for both the API and webinput csv.It looks like mod WSGI shares the utils globals (like the YAML loader) among those two applications.
One user was opening the IntelMQ web gui (loading the runtime config via the API), while the other was submitting a CSV (loading the harmonization config). 
Both calls failed, due to the YAML loader being shared, and trying to load two files at the same time.

sebix:
ping @waldbauer-certat for review of https://github.com/certtools/intelmq/pull/2237

sebix:
rebased on master

sebix:
rebased on master

PR #2253, opened 2022-10-31
SMTPBatchOutput bot

e3rd:
Hi, we use this bot in CSIRT.cz since 2015 to inform our constituents about all recent incidents twice a week. Events are aggregated in a redis cache by the abuse e-mail and sent at once.

Since we are reaching the external organisations, we've put together some experience to tune up the format of the CSV so that people of various IT skills were able to open it in their spreadsheet software, we use GPG signing and pack the events into ZIP files as we sometimes hit the ceiling of an SMTP server.

We've recently upgraded the syntax to honour the IntelMQ 3, cleaned up the code, put some tips (like CRON setup) to bots.rst and be glad if the community would accept the bot.

### DIFF
--- a/intelmq/bots/outputs/smtp_batch/output.py
+++ b/intelmq/bots/outputs/smtp_batch/output.py
@@ -1,0 +1,354 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+#
+# Initiate dialog ex like this:
+# ssh -t $USER@proki.csirt.cz 'docker exec -it -u 1000 intelmq intelmq.bots.outputs.smtp_batch.output smtp-batch-output-cz cli
+#   --tester [your-email] --ignore-older-than-days 4'
+#
+from __future__ import unicode_literals
+
+import argparse
+import csv
+import datetime
+import json
+import os
+import sys
+import time
+from typing import Any, Optional
+import zipfile
+from base64 import b64decode
+from collections import namedtuple, OrderedDict
+from io import StringIO
+
+from redis.exceptions import TimeoutError
+from envelope import Envelope
+
+from intelmq.lib.bot import Bot
+from intelmq.lib.cache import Cache
+
+
+Mail = namedtuple('Mail', ["key", "to", "path", "count"])
+
+
+class SMTPBatchOutputBot(Bot):
+    # configurable parameters
+    alternative_mails: Optional[str] = None
+    bcc: Optional[list] = None
+    email_from: str = ""
+    gpg_key: Optional[str] = None
+    gpg_pass: Optional[str] = None
+    mail_template: str = ""
+    ignore_older_than_days: Optional[int] = None
+    limit_results: Optional[int] = None
+    redis_cache_db: int = 15
+    redis_cache_host: str = ""
+    redis_cache_port: int = 0
+    redis_cache_ttl: int = 1728000
+    smtp_server: Any = "localhost"
+    subject: str = "IntelMQ warning (%Y-%m-%d)"
+    attachment_name : str = "intelmq_%Y-%m-%d"
+    testing_to: Optional[str] = None
+
+    # private parameters
+    TMP_DIR = "/tmp/intelmq-smtp-batch-output/"
+    mail_contents: str
+    alternative_mail: dict = {}
+    timeout: list
+    cache: Cache
+    key: str
+    send: bool = False
+    cli: bool = False  # whether we are in the CLI mode
+
+    def process(self):
+        message = self.receive_message()        
+        if "source.abuse_contact" in message:
+            field = message["source.abuse_contact"]
+            for mail in (field if isinstance(field, list) else [field]):
+                self.cache.redis.rpush(f"{self.key}{mail}", message.to_json())
+
+        self.acknowledge_message()
+
+    def set_cache(self):
+        self.cache = Cache(
+            self.redis_cache_host,
+            self.redis_cache_port,
+            self.redis_cache_db,
+            self.redis_cache_ttl
+        )
+
+    def init(self):        
+        self.set_cache()
+        self.key = f"{self._Bot__bot_id}:"
+
+    @classmethod
+    def run(cls, parsed_args=None):
+        if not parsed_args:
+            parsed_args = cls._create_argparser().parse_args()
+
+        if parsed_args.cli:            
+            instance = cls(parsed_args.bot_id)            
+            [setattr(instance, k ,v) for k, v in vars(parsed_args).items()]
+            instance.cli_run()
+        else:
+            super().run(parsed_args=parsed_args)
+
+    @classmethod
+    def _create_argparser(cls):
+        argparser = super()._create_argparser()
+        argparser.add_argument('--cli', help='initiate CLI interface', action="store_true")
+        argparser.add_argument('--tester', dest="testing_to", help='tester\'s e-mail')
+        argparser.add_argument('--ignore-older-than-days',
+                            help='1..n skip all events with time.observation'
+                                    ' older than 1..n day; 0 disabled (allow all)',
+                            type=int)
+        argparser.add_argument("--gpg-key", help="fingerprint of gpg key to be used")
+        argparser.add_argument("--limit-results", type=int, help="Just send first N mails.")
+        argparser.add_argument("--send", help="Sends now, without dialog.", action='store_true')        
+        return argparser
+    
+
+    def cli_run(self):
+        os.makedirs(self.TMP_DIR, exist_ok=True)
+        with open(self.mail_template) as f:
+            self.mail_contents = f.read()
+        if self.alternative_mails:
+            with open(self.alternative_mails, "r") as f:
+                self.alternative_mail = {row[0]: row[1] for row in csv.reader(f, delimiter=",")}
+
+        print("Preparing mail queue...")
+        self.timeout = []
+        mails = [m for m in self.prepare_mails() if m]
+
+        print("")
+        if self.limit_results:
+            print(f"Results limited to {self.limit_results} by flag. ", end="")
+
+        if self.timeout:
+            print("Following address has timed out and will not be sent! :(")
+            print(self.timeout)
+
+        if not len(mails):
+            print(" *** No mails in queue ***")
+            sys.exit(0)
+        else:
+            print("Number of mails in the queue:", len(mails))
+
+        while True:
+            print("GPG active" if self.gpg_key else "No GPG")
+            print("\nWhat you would like to do?\n"
+                  f"* enter to send first mail to tester's address {self.testing_to}.\n"
+                  "* any mail from above to be delivered to tester's address\n"
+                  "* 'debug' to send all the e-mails to tester's address")
+            if self.testing_to:
+                print("* 's' for setting other tester's address")
+            print("* 'all' for sending all the e-mails\n"
+                  "* 'clear' for clearing the queue\n"
+                  "* 'x' to cancel\n"
+                  "? ", end="")
+
+            if self.send:
+                print(" ... Sending now!")
+                i = "all"
+            else:
+                i = input()
+
+            if i in ["x", "q"]:
+                sys.exit(0)
+            elif i == "all":
+                count = 0
+                for mail in mails:
+                    if self.build_mail(mail, send=True):
+                        count += 1
+                        print(f"{mail.to} ", end="", flush=True)
+                        try:
+                            self.cache.redis.delete(mail.key)
+                        except TimeoutError:
+                            time.sleep(1)
+                            try:
+                                self.cache.redis.delete(mail.key)
+                            except TimeoutError:
+                                print(f"\nMail {mail.to} sent but could not be deleted from redis."
+                                      f" When launched again, mail will be send again :(.")
+                        if mail.path:
+                            os.unlink(mail.path)
+                print(f"\n{count}× mail sent.\n")
+                sys.exit(0)
+            elif i == "clear":
+                for mail in mails:
+                    self.cache.redis.delete(mail.key)
+                print("Queue cleared.")
+                sys.exit(0)
+            elif i == "s":
+                self.set_tester()
+            elif i in ["", "y"]:
+                self.send_mails_to_tester([mails[0]])
+            elif i == "debug":
+                self.send_mails_to_tester(mails)
+            else:
+                for mail in mails:
+                    if mail.to == i:
+                        self.send_mails_to_tester([mail])
+                        break
+                else:
+                    print("Unknown option.")
+
+    def set_tester(self, force=True):
+        if not force and self.testing_to:
+            return
+        print("\nWhat e-mail should I use?")
+        self.testing_to = input()
+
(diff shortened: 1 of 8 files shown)


e3rd:
I understand tests showed up I can make the code even better. But I'd rather start the discussion first and finish that when I'll implement some changes that are to be demanded.

intelmq/bots/outputs/smtp_batch/output.py
```diff
@@ -0,0 +1,354 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+#
+# Initiate dialog ex like this:
+# ssh -t $USER@proki.csirt.cz 'docker exec -it -u 1000 intelmq intelmq.bots.outputs.smtp_batch.output smtp-batch-output-cz cli
+#   --tester [your-email] --ignore-older-than-days 4'
+#
+from __future__ import unicode_literals
```
sebix: This is already mandatory in Python >= 3.0

intelmq/bots/outputs/smtp_batch/output.py
```diff
@@ -0,0 +1,354 @@
+#!/usr/bin/env python
```
sebix:
```suggestion
```
This file is not an executable

sebix (changes requested):

intelmq/bots/outputs/smtp_batch/output.py
e3rd: Ok, thanks!

intelmq/bots/outputs/smtp_batch/output.py
e3rd: Right!

e3rd:
I've implemented the minor changes you requested. If you like the overall concept of the bot and you are not to deny it in the future, I'll start working to make the tests better. :)

For the instant, I know I'll need a little help with adding an external dependency.

```
intelmq/tests/bots/outputs/smtp_batch/test_output.py:14: in <module>
    raise MissingDependencyError('envelope', '>=2.0.0')
E   intelmq.lib.exceptions.MissingDependencyError: Could not load dependency 'envelope', please install it with apt/yum/dnf/zypper (possibly named python3-envelope) or pip3. Please note that this bot requires envelope version >=2.0.0!
```

I tried to copy the external dependencies like jinja2 or pymisp but I cannot replicate. Should I put the installation of such requirement somewhere into the .github folder or elsewhere?

intelmq/bots/outputs/smtp_batch/output.py
```diff
@@ -0,0 +1,349 @@
...
+                               help='1..n skip all events with time.observation'
+                               ' older than 1..n day; 0 disabled (allow all)',
+                               type=int)
+        argparser.add_argument("--gpg-key", help="fingerprint of gpg key to be used")
+        argparser.add_argument("--limit-results", type=int, help="Just send first N mails.")
+        argparser.add_argument("--send", help="Sends now, without dialog.", action='store_true')
+        return argparser
+
+    def cli_run(self):
+        os.makedirs(self.TMP_DIR, exist_ok=True)
```
kamil-certat: Is there any reason you manually manage a temporary directory, instead of using [tempfile](https://docs.python.org/3/library/tempfile.html) module?

intelmq/bots/outputs/smtp_batch/output.py
```diff
@@ -0,0 +1,349 @@
...
+    def prepare_mails(self):
+        """ Generates Mail objects """
+        allowed_fieldnames = ['time.source', 'source.ip', 'classification.taxonomy', 'classification.type',
+                              'time.observation', 'source.geolocation.cc', 'source.asn', 'event_description.text',
+                              'malware.name', 'feed.name', 'feed.url', 'raw']
+        fieldnames_translation = {'time.source': 'time_detected', 'source.ip': 'ip', 'classification.taxonomy': 'class',
+                                  'classification.type': 'type', 'time.observation': 'time_delivered',
+                                  'source.geolocation.cc': 'country_code', 'source.asn': 'asn',
+                                  'event_description.text': 'description', 'malware.name': 'malware',
+                                  'feed.name': 'feed_name', 'feed.url': 'feed_url', 'raw': 'raw'}
```
kamil-certat: Would be nice if those fields could be customized by some settings :) But not pushing this

intelmq/bots/outputs/smtp_batch/output.py
```diff
@@ -0,0 +1,349 @@
...
+            dict_writer.writerow(dict(zip(ordered_fieldnames, ordered_fieldnames)))
+            dict_writer.writerows(rows_output)
+
+            email_to = str(mail_record[len(self.key):], encoding="utf-8")
+            count = len(rows_output)
+            if not count:
+                path = None
+            else:
+                filename = f'{time.strftime("%y%m%d")}_{count}_events'
+                path = self.TMP_DIR + filename + '_' + email_to + '.zip'
```
kamil-certat: I don't really like the just putting the email in the name, especially that it comes from the abuse data. I know it _shouldn't_ be anything malicious, but even if not widely used, the email address specification is extremely generous, allowing even a `/` character, what would potentially break your code. [1](https://en.wikipedia.org/wiki/Email_address#Local-part) [2](https://www.rfc-editor.org/rfc/rfc6531#section-3.3) 

I would suggest either using hash of the email, or something else. I would suggest it for the redis key as well, although I don't see any harmful case at the moment.

intelmq/bots/outputs/smtp_batch/output.py
```diff
@@ -0,0 +1,349 @@
...
+
+            # write data to csv
+            output = StringIO()
+            dict_writer = csv.DictWriter(output, fieldnames=ordered_fieldnames)
+            dict_writer.writerow(dict(zip(ordered_fieldnames, ordered_fieldnames)))
+            dict_writer.writerows(rows_output)
+
+            email_to = str(mail_record[len(self.key):], encoding="utf-8")
+            count = len(rows_output)
+            if not count:
```
kamil-certat: A general comment, probably not to be addressed now, but I think it's worth thinking for the future: this part of code looks like something potentially nice to be generated on the fly, rather than keeping the whole content in the memory. If you ever find it useful to save memory in this bot, I would suggest:
 - refactoring preparing row message as a generator, then
 - create a ZipFile and use [ZipFile.open](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.open) to open a file in the context manager,  
 - pass this file-like object to the DictWriter,
 - let writer iterate over generator to save data.

This should at lest remove the need to have the entire data kept in `rows_output` and `output` during the whole process.

intelmq/tests/bots/outputs/smtp_batch/test_output.py
```diff
@@ -0,0 +1,114 @@
...
+from pathlib import Path
+import time
+from types import SimpleNamespace
+import unittest
+
+try:
+    from envelope import Envelope
+except ImportError:
+    from intelmq.lib.exceptions import MissingDependencyError
+    raise MissingDependencyError('envelope', '>=2.0.0')
```
kamil-certat: Please don't rise here – it will rise on every importing the file, even just discovering tests.

The CI runs two sets of tests: one are `basic` for bots that work without any additional requirements, and the second one is `full` when additional requirements should be installed. Your test case should be ignored in the basic scenario, so just assign the `None` here.

intelmq/tests/bots/outputs/smtp_batch/test_output.py
```diff
@@ -0,0 +1,114 @@
...
+
+REDIS_LOGIN = {"redis_cache_host": "127.0.0.1",
+               "redis_cache_port": 6379,
+               "redis_cache_db": "14",
+               "redis_cache_ttl": ""}
+MAIL_TEMPLATE = Path("mail_template.txt")
+FROM_IDENTITY = "from-example@example.com"
+
+
+class TestSMTPBatchOutputBot(test.BotTestCase, unittest.TestCase):
```
kamil-certat, line 39: Let's decorate the whole test class with `skip_exotic` and `skip_redis` marks from `intelmq.lib.tests`. This should make basic test scenario happy :)

intelmq/tests/bots/outputs/smtp_batch/test_output.py
```diff
@@ -0,0 +1,114 @@
...
+EVENT2 = {'__type': 'Event',
+          'source.ip': '127.0.0.2',
+          'source.url': 'http://example2.com/',
+          'source.abuse_contact': IDENTITY2
+          }
+
+REDIS_LOGIN = {"redis_cache_host": "127.0.0.1",
+               "redis_cache_port": 6379,
+               "redis_cache_db": "14",
+               "redis_cache_ttl": ""}
```
kamil-certat: Let's see if you could reuse `use_cache` from the test framework. It handles configuring etc.

intelmq/tests/bots/outputs/smtp_batch/test_output.py
```diff
@@ -0,0 +1,114 @@
...
+
+    @classmethod
+    def set_bot(cls):
+        cls.bot_reference = SMTPBatchOutputBot
+        cls.sysconfig = {"alternative_mails": "",
+                         "attachment_name": "events_%Y-%m-%d",
+                         "bcc": [],
+                         "email_from": FROM_IDENTITY,
+                         "gpg_key": "",
+                         "gpg_pass": "",
```
kamil-certat, line 53: FYI: there are some example keys in the repo, you can use them to test the GPG part.

intelmq/tests/bots/outputs/smtp_batch/test_output.py
```diff
@@ -0,0 +1,114 @@
...
+        self.assertEqual(subject, envelope.subject())
+        self.assertEqual(message, envelope.message())
+        self.assertEqual(from_, envelope.from_())
+        self.assertEqual(to, envelope.to())
+
+    @staticmethod
+    def send_message(envelope):
+        global SENT_MESSAGES
+        SENT_MESSAGES.append(envelope)
+        return True  # let's pretend the message sending succeeded
```
kamil-certat: I like the idea of collecting sent messages, but instead of playing with global vars, I would do it as a normal instance variable and method. You can then pass it to the mock exactly as you did – nothing should be broken, and your list won't be kept between test cases. At the moment, mails sent in one test, would be kept in the list if you introduce next tests.

kamil-certat (commented):
I have some comments, but I'm for working on it (adding some more tests ofc) and including in the IntelMQ :)

intelmq/tests/bots/outputs/smtp_batch/test_output.py
e3rd: Thanks for the explanation, I'll move the raising to the `set_bot` so that it happens before every test.

intelmq/tests/bots/outputs/smtp_batch/test_output.py
e3rd: Ok!

intelmq/tests/bots/outputs/smtp_batch/test_output.py
e3rd: Amazing tip! The only thing, now I use redis_cache_db=4 as defined in `test.py/BOT_CONFIG/redis_cache_db` instead of 14. But I suppose there will be no side effects.

intelmq/bots/outputs/smtp_batch/output.py
e3rd: Ok, I've just added a comment to the code referencing this so that it can be refactored in the future.

intelmq/tests/bots/outputs/smtp_batch/test_output.py
e3rd: An excellent catch! I've repaired with a closure, that's what you had on your mind?

```python3
with unittest.mock.patch('envelope.Envelope.send', new=self.send_message()):
...
    def send_message(self):
        def _(envelope):
            self.sent_messages.append(envelope)
            return True  # let's pretend the message sending succeeded
        return _
```

intelmq/bots/outputs/smtp_batch/output.py
e3rd: Totally agreed! Since I see no security risk, I used bare base64 (ok?).

```python
path = self.TMP_DIR + filename + '_' + b64encode(email_to.encode("utf-8")).decode("utf-8") + '.zip'
```

sebix: Please use [`tempfile`](https://docs.python.org/3/library/tempfile.html) for temporary files and directories.

intelmq/tests/bots/outputs/smtp_batch/test_output.py
e3rd: I've found the key that could be added with `"gpg_key": Path(__file__).parent / "../../../assets/key-private.pgp"`.
But since we replace the `envelope.Envelope.send` method of the 3rd party library, we skip any GPG processing.

intelmq/bots/outputs/smtp_batch/output.py
e3rd: No, I'll remove this

intelmq/bots/outputs/smtp_batch/output.py
e3rd: You're right, much shorter

intelmq/bots/outputs/smtp_batch/output.py
e3rd: Ok, I made them customizable :)

e3rd:
Hi, I've implemented all comments here :) . What next? Will you squash the commits when accepting or should I squash them now? (I did not want to do it yet, don't know if that will destroy the conversation here.)

sebix:
> Will you squash the commits when accepting or should I squash them now? (I did not want to do it yet, don't know if that will destroy the conversation here.)

I can squash them when merging the PR, that's no problem.

intelmq/bots/outputs/smtp_batch/output.py
```diff
@@ -0,0 +1,355 @@
...
+import datetime
+import json
+import os
+import sys
+from tempfile import NamedTemporaryFile
+import time
+from typing import Any, Optional
+import zipfile
+from base64 import b64decode
+from collections import namedtuple, OrderedDict
```
sebix:
```suggestion
from collections import OrderedDict
```
`namedtuple` is not used

sebix (approved):

intelmq/bots/outputs/smtp_batch/output.py
e3rd: Ok!

intelmq/tests/bots/outputs/smtp_batch/test_output.py
kamil-certat: Exactly :)

kamil-certat (approved):

sebix:
Thank you @e3rd for the contribution! Finally IntelMQ has a proper output bot to send notifications to constituents directly :) :partying_face:

PR #2254, opened 2022-11-08
Fix to ensure self._mode does not get reset when reloading

enitinggous:
At startup, the Shadowserver Parser bot's init method sets self._mode to 'detect' since self.feedname is None. Next, when reports are parsed, self.feedname will contain the name of the last feed parsed. When the nightly logrotate run forces a reload of the bot, self._mode gets set to 'fixed' and the bot assumes erroneously that every report hereafter is of the type defined by self.feedname

### DIFF
--- a/intelmq/bots/parsers/shadowserver/parser.py
+++ b/intelmq/bots/parsers/shadowserver/parser.py
@@ -40,7 +40,7 @@ class ShadowserverParserBot(ParserBot):
     overwrite = False
 
     def init(self):
-        if self.feedname is not None:
+        if self.feedname is not None and self._mode is None:
             self._sparser_config = config.get_feed_by_feedname(self.feedname)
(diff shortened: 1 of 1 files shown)


sebix (commented):
IMHO the better solution would be to not set `self.feedname` (which is a configuration parameter set by the user) in the parsing code at all, but rather use an internal variable name for this purpose.

enitinggous:
I agree, but it seems this mixing of class variables and instance variables is unfortunately used all over by developers. If a common convention is established to get rid of this habit, I more than welcome it, but it requires many bots to be updated. What is preferred: a simple fix now, or, a better fix later involving many bot updates?

elsif2:
I agree with @sebix, lets fix the code to use an instance variable instead of repurposing the configuration parameter.

PR #2260, opened 2022-11-10
Used the supported dnspython method for resolving

### LINKED ISSUE #2204: Using deprecated method of dnspython
Changelog of version 2.0.0:
> dns.resolver.resolve() has been added, allowing control of whether search lists are used. dns.resolver.query() is retained for backwards compatibility, but deprecated. The default for search list behavior can be set at in the resolver object with the use_search_by_default parameter. The default is False.

`dns.resolver.query()` is used here:

https://github.com/certtools/intelmq/blob/651d0733f719579e96ec4b063a260023ebeadc50/intelmq/lib/harmonization.py#L705=

https://github.com/certtools/intelmq/blob/0c6b3269ac480e47c62c9c31ff8685ba90d91f55/intelmq/bots/experts/abusix/_lib.py#L29=
https://github.com/certtools/intelmq/blob/3bbf1201ad783664985f00be4ba7a44ce5b05e8d/intelmq/bots/experts/reverse_dns/expert.py#L70=
https://github.com/certtools/intelmq/blob/0c6b3269ac480e47c62c9c31ff8685ba90d91f55/intelmq/bots/experts/cymru_whois/_lib.py#L57=

We also need backwards compatibility with dnspython < 2 (e.g. Ubuntu 20.04), so we could add a DNS-Query functon in utils with a `if version < 2 then ... else ...` switch.

kamil-certat:
Closes #2204

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ CHANGELOG
     - Previously the dumped message was always the last message of a report if the report contained multiple lines leading to data-loss.
 - `intelmq.lib.pipeline`:
   - Changed `BRPOPLPUSH` to `BLMOVE`, because `BRPOPLPUSH` has been marked as deprecated by redis in favor of `BLMOVE` (PR#2149 by Sebastian Waldbauer, fixes #1827)
+- `intelmq.lib.utils`:
+  - Added wrapper `resolve_dns` for querying DNS, with the support for recommended methods from `dnspython` package in versions 1 and 2.
 
 ### Development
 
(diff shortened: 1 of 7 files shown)


sebix (changes requested):
Would you be so kind to also add an entry to the changelog file?

Code-wise it's perfect.

kamil-certat:
Thanks for the comment - sure, changelog updated.

waldbauer-certat (approved):
LGTM

PR #2267, opened 2022-11-15
FIX: For Python 3.6, restrict the pymssql version

### LINKED ISSUE #2262: Python 3.6 full test fails to build due to pymssql wheel lack
Yesterday, pymssql has uploaded a new version that lacks of wheel for Python 3.6. It stopped our tests to build on this platform.

They did not officially announce the drop of support for Python 3.6, so I've asked in the issue https://github.com/pymssql/pymssql/issues/787 if it was the plan. If no, we should wait for new wheel; if yes, then we should decide to pin the version or support building the package in the CI.

kamil-certat:
Starting version 2.2.6 pymssql is not providing the wheels for Python 3.6 anymore. Building the package is not trivial, thus restricting version is currently the best option.

Closes: #2262


kamil-certat:
The upstream restored builds for Python 3.6.

Issue #1499, opened 2020-01-31
v2.1.2 default installation after instruction lacks dependencies for parsers/html_table


bernhardreiter:
Following v2.1.2 Install and Development instructions, then starting the example default bot net,
the feodo-tracker-browse-parser stopped with

```sh
  File "/opt/dev_intelmq/intelmq/bots/parsers/html_table/parser.py", line 37, in init
    raise MissingDependencyError("beautifulsoup4")
intelmq.lib.exceptions.MissingDependencyError: Could not load dependency 'beautifulsoup4', please install it with apt/yum/dnf/zypper (possibly named python3-beautifulsoup4) or pip3.
```
Expectation: the instructions or the setup process should make sure that the dependencies for the standard bots are there.

Workaround on Ubuntu 18.04 LTS:
```sh
apt-install python-bs4 python3-lxml
```

Idea: the intelmqctl check command should also check if dependencies can be loaded. Probably need changes in the bots implementation.

wagner-certat:
Sorry for the accidental close, I mixed up the issue numbers -.-

wagner-certat:
See also #552 

Requires some changes in the core. The requirements of bots need to be defined somewhere

motok:
intelmq 3.0.1 seems still have this issue.
```
$ intelmqctl -v
3.0.1
```
```
2022-01-18 16:30:35,475 - feodo-tracker-browse-parser - INFO - HTMLTableParserBot initialized with id feodo-tracker-browse-parser and intelmq 3.0.1 and python 3.8.12 (default, Nov  4 2021, 01:12:24) as process 769.
2022-01-18 16:30:35,475 - feodo-tracker-browse-parser - INFO - Bot is starting.
2022-01-18 16:30:35,650 - feodo-tracker-browse-parser - ERROR - Bot initialization failed.
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/intelmq/lib/bot.py", line 209, in __init__
    self.init()
  File "/usr/local/lib/python3.8/site-packages/intelmq/bots/parsers/html_table/parser.py", line 55, in init
    raise MissingDependencyError("beautifulsoup4")
intelmq.lib.exceptions.MissingDependencyError: Could not load dependency 'beautifulsoup4', please install it with apt/yum/dnf/zypper (possibly named python3-beautifulsoup4) or pip3.
2022-01-18 16:30:35,653 - feodo-tracker-browse-parser - INFO - Bot stopped.
```

[...]

PR #2268, opened 2022-11-15
ENH: Improved Abuse.ch Feodo Tracker parser bot and documentation

gethvi:
This PR replaces two duplicate feeds with an unified and improved one with greatly simplified code. The JSON source from feodotracker.abuse.ch has the most complete data (compared to CSV and Browse HTML table).


gethvi:
Tests are failing because of #2278. Merging the PR #2279 should fix the failing tests.

sebix:
> Tests are failing because of #2278. Merging the PR #2279 should fix the failing tests.

PR #2279 is now merged

sebix (changes requested):
I'm missing the upgrade path. All user which have any of the two parsers (domain or IP) in place, will run into troubles with the upgrade.
Please add the removal of the two bots to the NEWS file and add an upgrade function to `intelmq/lib/upgrades.py`.

gethvi:
This also fixes #1499.

intelmq/lib/upgrades.py
```diff
@@ -824,6 +835,21 @@ def v310_feed_changes(configuration, harmonization, dry_run, **kwargs):
...
+
+    if found_abusech_feodotracker_browse:
+        messages.append('A discontinued feed "Abuse.ch Feodo Tracker Browse" has been found'
+                        'as bot %s.\nPlease manually replace with the feed'
+                        '"Abuse.ch Feodo Tracker".' % ', '.join(sorted(found_abusech_feodotracker_browse)))
+
+    if found_abusech_removed_parsers:
+        messages.append('A discontinued bot module has been found'
+                        'as bot %s.' % ', '.join(sorted(found_abusech_removed_parsers)))
+
```
sebix, line 852: Why not changing the config directly? Replacing the module and adapting the `http_url` can be fully automated.

intelmq/tests/test_conf.py
```diff
@@ -27,6 +27,7 @@
 from intelmq.lib.utils import lazy_int
 
 yaml = YAML(typ="safe", pure=True)
+yaml.default_flow_style = False
```
sebix: Can you please add an inline comment why this is necessary/useful?

sebix (changes requested):
Thank you for rewriting this parser and simplifying the setup by doing so.

I suggest to migrate the configuration directly instead of asking the user to do so. The user doesn't have a choice anyway.

intelmq/lib/upgrades.py
gethvi: I made the explanation into an issue #2297.

intelmq/tests/test_conf.py
gethvi: Fixed.

intelmq/lib/upgrades.py
gethvi: If you insist on changing the config directly I will do it. Just let me know.

However as a user I prefer only being notified that I need to change to the config. I like to understand the necessary changes even if it requires more (manual) effort. That's why I implemented the upgrades this way.

kamil-certat: I'd believe that the change shouldn't be done automatically - but it could be suggested. Either as "run command xyz to automatically update documentation" or "please change following things ... <the detailed change>".

sebix: > Either as "run command xyz to automatically update documentation"

Originally that was the purpose of the `intelmqctl upgrade-config` command.

PR #2271, opened 2022-11-17
FIX: intelmqdump: respected logging_path settings

### LINKED ISSUE #1605: intelmqdump: logging_path parameter not honoured
The logging path given in the `logging_path` parameter (defaults and runtime) can differ from the `intelmq.DEFAULT_LOGGING_PATH` variable, therefore dumps can be at a different place than expected.

kamil-certat:
'intelmqdup' supports now configuration of logging_path,
both global and custom per-both. It can find and restore
dumps from all possibly configured locations. For readibility
reasons, the behaviour of showing just the filename in the
dumps list is left unchanged. If there is more than one
dump listed with the same filename, selecting them using
filename will throw an error, and the number should be used.

Unit testes for the command are introduced. For the purpose
of the change, just checking an empty dump file is enough.

closes: #1605


waldbauer-certat (approved):
LGTM

PR #2274, opened 2022-11-21
ENH: Removed creating objects to call staticmethod

kamil-certat:
In the harmonitation.py we had multiple creations
of empty objects to call static methods. This commit
cleared it and reduced number of uneccessary object
creations.

The changelog entry was intentionally omitted due
to the small size and impact of the change.


waldbauer-certat (approved):
LGTM

PR #2279, opened 2022-11-22
FIX: Fixes DateTime handling of naive time string. Removes pytz dependency.

### LINKED ISSUE #2278: Harmonization DateTime incorrect handling of naive time strings
Considering this function of `DateTime` class:
https://github.com/certtools/intelmq/blob/122e3a386581688593878014cba457dd1c1e5599/intelmq/lib/harmonization.py#L421-L436

In particular the line:
```python
value = value.astimezone(pytz.utc) 
```

This causes that naive time strings (without timezone information) are being considered as being in a timezone of the local computer. My computer is set to a UTC +01:00 timezone and this is the result:
```python
DateTime.sanitize("2021-01-17 07:44:46")    # this calls DateTime.__parse

'2021-01-17T06:44:46+00:00'
```

Because the parsed timestamps are usually produced by feeds outside of the local computer **I believe it makes much more sense to assume they are produced in UTC rather than processing them based on whatever is the local timezone of the IntelMQ installation**.

---

On a side note: There are several cases of this kind of code:
```python
if not value.tzinfo and sys.version_info <= (3, 6):
    value = pytz.utc.localize(value)
elif not value.tzinfo:
    value = value.astimezone(pytz.utc)
```

Such expressions are not equal in what they do. The first one (`pytz.utc.localize`) merely attaches a time zone object to a datetime without adjustment of date and time data. The second one (`astimezone(pytz.utc)`) adjusts time and date based on local time ([docs](https://docs.python.org/3/library/datetime.html#datetime.datetime.astimezone)). Example (with the computer in UTC +01:00):

```python
value = datetime.datetime.strptime("2021-01-17 07:44:46", "%Y-%m-%d %H:%M:%S")
value.isoformat()
'2021-01-17T07:44:46'

pytz.utc.localize(value).isoformat()
'2021-01-17T07:44:46+00:00'

value.astimezone(pytz.utc).isoformat()
'2021-01-17T06:44:46+00:00'
```

I think this is also the reason why the following test works only when the machine running the test is in UTC timezone:
https://github.com/certtools/intelmq/blob/122e3a386581688593878014cba457dd1c1e5599/intelmq/tests/lib/test_harmonization.py#L252-L262

**PROPOSED SOLUTION:**
I think the better way of handling of naive time strings is to assume they are in UTC rather than local timezone of IntelMQ installation:
```python3
if not value.tzinfo:
    value = value.replace(tzinfo=datetime.timezone.utc)
else:
    value = value.astimezone(tz=datetime.timezone.utc)
```

The function `value.replace(tzinfo=datetime.timezone.utc)` does effectively the same thing as `pytz.utc.localize(value)`: it attaches a time zone object to a datetime without adjustment of date and time data. And without needing pytz as dependency.

Quick tests showed that my solution seems to work and passes all the tests including the one previously skipped.

gethvi:
Fixes #2278. Globally removes pytz dependency.

It removes a very small functionality that is not used at all: when making DateTime object from timestamp it is now not possible to also set the timezone of the resulting time. This is also contradictory with the documentation on DateTime which says that UTC is the only allowed timezone.

~~Also fixes following deprecation warning:~~
~~`intelmq/intelmq/lib/harmonization.py:718: DeprecationWarning: please use dns.resolver.resolve() instead`~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,11 @@ CHANGELOG
   - Changed `BRPOPLPUSH` to `BLMOVE`, because `BRPOPLPUSH` has been marked as deprecated by redis in favor of `BLMOVE` (PR#2149 by Sebastian Waldbauer, fixes #1827)
 - `intelmq.lib.utils`:
   - Added wrapper `resolve_dns` for querying DNS, with the support for recommended methods from `dnspython` package in versions 1 and 2.
+- `intelmq.lib.harmonization`:
+  - Fixed DateTime handling of naive time strings (previously assumed local timezone, now assumes UTC) (PR#2279 by Filip Pokorný, fixes #2278)
+  - Removes `tzone` argument from `DateTime.from_timestamp` and `DateTime.from_epoch_millis`
+  - `DateTime.from_timstamp` now also allows string argument
+- Removes `pytz` global dependency
 
 ### Development
 - Removed Python 3.6 from CI.
--- a/intelmq/lib/harmonization.py
+++ b/intelmq/lib/harmonization.py
@@ -34,16 +34,13 @@ import ipaddress
 import json
 import re
 import socket
-import sys
 import urllib.parse as parse
+from typing import Optional, Union
 
 import dateutil.parser
 import dns.resolver
-import pytz
 
 import intelmq.lib.utils as utils
-
-from typing import Optional, Union
 
 __all__ = ['Base64', 'Boolean', 'ClassificationType', 'DateTime', 'FQDN',
            'Float', 'Accuracy', 'GenericType', 'IPAddress', 'IPNetwork',
@@ -429,7 +426,10 @@ class DateTime(String):
 
         try:
             value = dateutil.parser.parse(value, fuzzy=True)
-            value = value.astimezone(pytz.utc)
+            if not value.tzinfo:
+                value = value.replace(tzinfo=datetime.timezone.utc)
+            else:
+                value = value.astimezone(tz=datetime.timezone.utc)
             value = value.isoformat()
         except (ValueError, OverflowError):
             return None
@@ -451,13 +451,18 @@ class DateTime(String):
             # With microseconds
             dtvalue = datetime.datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f+00:00')
 
-        if return_datetime:
-            return pytz.utc.localize(dtvalue)
+        if not dtvalue.tzinfo:
+            dtvalue = dtvalue.replace(tzinfo=datetime.timezone.utc)
         else:
-            return value
+            dtvalue = dtvalue.astimezone(tz=datetime.timezone.utc)
+
+        if return_datetime:
+            return dtvalue
+
+        return value
 
     @staticmethod
-    def from_epoch_millis(tstamp: str, tzone='UTC') -> datetime.datetime:
+    def from_epoch_millis(tstamp: Union[int, str]) -> str:
         """
         Returns ISO formatted datetime from given epoch timestamp with milliseconds.
         It ignores the milliseconds, converts it into normal timestamp and processes it.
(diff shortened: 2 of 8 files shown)


intelmq/lib/harmonization.py
```diff
@@ -702,7 +702,7 @@ def sanitize(value: str) -> Optional[str]:
     @staticmethod
     def to_ip(value: str) -> Optional[str]:
         try:
-            value = str(dns.resolver.query(value, 'A')[0])
+            value = str(dns.resolver.resolve(value, 'A')[0])
```
kamil-certat: Please note, this is in the https://github.com/certtools/intelmq/pull/2260 handled (see issue #2204)

gethvi: Thanks, I removed it.

CHANGELOG.md
```diff
@@ -37,6 +37,10 @@ CHANGELOG
   - Changed `BRPOPLPUSH` to `BLMOVE`, because `BRPOPLPUSH` has been marked as deprecated by redis in favor of `BLMOVE` (PR#2149 by Sebastian Waldbauer, fixes #1827)
 - `intelmq.lib.utils`:
   - Added wrapper `resolve_dns` for querying DNS, with the support for recommended methods from `dnspython` package in versions 1 and 2.
+- `intelmq.lib.harmonization`:
+  - Fixed DateTime handling of naive time strings (PR#2279 by Filip Pokorný, fixes #2278)
```
kamil-certat: I'd suggest explicitly mentioning that previously the local time zone was assumed, and it isn't anymore. It may be important if someone has servers in non-UTC timezone.

setup.py
```diff
@@ -16,7 +16,6 @@
     'psutil>=1.2.1',
     'python-dateutil>=2.5',
     'python-termstyle>=0.1.10',
-    'pytz>=2012c',
```
kamil-certat, line 19: Please remove pytz also from `docs/user/installation.rst`. Probably also from `/docs/requirements.txt`, but this should be tested - I think this is there just for generating some docs from IntelMQ code or so.

intelmq/lib/harmonization.py
```diff
@@ -529,24 +532,21 @@ def convert_from_format_midnight(value: str, format: str) -> str:
         date = datetime.datetime.strptime(value, format)
         if sys.version_info <= (3, 6):
             value = datetime.datetime.combine(date, DateTime.midnight)
-            value = pytz.utc.localize(value)
+            value = value.replace(tzinfo=datetime.timezone.utc)
```
kamil-certat: This case is Python 3.6-specific, maybe just get rid of it?

intelmq/tests/lib/test_harmonization.py
```diff
@@ -210,25 +209,13 @@ def test_datetime_from_epoch_millis(self):
...
                          harmonization.DateTime.from_timestamp(1441008970))
-        self.assertEqual('2015-08-31T07:16:10-01:00',
-                         harmonization.DateTime.from_timestamp(1441008970,
-                                                               'Etc/GMT+1'))
-        self.assertEqual('2015-08-31T04:16:10-04:00',
-                         harmonization.DateTime.from_timestamp(1441008970,
-                                                               'America/'
-                                                               'Guyana'))
+        self.assertEqual('2015-08-31T08:16:10+00:00',
+                         harmonization.DateTime.from_timestamp("1441008970"))
```
kamil-certat, line 218: I'd add to the changelog, that this case (string timestamp) is now supported)

kamil-certat (approved):
LGTM :+1:

CHANGELOG.md
gethvi: Fixed.

intelmq/lib/harmonization.py
gethvi: Fixed.

intelmq/tests/lib/test_harmonization.py
gethvi: Fixed.

setup.py
gethvi: Fixed.

kamil-certat (approved):

sebix:
Great that the timezone support in Python has finally been improved so that these ugly workarounds and the external library are no longer necessary.
Thanks @gethvi for digging deep in the code to analyse and fix the issue!

Issue #1767, opened 2021-03-01
Option to set default fields for generic parsers


gethvi:
Let's say I have a generic csv parser with following parameters:
```json
"parameters": {
  "columns": [
    "source.url",
    "source.fqdn",
    "source.ip",
    "time.source",
    "__IGNORE__",
    "__IGNORE__"
  ],
  "delimiter": ",",
  "skip_header": true,
  "type": "phishing"
}
```

I would like to be able to set any other fields (including `extra.somekey`) to a fixed value. For instance:
```
protocol.transport = tcp
protocol.application = http
source.port = 80
```
Setting a fixed value to a field that is assigned in the `columns` array could also server as default value in case when a particular column of a row is empty.

This is currently not possible with generic parsers and a custom parser would have to be implemented. Only the `type` parameter can be set to have proper classification. I would like to request a feature to have this option.

**Proposed realization:**
- add `default_fields` as subkey of `parameters`
- rename `type` to `classification.type` and move to `default_fields` for consistency
- allow any other field defined in [Harmonization Fields](https://intelmq.readthedocs.io/en/maintenance/dev/harmonization-fields.html) (with the exception of `raw`, `time.source` and `time.observation` fields) to be set to a fixed value (serving also as a default value) in the `default_fields` of bot configuration
- allow fields starting with `extra.` to be set to a fixed value in the `default_fields` of bot configuration

**Example of proposed realization:**
```json
"parameters": {
  "columns": [
    "source.url",
    "source.fqdn",
    "source.ip",
    "time.source",
    "__IGNORE__",
    "__IGNORE__"
  ],
  "delimiter": ",",
  "skip_header": true,
  "default_fields": {
    "classification.type": "phishing",
    "protocol.transport": "tcp",
    "protocol.application": "http",
    "source.port": 80,
    "extra.different_system_tag": "web-phishing"
  }
}
```

I can work on this feature, but I would like to know first if it would be accepted.

wagner-certat:
Thanks for proposing and accurately describe the idea! I very much like it and am looking forward to your PR =)

> This is currently not possible with generic parsers and a custom parser would have to be implemented.

Or an additional expert (modify, sieve) is used.

> allow any other field defined in Harmonization Fields (with the exception of raw, time.source and time.observation fields)

I wouldn't handle them differently. There could be a use-case for that and implementing the restriction would only restrict the possibilities for the user. (Also, the list of disallowed fields would be kind of arbitrary)

[...]

PR #2293, opened 2023-01-04
ENH: Adds default_fields parameter

gethvi:
New feature based on #1767.

Still work in progress.


intelmq/lib/bot.py
```diff
@@ -952,6 +954,25 @@ def __init__(self, bot_id: str, start: bool = False, sighup_event=None,
...
+                except exceptions.InvalidValue:
+                    self.logger.error("Invalid value of key '%s' in default_fields parameter.", key)
+                    stop = True
+
+                except exceptions.InvalidKey:
+                    self.logger.error("Invalid key '%s' in default_fields parameter.", key)
+                    stop = True
+
+            if stop:
+                self.stop()
```
aaronkaplan, line 974: do we really want to stop a bot if an invalid key is given? What about logging it as warning and ignoring it?
I am just a bit extra careful since this is in the main Bot class.

aaronkaplan (commented):
I think we should clarify the stop() behaviour

intelmq/lib/bot.py
gethvi: I believe so. This check uses the same code that is used for the actual parsing: `event.add_key`. If an invalid key is added during the actual parsing then the bot would crash, stop and restart on every event anyway.

Also when the bots starts a and runs without crashing/stopping (but ignoring invalid key) I wouldn't check the logs assuming everything is fine.

aaronkaplan: okay, since it's configured at start time via config file, at least a stop() will be noticed immediately. So, from the UX side, it's clear that something does not work. Yeah, okay... convinced. Thanks for thinking this through and for the PR! Merging once, you tell me the PR is ready and can be tested.

gethvi:
Ready for review. Can this PR make it into 3.1 please? :-)

aaronkaplan (approved):
unittests work. Thanks, merging

aaronkaplan (approved):
> Ready for review. Can this PR make it into 3.1 please? :-) 

very happy to do so :) Thank you *very* much for your contribution!

aaronkaplan (approved):
> Ready for review. Can this PR make it into 3.1 please? :-) 

very happy to do so :) Thank you *very* much for your contribution!

aaronkaplan:
All unittests passed. Merging.

PR #2295, opened 2023-01-09
FIX: Disabled Feodotracker bots in default config

### LINKED ISSUE #1499: v2.1.2 default installation after instruction lacks dependencies for parsers/html_table

kamil-certat:
Feodotracker bots requires additional, heavy dependencies
(beautifulsoup4) to start. Disabling them in the default
conf allows us to ship the working config with the package
without increase the dependencies required to install
at the beginning.

Fix: #1499

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ CHANGELOG
 - Upgraded syntax to Python 3.6 (mostly Format-Strings) using pyuprade (PR#2136 by Sebastian Wagner).
 
 ### Configuration
+- Disabled the Feodo bots in the default configuration as they required additional dependencies to run (fixes [#1499](https://github.com/certtools/intelmq/issues/1499)).
 
 ### Core
 - `intelmq.lib.upgrades`:
--- a/intelmq/etc/runtime.yaml
+++ b/intelmq/etc/runtime.yaml
@@ -40,7 +40,7 @@ deduplicator-expert:
 feodo-tracker-browse-collector:
   description: Generic URL Fetcher is the bot responsible to get the report from an
     URL.
-  enabled: true
+  enabled: false
   group: Collector
   module: intelmq.bots.collectors.http.collector_http
   name: URL Fetcher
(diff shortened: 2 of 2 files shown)


gethvi:
Please check #2268 which fixes this as a side product.

kamil-certat:
Right, your solution looks much better as the final state. I have a doubt if we want to add bot removal to 3.1 release (then we could include this one in the 3.1 to ship working conf, and yours in the next one). But it's not my decision, I'm uncertain if we keep the principles of *minor* releases or not.

sebix:
> Right, your solution looks much better as the final state. I have a doubt if we want to add bot removal to 3.1 release (then we could include this one in the 3.1 to ship working conf, and yours in the next one). But it's not my decision, I'm uncertain if we keep the principles of _minor_ releases or not.

PR #2268 doesn't break (compatibility) so it is in line with semantic versioning's definition of minor releases IMO.

aaronkaplan:
So I take it from the discussion that #2295 is superseded by #2268 and we should not merge in #2295 then?

Please advise...

sebix:
I think so.

Correct @kamil-certat ?

kamil-certat:
Exactly, it doesn't make sense any more.

PR #2296, opened 2023-01-16
FIX: custom HTTP Header

marjatech:
custom HTTP Header is overwritten by __init__ call here. 
Prevents usage of custom Authentication headers. 
since this is initialized in class definition already it is not needed.


sebix:
Can you please explain in which circumstances this line causes trouble? That line has actually been added to solve an issue.

marjatech:
Sure, 

I have to use the HTTP-Collector using a custom Authorization-Header by using the bot-configuration parameter `http_header: {"Authorization":"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"}`
Explicitly I can not use username and password for authentication. 

But effectively this bot parameter is not taking any effect. Since it is overwritten in the init function call with an empty dict, the final request is done without the explicitly defined custom http-header, so I can not authenticate.

and i can't find a related issue, looks like this line has been there since 2d291732934d027b3ffb6b6b897ae4ee76fe10a9

sebix:
IIRC these test fails are related to exactly that line of code.

marjatech:
Don't quite get the exact source of the `True not being a string`-error, but sure we can leave that line for the case of this variable being `None` there. 
Tests complete successfully this way while it can still be used to set custom headers via bot-configuration.

intelmq/lib/mixins/http.py
```diff
@@ -50,7 +50,7 @@ class HttpMixin:
 
     def __init__(self, **kwargs):
         self.logger.debug("Running HTTP Mixin initialization.")
-        self.http_header = {}
+        self.http_header = self.http_header or {}
```
aaronkaplan, line 53: I tested and it makes no difference when the http_header was not set yet.  thank you for your contribution! Merging.

aaronkaplan (approved):
thanks for your contribution!

PR #2298, opened 2023-01-24
FIX Very slow dataplane parser due to loading the config for every event

DigitalTrustCenter:
The Dataplane parser creates a new event for every data line. The harmonization config is not passed to this constructor.
This causes the code to load the harmonization config for every data line.

For feeds like http://dataplane.org/sshclient.txt, this will load the harmonization config almost 40,000 times. Hence it can take hours to parse all the feeds.
This fix reuses the harmonization config that is already loaded in the bot. This reduces the parse time to minutes (if not seconds).


wagner-intevation (approved):

sebix:
Thank you for finding and fixing this issue!

PR #2301, opened 2023-01-26
TST: bin/ctl.check: Skip executables check in tests

### LINKED ISSUE #2300: ctl tests require an installed IntelMQ instance
@kamil-certat `test_check_pass_with_default_runtime` and `test_check_passed_with_default_harmonization_and_empty_runtime` assume a local installation of IntelMQ, including a state file and the bot executables in $PATH. Is that intentional?

wagner-intevation:
in test environments, the executables don't exist necessarily add a parameter to check() to allow skipping these tests

fixes certtools/intelmq#2300

### DIFF
--- a/intelmq/bin/intelmqctl.py
+++ b/intelmq/bin/intelmqctl.py
@@ -822,7 +822,7 @@ Get some debugging output on the settings and the environment (to be extended):
         self.log_log_messages(messages[::-1])
         return 0, messages[::-1]
 
-    def check(self, no_connections=False):
+    def check(self, no_connections=False, check_executables=True):
         retval = 0
         if self._returntype is ReturnType.JSON:
             check_logger, list_handler = utils.setup_list_logging(name='check', logging_level=self._logging_level)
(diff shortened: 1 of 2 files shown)


wagner-intevation:
@kamil-certat can you please review this?

Thanks!

intelmq/bin/intelmqctl.py
```diff
@@ -945,13 +945,14 @@ def check(self, no_connections=False):
...
-            for bot_id, bot in group.items():
-                if subprocess.call(['which', bot['module']], stdout=subprocess.DEVNULL,
-                                   stderr=subprocess.DEVNULL):
-                    check_logger.error('Incomplete installation: Executable %r for %r not found in $PATH (%r).',
-                                       bot['module'], bot_id, os.getenv('PATH'))
-                    retval = 1
+        if check_executables:
+            for group in utils.list_all_bots().values():
+                for bot_id, bot in group.items():
+                    if subprocess.call(['which', bot['module']], stdout=subprocess.DEVNULL,
```
aaronkaplan, line 951: assumes that `which` is in the $PATH. Might want to make this explicit. Also think about docker images where the `which` command might be stripped out ? Would there be a more self-reliant mechanism to check if the executable exists? Like we should know, where it was installed, no?

aaronkaplan (commented):
Minor feedback

intelmq/bin/intelmqctl.py
aaronkaplan: Addendum: if you think it's ok enough in all setups where intelmqctl could be called , to use the `which` command, then I am fine with this line. I can't assess it myself.

aaronkaplan: ignore my comment... sorry.

wagner-intevation: This PR only adds the enclosing `if check_executables:` condition, not the subprocess call, which is in place since 2017 (05505037866ee84cb214263a88d6adef61c2d4de)

PR #2309, opened 2023-02-07
DOC: Fix the confusing config in API docs

kamil-certat:
The documentation for installing API from pip was wrongly suggesting renaming Apache config to `manager-apache.conf`. This could suggest that it's the same as Apache config for IntelMQ Manager, but it's not.


kamil-certat:
@aaronkaplan a small change to fix confusion caused by docs

sebix (approved):

PR #2314, opened 2023-02-08
FIX: Removes duplicate upgrade function

gethvi:
Function `v310_shadowserver_feednames` was twice in the code for some reason.


sebix:
Thanks for catching this!

Was introduced by 462d0d23ee

PR #2319, opened 2023-02-13
FIX: Fixes a bug in default_fields upgrade function

gethvi:
Not all CSV parsers have `type` key unfortunately... :cry:

### DIFF
--- a/intelmq/lib/upgrades.py
+++ b/intelmq/lib/upgrades.py
@@ -808,10 +808,11 @@ def v310_feed_changes(configuration, harmonization, dry_run, **kwargs):
                 bot["module"] == "intelmq.bots.parsers.abusech.parser_domain"):
             found_abusech_removed_parsers.append(bot_id)
         if bot["module"] == "intelmq.bots.parsers.generic.parser_csv":
-            bot["parameters"]["default_fields"] = {
-                "classification.type": bot["parameters"]["type"]
-            }
-            del bot["parameters"]["type"]
+            if bot.get("parameters") and bot["parameters"].get("type"):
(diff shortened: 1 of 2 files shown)


sebix:
changing the upgrade function for 3.1.0 has no effect on installations which are already upgraded to 3.1.0

kamil-certat:
Hmm, but it could affect any new installation or people who don't manged to upgrade because of that, right?

sebix:
> Hmm, but it could affect any new installation or people who don't manged to upgrade because of that, right?

Yes. The fix is good for these cases, but not sufficient for all cases.

sebix:
rebased on develop and added a changelog entry

@kamil-certat please have a quick look / merge

intelmq/lib/upgrades.py
```diff
@@ -808,10 +808,11 @@ def v310_feed_changes(configuration, harmonization, dry_run, **kwargs):
                 bot["module"] == "intelmq.bots.parsers.abusech.parser_domain"):
             found_abusech_removed_parsers.append(bot_id)
         if bot["module"] == "intelmq.bots.parsers.generic.parser_csv":
-            bot["parameters"]["default_fields"] = {
-                "classification.type": bot["parameters"]["type"]
-            }
-            del bot["parameters"]["type"]
+            if bot.get("parameters") and bot["parameters"].get("type"):
```
kamil-certat, line 811: Just my style preferences comment: if the only `parameter` value we expect is a dict, it could be written a little shorter, as: `bot.get("parameters", {}).get("type")`

kamil-certat (approved):
Good!

Issue #2326, opened 2023-03-06
Having issues with intelmq.bots.collectors.file.collector_file


x8006:
Hello everybody,

I would like to start by saying how interesting and effective this tool is. 

I am trying to use the File Collector but unable to to do so. I am using manager and added the file collector bot, but even though I have entered correct path, I am getting the following error: 
```
2023-03-06 10:17:19,415 - File-Collector - INFO - FileCollectorBot initialized with id File-Collector and intelmq 3.1.0 and python 3.8.10 (default, Nov 14 2022, 12:59:47) as process 118089.
2023-03-06 10:17:19,415 - File-Collector - INFO - Bot is starting.
2023-03-06 10:17:19,486 - File-Collector - ERROR - Bot initialization failed.
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/intelmq/lib/bot.py", line 214, in __init__
    self.init()
  File "/usr/lib/python3/dist-packages/intelmq/bots/collectors/file/collector_file.py", line 46, in init
    raise exceptions.InvalidArgument('path', got=self.path,
intelmq.lib.exceptions.InvalidArgument: Argument 'path' is invalid. Should be of type: directory. Got '/tmp/test1/'.
2023-03-06 10:17:19,487 - File-Collector - INFO - Bot stopped.

```
However the path is a directory
```
$ file /tmp/test1/
/tmp/test1/: directory

```

Am I missing something ??

Thanks.

x8006:
UPDATE: If I use `/tmp/` directory everything works fine but if I create a new directory and put files into that directory, intelmq can not access it ?!

I have used 777 permission on the folder but still not working.

sebix:
Did you start the Bot with the Manager? Is the result different, when you run `intelmqct start .. ` on the command line? (If your webserver has a confined env (`PrivateTmp`) its subprocess may not access other dirs)

To check what IntelMQ sees, can you please run
```bash
sudo -u intelmq python3 -c "import pathlib; pathlib.Path('/tmp/test1/').stat()"
```

sebix:
Closing because of inactivity. Please reopen if needed.

Issue #2105, opened 2021-09-09
Generic CSV parser: time conversion and time format


MariusUrkis:
When CSV parser parses the file and event line contains time format like %d-%m-%Y, fuzzy time conversion fails to guess correct time format. Example of the feed is http://benkow.cc/export.php
Fuzzy conversion works with such date entries:
>>> DateTime.convert_fuzzy('12-08-2021')
'2021-12-08T00:00:00'
And correct with days >12 :
>>> DateTime.convert_fuzzy('16-08-2021')
'2021-08-16T00:00:00'

Probably better option to use from_format:
>>> DateTime.convert_from_format('12-08-2021', '%d-%m-%Y')
'2021-08-12T00:00:00+00:00'

ghost:
The parameter `time_format` of the CSV Parser could be extended for this purpose

PR #2329, opened 2023-03-08
FIX+ENH: Time conversions for CSV parser and HTML table parser

gethvi:
This PR:
* Adds new `TimeFormat` class for `time_format` bot parameter. It improves performance, as it validates the parameter only once on instantiation of the bot class and not every time datetime is parsed (looking at you HTML Table parser). Also removes some code duplicity.
* Changes CSV Parser time conversions. For some reason the CSV parser had it's own `TIME_CONVERSIONS` and it was very limited (e.g. `from_format` couldn't be used, see #2326). This PR changes it to use `DateTime.TIME_CONVERSIONS`. Now CSV parser uses `TimeFormat` class for `time_format` parameter.
* Changes HTML Table parser to use `TimeFormat` class for `time_format` parameter as well.
* Changes `DateTime` conversion function names to consistent naming scheme starting with `from_`. Changes function signature to be consistent. Backwards compatible.
* Updates some docstrings.
* Fixes a bug in `InvalidArgument` exception.


gethvi:
Ready for review.

intelmq/lib/harmonization.py
```diff
@@ -436,9 +436,9 @@ def __parse(value: str) -> Optional[str]:
         return utils.decode(value)
 
     @staticmethod
-    def parse_utc_isoformat(value: str, return_datetime: bool = False) -> Union[datetime.datetime, str]:
+    def from_isoformat(value: str, return_datetime: bool = False) -> Union[datetime.datetime, str]:
```
sebix, line 440: Do we consider these methods a publicly-facing interface of IntelMQ? If yes, considering semantic versioning guidelines, such changes require a new major version. If not, we can go forward with the rename.

gethvi: Thank you, good point. It is probably safer to consider it public interface. I do not wish to break someone else's code. I added the old functions for compatibility purposes and marked them for deletion with next major release.

intelmq/lib/datatypes.py
```diff
@@ -1,26 +1,31 @@
...
+from intelmq.lib.exceptions import InvalidArgument
+from intelmq.lib.harmonization import DateTime
 
 
 class BotType(str, Enum):
-    COLLECTOR   = "Collector"
-    PARSER      = "Parser"
-    EXPERT      = "Expert"
-    OUTPUT      = "Output"
+    COLLECTOR = "Collector"
```
kamil-certat: NIT: Generally I prefer your version, however I don't like unexpected formatting changes

intelmq/lib/datatypes.py
```diff
@@ -54,8 +58,85 @@ def toJson(self):
...
     CRITICAL = 4
+
+
+class TimeFormat(str):
+    """
+    Pydantic style Field Type class for bot parameter time_format. Used for validation.
+    """
+
+    def __new__(cls, value: Optional[str] = None):
+        value = "fuzzy" if (value is None or value == "") else value
```
kamil-certat: Doesn't it the same as `value = value or "fuzzy"`, assuming the type is correct?

intelmq/lib/datatypes.py
```diff
@@ -54,8 +58,85 @@ def toJson(self):
...
+        return super().__new__(cls, value)
+
+    def __init__(self, value: Optional[str] = None):
+
+        self.convert: Callable
+        self.format_string: Optional[str] = None
+
+        super(TimeFormat, self).__init__()
```
kamil-certat: `super().__init__()` should be enough, the extended version is a Python 2-style

intelmq/lib/datatypes.py
kamil-certat: BTW, do we need bot `__new__` and `__init__` in this case? Why not just do it in the `__init__`? There is no alternate path with a different constructor, also I don't see any reason to use `__new__`.

intelmq/lib/datatypes.py
```diff
@@ -54,8 +58,85 @@ def toJson(self):
...
+
+        if isinstance(value, TimeFormat):
```
kamil-certat, line 88: According to the `__init__` definition, 'value' should never be `TimeFormat`.

intelmq/lib/datatypes.py
```diff
@@ -54,8 +58,85 @@ def toJson(self):
...
+        """
+        if self.format_string:
+            return self.convert(value=value, format=self.format_string, return_datetime=return_datetime)
+        else:
+            return self.convert(value=value, return_datetime=return_datetime)
+
+    @classmethod
+    def __get_validators__(cls):
+        """
+        This function is for Pydantic compatibility.
```
kamil-certat: Do we use Pydentic anywhere? If not, the compatibility is rather unnecessary until we decide to

intelmq/lib/datatypes.py
```diff
@@ -54,8 +58,85 @@ def toJson(self):
...
+        """
+        yield cls
+
+    @staticmethod
+    def validate(value: str) -> [Callable, Optional[str]]:
+        """
+        This function validates the time_format parameter value.
```
kamil-certat, line 110: Because it returns data, I would say that it's more parsing/extraction, and validation is just additional. But I suppose it's the convention from Pydentic?

kamil-certat (commented):
I like the idea :) I have a few small questions about some implementation details that I'm unsure what their goal is

intelmq/lib/datatypes.py
gethvi: Sorry, I use auto lint of Pycharm. He is the culprit.

intelmq/lib/datatypes.py
gethvi: > Doesn't it the same as `value = value or "fuzzy"`, assuming the type is correct?

Yes you are correct, this is just more explicit. But `value = value or "fuzzy"` is more elegant definitely.

> BTW, do we need bot `__new__` and `__init__` in this case? Why not just do it in the `__init__`? There is no alternate path with a different constructor, also I don't see any reason to use `__new__`.

Because `str` is immutable and I want to manipulate it, it must be done before the object is instantiated. Therefore I need to use the `__new__` method. The manipulation is kind of a defensive measure when for example intelmq-manager badly interprets empty parameter. No need to fail/raise exception, just use the default value. This couldn't be done in the `__init__` as far as I know and from what I tried.

However adding the `self.convert` and `self.format_string` should be done in the `__init__` method. This is also where the validation is done.

This is why both `__new__` and `__init__` are used.

intelmq/lib/datatypes.py
gethvi: You are right, my excuse is again Pycharm code auto completion. I just forgot to remove it. :innocent:

intelmq/lib/datatypes.py
gethvi: Well the `TimeFormat` _is_ `str` as well.

This is to allow code such as this (which is generally fairly common pattern):
```python
t1 = TimeFormat("from_format|%Y-%m-%d %H:%M:%S")
t2 = TimeFormat(t1)
```

By recognizing that the value is `TimeFormat` and not just `str` we can save ourselves some computation because the validation doesn't need to be called again.

I understand the annotations only as recommendations for other code on what should be passed as an argument. Because in the end Python lets you pass any data type even if it contradicts the annotations. So person can't fully count on the annotations being strictly followed. Expecting the above mentioned pattern in some way seems reasonable to me.

intelmq/lib/datatypes.py
gethvi: You are correct, it is a convention from Pydantic. It does return a value, but it's primary function is to check the validity of the argument or raise an exception. The return value is kind of a side product of the validation. But it's computed already at this point so why not use it?

intelmq/lib/datatypes.py
gethvi: Okay I will remove it.

intelmq/lib/datatypes.py
sebix: Can you remove this change from the PR?

I like using `git add -p` to choose which changes I want to be in a commit, avoiding committing accidental changes like these.

intelmq/lib/datatypes.py
sebix: I agree `value = value or "fuzzy"` is nicer and quicker to understand.

Can you please add this explanation of `__new__` into an inline comment?

intelmq/lib/datatypes.py
sebix: > `super().__init__()` should be enough, the extended version is a Python 2-style

IIRC it was necessary in earlier Python 3 versions too, but not in any Python version that IntelMQ is developed for

intelmq/lib/datatypes.py
kamil-certat: Thanks for the explanation, I've missed it extends string

intelmq/lib/datatypes.py
kamil-certat: True, I missed inheritance from `str`. Thanks for explaining :+1:

intelmq/lib/datatypes.py
kamil-certat: Yeah, but if we aren't Pydantic, and do not use the `validate` in any other place (true? or I missed it?), maybe just rename it to `_parse` or something? It's not a blocker, but I don't think it makes sense if some isn't familiar with the Pydantic convention. I would suggest, that a parsing method is more accurate to what we us it for. But yeah, it's a philosophical discussion

intelmq/lib/datatypes.py
gethvi: Reverted.

intelmq/lib/datatypes.py
gethvi: Explanation added.

intelmq/lib/datatypes.py
gethvi: Fixed.

kamil-certat (approved):

gethvi:
Added one more fix (#2105) so that `from_fuzzy` has more predictable behavior.

PR #2332, opened 2023-03-09
FIX: Reset destination path statistics on restart

### LINKED ISSUE #2331: Statistics never reset path_total value
IntelMQ built-in statistics feature works in the way, that after every restart of the bot, statistics are considered to be counted from 0 ([code](https://github.com/certtools/intelmq/blob/eba91a6022c6805da33fbbd9c566ebb8c6b24c73/intelmq/lib/bot.py#L126-L133)).

The statistics cache is updated in `__stats` method. Successful and failure stats are sent every time (with the respect to the delay), however the number of sent messages (lines 472-474) is updated only when a message was sent - paths are never filled with 0, the `defaultdict` is used. This means, that from the restart to the first sent message, the Redis statistic cache and IntelMQ message counter are out of sync.

Why is it important? 

I have a case with daily restarted collector that produces just a one message daily. Due to the described behavior, the cached stats are always filed with 1 - either because it wasn't updated, or because the first message was sent.

kamil-certat:
After restarting a bot, ensure that statistics about sent messages are zeroes in the stats cache. Previously, it was done after the first message to a path. Before this, the internal counter and stats cache were out of sync.

Fixes #2331

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,34 @@
 CHANGELOG
 ==========
 
+3.1.1 (unreleased)
+------------------
+
+### Core
+- Fixed not resetting destination path statistics in the stats cache after restarting bot (Fixes [#2331](https://github.com/certtools/intelmq/issues/2331))
+
+### Development
+
+### Bots
+
+#### Collectors
+
+#### Parsers
+
+#### Experts
+
+#### Outputs
+
+### Documentation
+
+### Tests
+
+### Packaging
+
+### Tools
+
+### Known Errors
+
 3.1.0 (2023-02-10)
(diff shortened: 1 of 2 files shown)


kamil-certat:
Any idea, why codespell complains about unrelated? Could you restart check, @sebix?

kamil-certat:
@sebix @gethvi @waldbauer-certat - may I ask for a review?

kamil-certat:
Codespell fixed in #2333

sebix:
> Codespell fixed in #2333

I merged the PR and rebased this one

sebix (approved):

Issue #2071, opened 2021-08-26
Wishlist for IntelMQ 4.x


aaronkaplan:
- [ ] all CLI components could highly benefit from using [typer](https://typer.tiangolo.com/tutorial/first-steps/)
- [ ] support redis streams
- [ ] LDAP enrichment bot (which must also be compatible with AD)
- [ ] Full Splunk support: interface the [Splunk API](https://docs.splunk.com/Documentation/Splunk/8.2.2/RESTUM/RESTusing)
- [ ] VT API enrichtment bot
- [ ] Support Jira as ticket system
- [ ] Support OTRS as ticket system
- [ ] TheHive interface
- [ ] new data format (see [IEP04](https://github.com/certtools/ieps/tree/main/004))

aaronkaplan:
okay, but it should not be a big problem to bring it and package it for bullseye for example. So, yeah... not a blocker. 
The question is - did you read the docs and would it make sense ? It looks really good for me, make the code very clean and lean and simple.

ghost:
It also supports shell completion (https://typer.tiangolo.com/typer-cli/), which is a big pro (our current solutions some some issues, e.g. #1240 #1561 #2094 ).
Still, before going for typer I'd like to see a comparison of different solution to check if it matches our requirements best (could even also be an IEP). E.g. one other (even more) widespread cli-library is click.

aaronkaplan:
+1 for IEP. But I am collecting ideas first in this issue :)

ghost:
maybe it makes sense to first define the problem(s) that should be solved

gethvi:
### Extending format based on classification.type.

Example 1:
`classification.type: "phishing"`

Optional additional fields:
`phishing.brand`: "Big-Bank-Name"
`phishing.screenshot` (maybe base64 data?)
`phishing.sector`: "banking"
`phishing.language`: EN
etc

Example 2:
`classification.type: "system-compromise"`

Optional additional fields:
`system-compromise.commands`: ["wget evil-url", "./evil-file"]
`system-compromise.vector`: "ssh" (or maybe "web-attacks", "sqli", etc)
etc.

### Add `destination.password`
When an event describes unauthorized login attempt, it would be nice to officialy "recognize" the password (for example honeypots provide this information).

### Add TLS info
If some service is running using TLS, we could accomodate info about certificate serial number or perhaps an issuer of the cert.


[...]

PR #2339, opened 2023-03-21
Update Spam-URL report.

elsif2:
Update the 'Spam-URL' report and add test script.


intelmq/bots/parsers/shadowserver/_config.py
```diff
@@ -4131,21 +4131,21 @@ def scan_exchange_identifier(field):
...
     'optional_fields': [
-        ('source.url', 'url', convert_http_host_and_url, True),
-        ('source.fqdn', 'http_host', validate_fqdn),
-        ('source.asn', 'asn', invalidate_zero),
-        ('source.geolocation.cc', 'geo'),
-        ('source.geolocation.region', 'region'),
-        ('source.geolocation.city', 'city'),
-        ('extra.', 'naics', invalidate_zero),
-        ('extra.', 'sector', validate_to_none),
+        ('destination.url', 'url', convert_http_host_and_url, True),
```
sebix: Does this report type really report a network connection with source/destination?

According to https://www.shadowserver.org/what-we-do/network-reporting/spam-url-report/ the url is "URL that was extracted from a Spam message" and `src_ip` is "IP [sic] of the URL", so both denote the same host, not two different ones.

sebix (changes requested):

intelmq/bots/parsers/shadowserver/_config.py
elsif2: The `src_ip` is the IP that originated the spam message and the `ip` is resolved from the URL so I mapped the URL details to _destination_.   Is there a better option?

sebix: Is https://www.shadowserver.org/what-we-do/network-reporting/spam-url-report/ correct?

elsif2: The web page needs an update as well:

- src | IP address of the Spam relay that delivered the message (last hop)
+ src_ip | IP address of the Spam relay that delivered the message (last hop)

sebix: Then the question is, what is this report primarily about? The spam relays or the URLs in the spam mails? Previously it was the URL (mapping `('source.url', 'url', convert_http_host_and_url, True),`). If we change the mapping (`url` to `destination.url` instead of `source.url`) IntelMQ users need to adapt their workflows to get the same result as before.
What do others think, especially users which process Shadowserver feeds?

elsif2: The report is about the URL as the src_ip is secondary.  Using source or destination for the URL isn't ideal.

elsif2: If compatibility is the driving factor, I can map the URL details to _source_.  How should the `src_ip` details be included?

elsif2: Please let me know if this would be preferred:

```
# http://www.shadowserver.org/wiki/pmwiki.php/Services/Spam-URL
spam_url = {
    'required_fields': [
        ('time.source', 'timestamp', add_UTC_to_timestamp),
        ('source.ip', 'ip', validate_ip),
    ],
    'optional_fields': [
        ('source.url', 'url', convert_http_host_and_url, True),
        ('source.fqdn', 'host', validate_fqdn),
        ('extra.relay.asn', 'src_asn', invalidate_zero),
        ('extra.relay.geolocation.cc', 'src_geo', validate_to_none),
        ('extra.relay.geolocation.region', 'src_region', validate_to_none),
        ('extra.relay.geolocation.city', 'src_city', validate_to_none),
        ('extra.relay.naics', 'src_naics', invalidate_zero),
        ('extra.relay.sector', 'src_sector', validate_to_none),
        ('source.asn', 'asn', invalidate_zero),
        ('source.geolocation.region', 'region'),
        ('source.geolocation.city', 'city'),
        ('extra.', 'naics', invalidate_zero),
        ('extra.', 'sector', validate_to_none),
        ('extra.', 'source', validate_to_none),
        ('extra.', 'sender', validate_to_none),
        ('extra.', 'subject', validate_to_none),
        ('malware.hash.md5', 'md5', validate_to_none),
    ],
    'constant_fields': {
        'classification.taxonomy': 'abusive-content',
        'classification.type': 'spam',
        'classification.identifier': 'spam-url',
    },
}
```

elsif2: Revised version submitted.

gethvi: I think the `source.ip` (and generally the whole `source` key) should describe the **relay**.

For the information about the spam message (subject, URL and it's IP, etc) I would suggest to use `extra.spam.*` keys.

---

This is something I mentioned in the Intelmq v4 wishlist issue:
https://github.com/certtools/intelmq/issues/2071#issuecomment-1105718722

Having an optional key based on `classification.type` (such as `spam`, `phishing`, etc) to hold additional information. If this is agreed on we can move from `extra.spam.*` keys to just `spam.*` keys.

kamil-certat: I agree we need better handling of additional information (like described by @gethvi), but for compatibility reasons, I would keep the previous meaning of the fields at least until the next major review release. Otherwise, users of IntelMQ may unexpectedly start notifying other parties as planned.

sebix: More "IntelMQ-like" is: emit two events, one event for the relay and one event for the Spam URL. (for the potential linking between those, see [IEP 004](https://github.com/certtools/ieps/tree/main/004))

elsif2: Can two IntelMQ events be generated from one source event?

sebix: > Can two IntelMQ events be generated from one source event?

Yes. Any parser can generate an arbitrary number of events (also zero!) from one report.

elsif2: I think that will have to wait until support for multiple outputs can be added to the parser.

I have two new reports in the queue to be added next week.  Should I put those in a separate PR?

gethvi: Yes I think it's better to put them in a separate PR.

Issue #1444, opened 2019-09-04
Deprecations and breaking changes in 4.0.0


wagner-certat:
- [ ] `intelmq.bots.collectors.mail.collector_mail_attach`: parameter `attach_unzip`

aaronkaplan:
??? --verbose pls.

wagner-certat:
Same as #1404 just for Version 4.0.0

gethvi:
* [ ] `intelmq.bots.parsers.generic.parser_csv`:  Remove parameter `type` (superseded by `default_fields`)

gethvi:
* [ ] `intelmq.bots.experts.url2fqdn.expert`:  Remove bot (superseded by URL expert)

sebix:
- [ ] drop `intelmq.lib.utils.resolve_dns` (#2352)

gethvi:
- [ ] remove functions `parse_utc_isoformat`, `convert_from_format`, `convert_from_format_midnight` and `convert_fuzzy` in harmonization.DateTime (deprecated and renamed for consistency in #2329)

gethvi:
Move to LSB paths by default as agreed in #2349

sebix:
- [ ] https://github.com/certtools/intelmq/issues/2543

PR #2352, opened 2023-04-19
ENH: cymru expert: prevent extra lookups, handle invalid asnames

### LINKED ISSUE #2132: Cymru whois expert: handle strange ASN symbols better
I discovered a slight issue with the Cymru whois expert bot with some strange AS names. For instance, AS266522.
```
$ dig +short TXT AS266522.asn.cymru.com         
"266522 | BR | lacnic | 2017-01-20 | Realnet Telecomunicacoes e Multimidia ltda\226\128\143 , BR"
```
`\226\128\143`, in unicode, decodes to u+200f, the right-to-left mark. When the IMQ bot attempts to decode it with `result['as_name'] = items[4].encode('latin1').decode('utf8')`, it runs into this issue:
`UnicodeEncodeError: 'latin-1' codec can't encode character '\u200f' in position 42: ordinal not in range(256)`

A solution which just permits decoding special symbols like RTL marks might not be the best, though just discarding these events into the dump file is also not a good option.

Various online tools for finding public data on ASes, found online via a quick search, give varying results.
* [ipinfo.io](https://ipinfo.io/AS266522) displays it in 3 different ways ("injected" RTL, `â€` and `aâ`)
* [qrator](https://radar.qrator.net/as266522) displays it as `&#8207`
* [bigdatacloud](https://www.bigdatacloud.com/asn-lookup/AS266522) displays it as `8207`
* [bgpview](https://bgpview.io/asn/266522) displays it as `ÂÂ `
* etc.

I suppose there's no existing expectation of invalid UTF-8 sequences or special symbols like RTL marks in AS names, nor any "standard" way of handling them, and I feel iffy about passing arbitrary symbols like that on to other systems.

Are there any ideas for how to handle cases like this? The simplest option would be to simply pass the undecoded string on in case the decoding step fails.

monoidic:
This PR fixes #2132 by ignoring AS names with invalid characters, as suggested by Sebastian.
It also ensures that Cymru DNS queries that return NXDOMAIN do not cause any extraneous queries due to search domains.

~~(TODO: changelog, add unit test with invalid characters)~~


sebix:
> (TODO: changelog, add unit test with invalid characters)

Yes, a unit test for one of these special cases would be very good. For example the AS266522, which you mentioned in the issue

The fix looks good on first glance.

monoidic:
Right. Well. Having the IP in the test event seemed to mess with things.

intelmq/bots/experts/cymru_whois/_lib.py
```diff
@@ -17,8 +17,8 @@
 import intelmq.lib.utils as utils
 from intelmq.lib.harmonization import IPAddress
 
-IP_QUERY = "%s.origin%s.asn.cymru.com"
-ASN_QUERY = "AS%s.asn.cymru.com"
+IP_QUERY = "%s.origin%s.asn.cymru.com."
+ASN_QUERY = "AS%s.asn.cymru.com."
```
kamil-certat: If it's possible, it would be nice to provide a test for this change as well

sebix: The tests cover IP and ASN lookups.

intelmq/bots/experts/cymru_whois/_lib.py
```diff
@@ -168,6 +168,8 @@ def __asn_query_parse(text):
         if items[4]:
             # unicode characters need to be decoded explicitly
             # with the help of https://stackoverflow.com/questions/60890590/
-            result['as_name'] = items[4].encode('latin1').decode('utf8')
+            as_name = items[4].encode('latin1', errors='ignore').decode('utf8')
+            if as_name:
```
sebix, line 172: is this the "Avoid extraneous search domain-based queries on NXDOMAIN result"?

intelmq/bots/experts/cymru_whois/_lib.py
monoidic: It technically doesn't cover the search domain part.

intelmq/bots/experts/cymru_whois/_lib.py
monoidic: That's the added periods in `IP_QUERY` and `ASN_QUERY`, actually.
Example: if you have a search domain of "lan" and resolving `homeserver` fails with NXDOMAIN, then the search domain will cause `homeserver.lan.` to be queried.
However, if, for instance, `1.2.3.4.origin4.cymru.com` returns NXDOMAIN, then `1.2.3.4.origin4.cymru.com.lan.` would be queried.
But, the added period at the end signifies that this is a fully qualified domain name and such queries should not be made.

intelmq/bots/experts/cymru_whois/_lib.py
kamil-certat: This is what I meant :)

monoidic:
The test explicitly sets the search domain to `cert.ee.`, which causes the test to fail if attempts are made to parse the TXT response because the responses for all subdomains of `cert.ee.` return an SPF record in response to TXT queries.

It is not a perfect solution (it relies on the `cert.ee.` domain, and the result for the IP in the test being NXDOMAIN, and the parser raising an exception on TXT records with invalid formatting instead of gracefully handling it somehow).

However, I'm wondering, is there any legitimate use case for `intelmq.lib.utils.dns.resolve_dns` to force the use of search domains in the first place? What about updating the minimum dnspython version to 2.x and removing `search=True` altogether?

sebix:
> However, I'm wondering, is there any legitimate use case for `intelmq.lib.utils.dns.resolve_dns` to force the use of search domains in the first place? What about updating the minimum dnspython version to 2.x and removing `search=True` altogether?

yes, I think we can safely require dnspython >= 2.0 now.

I don't have any reason at hand why search is on. If the tests succeed with search disabled, let's go for it :)

sebix:
@monoidic Is this PR ready to merge or do you want to change the DNS-related parts?

monoidic:
I'd been busy with other projects and forgout about this.
The latest commit *should* have everything necessary to deprecate dnspython <2.0.0 and remove two tests which are now outdated.
If the changes are acceptable, then I will add changelog entries as well.

intelmq/lib/utils.py
```diff
@@ -928,9 +928,6 @@ def resolve_dns(*args, **kwargs) -> dns.resolver.Answer:
 
     Parameters:
         see: https://dnspython.readthedocs.io/en/stable/resolver-class.html#dns.resolver.Resolver.resolve
-        The `search` parameter is always set to True for compatibility with dnspython version 1.
 
     """
-    if dns.version.MAJOR < 2:
-        return dns.resolver.query(*args, **kwargs)
-    return dns.resolver.resolve(*args, **kwargs, search=True)
+    return dns.resolver.resolve(*args, **kwargs)
```
sebix, line 933: With that change the function is obsolete now. But that's something for a major version as it is a change in the interface.

sebix (changes requested):
>  If the changes are acceptable, then I will add changelog entries as well.

Yes, please document the changed minimum required version of dnspython, this has impact on the supported OSs (dropping out Ubuntu 20.04, which need to be removed from the docs, from intelmq-vagrant and at the next release from the packages).

intelmq/lib/utils.py
kamil-certat: Could you leave a ticket to remove it?

kamil-certat (approved):
Looks good. What I'm not happy about is that we rely on the real answers in tests instead of some mocking solution (and DNS is relatively unstable), but it's a general problem, nothing related to the change :)

intelmq/lib/utils.py
sebix: Added to #1444

PR #2355, opened 2023-04-25
FIX: intelmqsetup - never take ownership of /

### LINKED ISSUE #2354: intelmqsetup changes the root directory ownership
`intelmqsetup` tries to fix ownership of related directories:

https://github.com/certtools/intelmq/blob/7674949656748164aede86e7dfc3aa39372add44/intelmq/bin/intelmqsetup.py#L172-L176

However, the `ROOT_DIR` can be set to `/`:

https://github.com/certtools/intelmq/blob/7674949656748164aede86e7dfc3aa39372add44/intelmq/__init__.py#L18-L20

This causes `intelmqsetup` to take over the ownership of the root directory, what could open some security risk as well as break some system actions (e.g. `systemd-tmpfiles`)

kamil-certat:
The tool intelmqsetup wants to change the owner ROOT_DIR path.
If instructed to install IntelMQ in LSB-style paths, it's set
to the '/' resulting in changing the owner of system root to
intelmq.

This case is rare to happen (requires explixitly set INTELMQ_PATHS_NO_OPT
variable and using PIP package or directly the source code,
as the native package doesn't contain intelmqsetup), but it's
still potentially dangerous and can cause the system degradation
(e.g. prevents systemd-tmpfiles from working correctly).

Fixes: #2354

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ CHANGELOG
 
 ### Tools
 - `intelmqsetup`:
-  - SECURITY: fixed a low-risk bug causing the tool to change owner of `/` path if running with the `INTELMQ_PATHS_NO_OPT` environment variable set to `True`. This affects only the PIP package as the DEB package doesn't contain this tool. (PR#2355 by Kamil Mańkowski)
+  - SECURITY: fixed a low-risk bug causing the tool to change owner of `/` if run with the `INTELMQ_PATHS_NO_OPT` environment variable set. This affects only the PIP package as the DEB/RPM packages don't contain this tool. (PR#2355 by Kamil Mańkowski, fixes #2354)
 
 ### Known Errors
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,8 @@ CHANGELOG
 ### Packaging
 
 ### Tools
+- `intelmqsetup`:
+  - SECURITY: fixed a low-risk bug causing the tool to change owner of `/` path if running with the `INTELMQ_PATHS_NO_OPT` environment variable set to `True`. This affects only the PIP package as the DEB package doesn't contain this tool. (PR#2355 by Kamil Mańkowski)
 
 ### Known Errors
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -41,6 +41,8 @@ CHANGELOG
 ### Packaging
 
 ### Tools
+- `intelmqsetup`:
+  - SECURITY: fixed a low-risk bug causing the tool to change owner of `/` path if running with the `INTELMQ_PATHS_NO_OPT` environment variable set to `True`. This affects only the PIP package as the DEB package doesn't contain this tool. (PR#2355 by Kamil Mańkowski)
```
sebix: The value of `INTELMQ_PATHS_NO_OPT` does not matter as long as it is set (the code checks for existence).
```suggestion
  - SECURITY: fixed a low-risk bug causing the tool to change owner of `/` if run with the `INTELMQ_PATHS_NO_OPT` environment variable set. This affects only the PIP package as the DEB/RPM packages don't contain this tool. (PR#2355 by Kamil Mańkowski, fixes #2354)
```

intelmq/bin/intelmqsetup.py
```diff
@@ -95,6 +95,9 @@ def create_directory(directory: str, octal_mode: int):
 
 
 def change_owner(file: str, owner: Optional[str] = None, group: Optional[str] = None, log: bool = True):
+    if Path(file).as_posix() in ['/', '//']:
```
sebix, line 98:
```suggestion
    if Path(file).as_posix() in ('/', '//'):
```

sebix: Why checking for `//`?

sebix (changes requested):
Except for the changelog change suggestion, ready for merge.
Thanks for the tests :)

intelmq/bin/intelmqsetup.py
kamil-certat: It's just an additional precaution, but if you look into [docs](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.root): posix root paths are standardized to `/`, except for `//`. Both are recognized as root directory. In fact, the later case is highly unlikely. It can be only manually set by the user, but even then - IntelMQ should never take the ownership of root.

kamil-certat:
@sebix Do you have any questions more, or we could merge it?

sebix (approved):

PR #2357, opened 2023-04-26
Some Bugfixes

wagner-intevation:
* doc: lib/pipeline: move arch docs to module docstring
* cache mixin: remove unused operation
* bugfix: domain suffix: fix parameter error message
* lib/bot: use *args, **kwargs for derived classes' constructors
  avoids errors when parameters of Bot change
  now all derived classes are always compatible
* lib/bot: load logging defaults from constants

PR #2358, opened 2023-05-03
IEP007 implementation: Library mode

wagner-intevation:
This is a draft implementation of [IEP 007 (Running IntelMQ bots as Python Library)](https://github.com/certtools/ieps/tree/iep-007/007/)

### DIFF
--- a/intelmq/lib/datatypes.py
+++ b/intelmq/lib/datatypes.py
@@ -1,13 +1,13 @@
-# SPDX-FileCopyrightText: 2021 Birger Schacht
+# SPDX-FileCopyrightText: 2021 Birger Schacht, 2023 Bundesamt für Sicherheit in der Informationstechnik (BSI)
 #
 # SPDX-License-Identifier: AGPL-3.0-or-later
-from datetime import datetime
 from enum import Enum
 from inspect import signature
+from sys import version_info
 from typing import Optional, Callable, Union, List
 
+from datetime import datetime
 from termstyle import green
-
 from intelmq.lib.exceptions import InvalidArgument
 from intelmq.lib.harmonization import DateTime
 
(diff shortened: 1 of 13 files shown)


wagner-intevation:
This branch is based on #2357 , merge that one first

intelmq/lib/datatypes.py
```diff
@@ -137,3 +137,21 @@ def validate(value: str) -> [Callable, Optional[str]]:
...
+        """
+        def __or__(self, other: dict) -> 'Dict39':
+            """
+            Create a new dictionary with the merged keys and values of d and other, which must both be dictionaries. The values of other take priority when d and other share keys.
+            """
+            ret = Dict39(self.copy())
+            ret.update(other)
+            return ret
+else:
+    Dict39 = dict
```
gethvi, line 157: Just a note: this built-in merging of dicts only works for the first level (root) keys. It is not recursive. Which can bring really nasty hidden bugs. I don't see it being a problem now, but if ever used for nested dicts (which is not unreasonable) it can wreak havoc.

Example:
```python
x = {
    "KEY" : {
        "AAA": 10,
        "BBB": 10,
    }
}


y = {
    "KEY" : {
        "AAA": 20,
        "XXX": 20,
    }
}

x.update(y)
print(x)

# {'KEY': {'AAA': 20, 'XXX': 20}}
# from "merging" perspective it should keep the "BBB": 10 as well in the result
```

For one of my projects I created this, might be useful:
```python
from copy import deepcopy
from typing import Dict, Optional

def merge_dicts(d1: Dict, d2: Dict, result: Optional[Dict] = None) -> Dict:
    if not result:
        result = deepcopy(d1)

    for key, value in d2.items():

        # if it's not conflict, we just add to the result
        if key not in result:
            result[key] = value

        # if there is a conflict, but it isn't a dict, we just replace
        elif key in result and not isinstance(value, dict):
            result[key] = value

        # if there is a conflict and it is a dict, we need to merge
        elif key in result and isinstance(value, dict):
            merge_dicts(result[key], value, result[key])

    return result
```

sebix: > Just a note: this built-in merging of dicts only works for the first level (root) keys. It is not recursive.

Good catch. But that's fine here and IMO also good. A bot-definition in `B` of `A | B` eliminates any existing definitions in `A`, so there are no leftovers which could lead to unexpected behaviour. Same for dict-typed parameters.

wagner-intevation:
rebased on develop to fix conflicts and test fails

Issue #2360, opened 2023-05-08
Shadowserver parser may fail to autodetect report type after reload


kamil-certat:
## Handling feed detection

During the `init` method the parser try to determine the feed type detection type [0]:

https://github.com/certtools/intelmq/blob/7674949656748164aede86e7dfc3aa39372add44/intelmq/bots/parsers/shadowserver/parser.py#L42-L43

If the `self.feedname` isn't set, then the mode is `detect` [1]:
https://github.com/certtools/intelmq/blob/7674949656748164aede86e7dfc3aa39372add44/intelmq/bots/parsers/shadowserver/parser.py#L53-L54

Later, during processing a report, the feed name is detected and eventually set [2]:

https://github.com/certtools/intelmq/blob/7674949656748164aede86e7dfc3aa39372add44/intelmq/bots/parsers/shadowserver/parser.py#L76-L79

## Reloading bot

Then, during the reload (not restart) happens, the bot is re-initialized without creating a new object:

https://github.com/certtools/intelmq/blob/7674949656748164aede86e7dfc3aa39372add44/intelmq/lib/bot.py#L271-L285

This mechanism is used during log rotation by our standard Debian configuration: 

https://github.com/certtools/intelmq/blob/7674949656748164aede86e7dfc3aa39372add44/contrib/logrotate/intelmq#L12

and of course re-load the configuration, refreshing all parameters set in the config file [3]:

https://github.com/certtools/intelmq/blob/7674949656748164aede86e7dfc3aa39372add44/intelmq/lib/bot.py#L786-L788

### Requirements

* Assuming you want parser to detect the feed, so you didn't set any value in `parametrs.feedname` in the parser config. 
* You have also installed our Debian package or manually configured log rotation with reloading.
* Your IntelMQ instance is working long enough for log rotation to occur.

### Case steps

1. After starting bot, the mode is correctly detect [1], and then handling every report a correct feed is detected and set [2].
2. The log rotation occurs, and the parser is reloaded.
3. The configuration is refreshed, overriding all fields set in the config file [3].
4. Bot `init` method is called once more and tries to detect mode, but **`self.feedname` is still set with the last value before reload, and the `fixed` mode is detected** [0].
5. **On every next report handled, the last previously detected feed is used**

## Workarounds

Set the `feedname` as `null` in the config. However, due to the IntelMQ Manager issue with nulls https://github.com/certtools/intelmq-manager/issues/294 this may be changed to an empty string and case the fail again.

## Fixes

[...]

PR #2361, opened 2023-05-08
Updated to reset report type on reload

elsif2:
Fix for #2360.

### DIFF
--- a/intelmq/bots/parsers/shadowserver/parser.py
+++ b/intelmq/bots/parsers/shadowserver/parser.py
@@ -206,5 +206,8 @@ class ShadowserverParserBot(ParserBot):
 
         yield event
 
+    def shutdown(self):
+        self.feedname = None
+
 
 BOT = ShadowserverParserBot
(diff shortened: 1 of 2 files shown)


intelmq/bots/parsers/shadowserver/parser.py
```diff
@@ -206,5 +206,7 @@ def parse_line(self, row, report):
 
         yield event
 
+    def shudown(self):
```
kamil-certat:
```suggestion
    def shutdown(self):
```

kamil-certat (commented):
Thank you!

intelmq/bots/parsers/shadowserver/parser.py
```diff
@@ -206,5 +206,7 @@ def parse_line(self, row, report):
 
         yield event
 
+    def shutdown(self):
+        self.feedname = None
```
kamil-certat, line 210: Linter would be happy with an additional new line after this one

kamil-certat (approved):

PR #2362, opened 2023-05-11
sql output: new parameter `fail_on_error`

wagner-intevation:
Introduce a new parameter for the SQL Output bot to immediately fail on errors, not trying to fix the issue or waiting for some time.

### DIFF
--- a/intelmq/lib/mixins/sql.py
+++ b/intelmq/lib/mixins/sql.py
@@ -27,6 +27,8 @@ class SQLMixin:
     # overwrite the default value from the OutputBot
     message_jsondict_as_string = True
     reconnect_delay = 0
+    # If process()/execute() should raise exceptions (True) or rollback (False)
+    fail_on_errors = False
 
     def __init__(self, *args, **kwargs):
         self._init_sql()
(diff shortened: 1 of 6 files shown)


intelmq/lib/mixins/sql.py
```diff
@@ -116,7 +116,7 @@ def execute(self, query: str, values: tuple, rollback=False):
             self.logger.debug('Done.')
         except (self._engine.InterfaceError, self._engine.InternalError,
                 self._engine.OperationalError, AttributeError):
-            if rollback:
+            if rollback and not fail_on_errors:
```
kamil-certat: There is no `fail_on_errors` in this context. Please add a negative test to catch such issue.

kamil-certat (changes requested):

intelmq/lib/mixins/sql.py
wagner-intevation: Added test cases for exceptions raised at execution, for `fail_on_errors` = True and `fail_on_errors` = False

kamil-certat (approved):

wagner-intevation:
Thanks, will merge after the tests succeed

PR #2368, opened 2023-05-24
Release 3.2.0

sebix:
Already based on #2358 and added a version update to 3.2.0 alpha 1 for testing. If that succeeds, will update the version to 3.2.0 (stable).


wagner-intevation:
rebased on develop to fix conflicts and test fails

sebix:
all depending PRs are merged
rebased on develop, so the PR is now much smaller and only contains the commits for the release

PR #2370, opened 2023-05-24
tst: skip tests which require an intelmq installation by default

### LINKED ISSUE #2369: intelmqctl and intelmqdump tests require an installed IntelMQ instance
This is a revival of #2300 (ctl tests require an installed IntelMQ instance). These tests are affected now:

- TestIntelMQController::test_check_handles_syntaxerror_when_importing_bots
- TestIntelMQController::test_check_pass_with_default_runtime
- TestIntelMQController::test_check_passed_with_default_harmonization_and_empty_runtime
- TestIntelMQDump::test_get_dump_for_one_bot
- TestIntelMQDump::test_list_and_select_dump_from_custom_location
- TestIntelMQDump::test_list_and_select_dump_from_global_location
- TestIntelMQDump::test_list_dumps_for_all_bots_from_custom_locations
- TestIntelMQDump::test_list_dumps_for_all_bots_from_default_log_path
- TestIntelMQDump::test_selecting_dump_warns_when_filename_is_ambiguous

the ctl tests have a traceback like this:
```
[  152s]   File "/usr/src/packages/BUILD/.pybuild/cpython3_3.9_intelmq/build/intelmq/lib/utils.py", line 224, in load_configuration
[  152s]     raise ValueError('File not found: %r.' % configuration_filepath)
[  152s] ValueError: File not found: '/etc/intelmq/runtime.yaml'.
[  152s] ERROR    intelmqctl:intelmqctl.py:127 No valid IntelMQ installation found: DistributionNotFound
```
and the dump tests don't find any dumps.

So I suggest to add a new skip-decorator and environment variable to skip tests depending on an installation by default. This also addresses https://lists.cert.at/pipermail/intelmq-dev/2023-May/000610.html

wagner-intevation:
fixes certtools/intelmq#2369

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -75,4 +75,9 @@ jobs:
 
     - name: Run full testsuite
       if: ${{ matrix.type == 'full' }}
-      run: TZ=utc INTELMQ_TEST_DATABASES=1 INTELMQ_TEST_EXOTIC=1 pytest --cov intelmq/ --cov-report=xml --cov-branch intelmq/ contrib/
+      run: pytest --cov intelmq/ --cov-report=xml --cov-branch intelmq/ contrib/
+      env:
+        TZ: utz
+        INTELMQ_TEST_DATABASES: 1
+        INTELMQ_TEST_EXOTIC: 1
+        INTELMQ_TEST_INSTALLATION: 1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ CHANGELOG
 ### Documentation
 
 ### Tests
+- New decorator `skip_installation` and environment variable `INTELMQ_TEST_INSTALLATION` to skip tests requiring an IntelMQ installation on the test host by default (PR#2370 by Sebastian Wagner, fixes #2369)
 
 ### Packaging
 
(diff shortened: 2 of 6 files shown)


kamil-certat (approved):
Approved, but I'd like to see it as an exceptional solution – and in the long term, rather develop solution to fully isolate IntelMQ tests.

sebix:
> Approved, but I'd like to see it as an exceptional solution – and in the long term, rather develop solution to fully isolate IntelMQ tests.

I agree, but I want to add two more points here:
1. as long as the tests require an installation, this solution is acceptable as it ensures isolation. The skips should be removed once the testes are isolated (using mocking).
2. Unit tests that depend on an installation (and are not isolated/mocked) are more accurate and have better coverage, as the mocking can be inaccurate for several reasons. But there are better ways than having such integration tests as part of a test suite which is expected to be isolated.

kamil-certat:
> Unit tests that depend on an installation (and are not isolated/mocked) are more accurate and have better coverage, as the mocking can be inaccurate for several reasons. But there are better ways than having such integration tests as part of a test suite which is expected to be isolated.

just a nit: I believe that if it's possible, it's better to isolate by providing e.g. temporary working directory for tests than mocking, so the full code path is run. I didn't mean `isolation` = `mocking` ;)

sebix:
> > Unit tests that depend on an installation (and are not isolated/mocked) are more accurate and have better coverage, as the mocking can be inaccurate for several reasons. But there are better ways than having such integration tests as part of a test suite which is expected to be isolated.
> 
> just a nit: I believe that if it's possible, it's better to isolate by providing e.g. temporary working directory for tests than mocking, so the full code path is run. I didn't mean `isolation` = `mocking` ;)

Yep, even better!

Issue #2367, opened 2023-05-24
Breaking changes in 'rt' 3.0 library


kamil-certat:
The library to connect with RT system has some breaking changes since version 3: [changelog](https://github.com/python-rt/python-rt/blob/master/CHANGELOG.md)

* The class for REST v1.0 was moved,
* Some exceptions may have new name.

RT-related bots requires pinning version below 3 to work at the moment, and probably minor changes to work with the new lib. For the future: probably would be good to move to the REST2.

PR #2371, opened 2023-05-25
FIX: Restrict python-rt version

kamil-certat:
The `python-rt` in version 3 introduced some small backward incompatible changes. 

Related changes are tracked in #2367

### DIFF
--- a/docs/user/bots.rst
+++ b/docs/user/bots.rst
@@ -554,7 +554,7 @@ You need the rt-library >= 1.9 and < 3.0 from nic.cz, available via `pypi <https
 
 .. warning::
 
-   Actually the bot supports the python-rt library in the version older than 3.0.
+   At the moment, the bot only supports `python-rt` versions below 3.0.
 
 This rt bot will connect to RT and inspect the given `search_queue` for tickets matching all criteria in `search_*`,
 Any matches will be inspected. For each match, all (RT-) attachments of the matching RT tickets are iterated over and within this loop, the first matching filename in the attachment is processed.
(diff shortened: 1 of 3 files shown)


sebix (changes requested):
Could you please add a small note in docs/user/bots.rst too?

If the user has rt 3.0.0 installed, the bot will just fail. by checking the rt version in the bot, the error message would be more precise (the mongodb output bot does this for example)

docs/user/bots.rst
```diff
@@ -550,7 +550,11 @@ Request Tracker
 * `cache (redis db):` none
 * `description:` Request Tracker Collector fetches attachments from an RTIR instance.
 
-You need the rt-library >= 1.9 from nic.cz, available via `pypi <https://pypi.org/project/rt/>`_: `pip3 install rt`
+You need the rt-library >= 1.9 and < 3.0 from nic.cz, available via `pypi <https://pypi.org/project/rt/>`_: `pip3 install rt`
```
sebix:
```suggestion
You need the rt-library >= 1.9 and < 3.0 from nic.cz, available via `pypi <https://pypi.org/project/rt/>`_: `pip3 install 'rt<3'`
```

docs/user/bots.rst
```diff
@@ -550,7 +550,11 @@ Request Tracker
...
+
+.. warning::
+
+   Actually the bot supports the python-rt library in the version older than 3.0.
```
sebix:
```suggestion
   At the moment, the bot only supports `python-rt` versions below 3.0.
```
"Actually" does not mean "now"

sebix (changes requested):

docs/user/bots.rst
kamil-certat: Agh, right!

PR #2372, opened 2023-05-28
Shadowserver dynamic config

elsif2:
## Dynamic configuration

Many thanks to Eric Halil for collaboration and testing on this update.

#### Parsers
- `intelmq.bots.parsers.shadowserver._config`:
  - Switch to dynamic configuration to decouple report schema changes from IntelMQ releases.  Please see intelmq/bots/parsers/shadowserver/README.md for details. 

#### Collectors
- `intelmq.bots.collectors.shadowserver.collector_reports_api`:
  - The 'json' option is no longer supported as the 'csv' option provides better performance.  Please see intelmq/bots/parsers/shadowserver/README.md for a sample configuration.


sebix:
@elsif2 the license check is failing

elsif2:
From the log:
```
The following files have no copyright and licensing information:
* intelmq/bots/parsers/shadowserver/schema.json.test
```

License added.

intelmq/bots/collectors/shadowserver/collector_reports_api.py
```diff
@@ -67,11 +67,10 @@ def init(self):
             self._report_list.append(self.country)
 
         if self.file_format is not None:
-            if not (self.file_format == 'csv' or self.file_format == 'json'):
+            if not (self.file_format == 'csv'):
                 raise ValueError('Invalid file_format')
```
kamil-certat: Could you write in this error, which format is currently supported?

intelmq/bots/collectors/shadowserver/collector_reports_api.py
```diff
@@ -34,7 +34,7 @@ class ShadowServerAPICollectorBot(CollectorBot, HttpMixin, CacheMixin):
             A list of strings or a comma-separated list of the mailing lists you want to process.
         types (list):
             A list of strings or a string of comma-separated values with the names of reporttypes you want to process. If you leave this empty, all the available reports will be downloaded and processed (i.e. 'scan', 'drones', 'intel', 'sandbox_connection', 'sinkhole_combined').
-        file_format (str): File format to download ('csv' or 'json').  The default is 'json' for compatibility. Using 'csv' is recommended for best performance.
+        file_format (str): File format to download ('csv').  The 'json' option is no longer supported.
```
kamil-certat: This is a breakable change, could you please implement the `check` method (see: https://github.com/certtools/intelmq/blob/bcccbf501ffa8b64c2ea011d457c3409ad1dd4cd/intelmq/lib/bot.py#L964) to validate the parameters during `intelmqctl check`? This would allow IntelMQ users to easy move to this new requirement.

And just to be sure, the parser still support CSV, right? In our instance we still rely on email reports (I know, we need to move to the API, but it's still in the backlog).

intelmq/bots/parsers/shadowserver/parser.py
```diff
@@ -22,14 +22,21 @@
...
 import intelmq.bots.parsers.shadowserver._config as config
 
 
 class ShadowserverParserBot(ParserBot):
-    """Parse all ShadowServer feeds"""
+    """
+    Parse all ShadowServer feeds
+
+    Parameters:
+        schema_file (str): Path to the report schema file
```
kamil-certat: This parameter isn't configured anywhere.

intelmq/bots/parsers/shadowserver/parser.py
```diff
@@ -40,6 +47,11 @@ class ShadowserverParserBot(ParserBot):
     overwrite = False
 
     def init(self):
+        config.set_logger(self.logger)
+        try:
+            config.update_schema()
+        except Exception as e:
+            self.logger.warning("Schema update failed: %s." % e)
```
kamil-certat: Please do not do an unconditional update on every change. Please allow setting a parameter whether the bot should update the schema or not. This is unfortunately a debugging hell, and in addition hard to monitor – a warning in log will mostly be silently ignored, and we would not recognize it for the long time. I'd like to have a control over this behavior, e.g. disable automated update when I have a cronjob configured.

intelmq/bots/parsers/shadowserver/README.md
```diff
@@ -7,3 +7,45 @@ This module is maintained by [The Shadowserver Foundation](https://www.shadowser
 
 Please contact intelmq@shadowserver.org with any issues or concerns.
 
+The report configuration is now stored in a _schema.json_ file downloaded from https://interchange.shadowserver.org/intelmq/v1/schema.
+
+For environments that have internet connectivity the `update_schema.py` script should be called from a cron job to obtain the latest revision.
+The parser will attempt to download a schema update on startup unless INTELMQ_SKIP_INTERNET is set.
```
kamil-certat: :red_circle: Please do not use the  `INTELMQ_SKIP_INTERNET` to manage the bot behavior. This is a test-only variable, not intended to configure bots. For bots, please use the bot parameter's for configuring it's behaviour.

intelmq/bots/parsers/shadowserver/README.md
```diff
@@ -7,3 +7,45 @@ This module is maintained by [The Shadowserver Foundation](https://www.shadowser
...
+
+For air-gapped systems automation will be required to download and copy the _schema.json_ file into this directory.
```
kamil-certat: Could you please include a starting `schema.json` for the initial start or air-gaping systems? It would be great to update it regularly (like once a month or so), so new version of IntelMQ will still get not really outdated schema even without downloading it, and then use provided service to keep up-to-date with ShadowServer. This would be truly user's decision, if they want to use automated upgrades or not.

kamil-certat: :red_circle: Please do not require saving any editable file _in this directory_. This directory lives in the library path, that shouldn't be updated besides of lib upgrades, but most importantly, it would break with default settings, because:

- the library path can be owned by root, and bots are usually run as `intelmq` user – the bot cannot write to this path;
- if using dockerized environment, this path can be set as not writable at all: in this case, regardless of the user and owner, bot may not be able to write update.

Instead, please:
- as a default path, use a subdirectory of `intelmq.VAR_STATE_PATH` – this is a great place to write data (and this is how bots usually work)
- then allow configuring this path in bot's parameters.

intelmq/bots/parsers/shadowserver/_config.py
```diff
@@ -241,4125 +258,92 @@ def scan_exchange_identifier(field):
...
+    'convert_date': convert_date,
+    'convert_date_utc': convert_date_utc,
+    'force_base64': force_base64,
+    'scan_exchange_taxonomy': scan_exchange_taxonomy,
+    'scan_exchange_type': scan_exchange_type,
+    'scan_exchange_identifier': scan_exchange_identifier,
+}
+
+
+def reload():
```
kamil-certat, line 314: :yellow_circle:  Here, the same as in other places, I'd like to suggest not doing a silent, not configurable reloading. As an administrator, I find it troublesome to debug what wrong has happened if configuration is changing unattended and irregularly. Although, I see a profit for first-time users and teams without dedicated admins. As so, I'd like to suggest:

1. Please add a configuration option whether the config should be reloaded automatically or not.
2. Load configuration on bot's startup, not when asked to parse (please see, delaying this process may hide configuration issues unless the bot is really needed for critical work!).
3. After the regular schema update (a.g. cron job), please trigger reloading configuration - you can find example how to do it at the end of this update procedure: https://github.com/certtools/intelmq/blob/bcccbf501ffa8b64c2ea011d457c3409ad1dd4cd/intelmq/bots/experts/asn_lookup/expert.py#L95

intelmq/bots/parsers/shadowserver/_config.py
```diff
@@ -241,4125 +258,92 @@ def scan_exchange_identifier(field):
...
+    mtime = 0.0
+
+    if os.path.isfile(__config.schema_file):
+        mtime = os.path.getmtime(__config.schema_file)
+        if __config.schema_mtime == mtime:
+            return
+    else:
+        __config.logger.info("The schema file does not exist.")
+
+    if __config.schema_mtime == 0.0 and mtime == 0.0 and not os.environ.get('INTELMQ_SKIP_INTERNET'):
```
kamil-certat: As being told, do not use test-only env variables for production behavior. Use your bot's parameters.

intelmq/bots/parsers/shadowserver/_config.py
```diff
@@ -241,4125 +258,92 @@ def scan_exchange_identifier(field):
...
-    'constant_fields': {
-        'classification.taxonomy': 'vulnerable',
-        'classification.type': 'vulnerable-system',
-        'classification.identifier': 'special',
-    },
-}
+    (th, tmp) = tempfile.mkstemp(dir=os.path.dirname(__file__))
+    url = 'https://interchange.shadowserver.org/intelmq/v1/schema'
+    try:
+        urllib.request.urlretrieve(url, tmp)
```
kamil-certat: :red_circle: This code do not respect the proxy configuration (from the intelmq config), what means, will not work in many environments.

Instead, please take a look how downloading of such things is done in other bots, like here:
https://github.com/certtools/intelmq/blob/bcccbf501ffa8b64c2ea011d457c3409ad1dd4cd/intelmq/bots/experts/asn_lookup/expert.py#L118-L120C45

intelmq/bots/parsers/shadowserver/parser.py
kamil-certat: Please add `exc_info=True` to the logger, otherwise you lose the stack trace.

intelmq/bots/parsers/shadowserver/_config.py
```diff
@@ -241,4125 +258,92 @@ def scan_exchange_identifier(field):
...
-        ('extra.', 'method', validate_to_none),
-        ('extra.', 'device_vendor', validate_to_none),
-    ],
-    'constant_fields': {
-        'classification.taxonomy': 'vulnerable',
-        'classification.type': 'vulnerable-system',
-        'classification.identifier': 'special',
-    },
-}
+    (th, tmp) = tempfile.mkstemp(dir=os.path.dirname(__file__))
```
kamil-certat: :red_circle: Please note, forcing the directory were tempfile should be created, can cause a lot of trashes – this file may not be automatically cleared – or am I wrong, and the tempfile will be treated as it regardless of the location?

Instead, please use maybe the default temp directory, is any reason why not do it? I see the idea of leaving tempfile for debugging, but maybe do it in a way, that it's copied on fail? Or just use the VAR path, not the library path - this would be okay for me as well.

intelmq/bots/parsers/shadowserver/_config.py
```diff
@@ -241,4125 +258,92 @@ def scan_exchange_identifier(field):
...
-# END CONFGEN
+    new_version = ''
+    old_version = ''
 
-feedname_mapping = {feedname: function for feedname, filename, function in mapping}
-filename_mapping = {filename: (feedname, function) for feedname, filename, function in mapping}
+    try:
+        with open(tmp) as fh:
+            schema = json.load(fh)
+            new_version = schema['_meta']['date_created']
```
kamil-certat, line 372: Could you please log which version was downloaded?

intelmq/bots/parsers/shadowserver/_config.py
```diff
@@ -241,4125 +258,92 @@ def scan_exchange_identifier(field):
...
+                old_version = schema['_meta']['date_created']
+            if new_version != old_version:
+                os.replace(__config.schema_file, ".".join([__config.schema_file, 'bak']))
+        except:
+            pass
+
+    if new_version != old_version:
+        os.replace(tmp, __config.schema_file)
+    else:
+        os.unlink(tmp)
```
kamil-certat, line 395: If any error occurs in the meantime, this code would be reached – even if the problem is not related to the downloaded schema. Maybe use a context manager? But okay, I'm not sure here – leaving the downloaded file sounds nice as well.

intelmq/bots/parsers/shadowserver/_config.py
kamil-certat: Would be nice to add debug log to note where the file will be saved (path of tmp file)

intelmq/bots/parsers/shadowserver/_config.py
```diff
@@ -241,4125 +258,92 @@ def scan_exchange_identifier(field):
...
+        # leave tempfile behind for diagnosis
+        raise ValueError("Failed to validate %r" % tmp)
+
+    if os.path.exists(__config.schema_file):
+        try:
+            with open(__config.schema_file) as fh:
+                schema = json.load(fh)
+                old_version = schema['_meta']['date_created']
+            if new_version != old_version:
+                os.replace(__config.schema_file, ".".join([__config.schema_file, 'bak']))
```
kamil-certat, line 385: Nice move with renaming! Could you add info log to note, that the update was made and which version were and is now?

intelmq/bots/parsers/shadowserver/update_schema.py
```diff
@@ -0,0 +1,11 @@
+# SPDX-FileCopyrightText: 2023 The Shadowserver Foundation
```
kamil-certat: As a general thing, I'd suggest using the approach from e.g. ASN expert bot, where the updating code is a method in bot class and can be called with the right parameter, see: https://github.com/certtools/intelmq/blob/bcccbf501ffa8b64c2ea011d457c3409ad1dd4cd/intelmq/bots/experts/asn_lookup/expert.py#L77

intelmq/tests/bots/parsers/shadowserver/test_report_smb.py
```diff
@@ -0,0 +1,117 @@
...
+    """
+    A TestCase for a ShadowserverParserBot.
+    """
+
+    @classmethod
+    def set_bot(cls):
+        cls.bot_reference = ShadowserverParserBot
+        cls.default_input_message = EXAMPLE_REPORT
+
+    def test_event(self):
```
kamil-certat, line 109: Please note, this test would work only if the internet connection is available. Please add the `skip_internet` decorator.

intelmq/tests/bots/parsers/shadowserver/test_report_smb.py
```diff
@@ -0,0 +1,117 @@
+# SPDX-FileCopyrightText: 2019 Guillermo Rodriguez
```
kamil-certat, line 1: Would you mind adding tests for the updating feature? It's a core thing of the parser now, and it's just silently tested for one case

kamil-certat (commented):
Sorry for the long waiting for the review. I have noted a few things that are important for me, mainly from the perspective of the IntelMQ user. I have marked with :red_circle: things that are the most critical for me.

Generally, I like separating the schema from IntelMQ releases. I have unfortunately doubts in some decisions as I'd like to better control when and if the update is happening. I'd also recommend taking a look on how other bots handle such topic, e.g. ASN expert: https://github.com/certtools/intelmq/tree/bcccbf501ffa8b64c2ea011d457c3409ad1dd4cd/intelmq/bots/experts/asn_lookup

I have also identified a few points, where the code can break e.g. in our environment.

In addition, I'd recommend adding the regular update to the crontab installed with native packages: https://github.com/certtools/intelmq/tree/develop/contrib/cron-jobs

If you have any questions or need any help, you can hit me on the ShadowServer MM or email (but unfortunately I go on the vacation and be back at 21.08).

kamil-certat:
In addition, I'd like to ask if you can publish a policy on changes in the schema, the most important for me is in which cases you can change classification config (e.g. identifier, taxonomy)

CHANGELOG.md
```diff
@@ -24,10 +24,13 @@ CHANGELOG
 ### Bots
 
 #### Collectors
+- `intelmq.bots.collectors.shadowserver.collector_reports_api`:
+  - The 'json' option is no longer supported as the 'csv' option provides better performance.  Please see intelmq/bots/parsers/shadowserver/README.md for a sample configuration. (PR#2372)
 
 #### Parsers
 - `intelmq.bots.parsers.shadowserver._config`:
   - Reset detected `feedname` at shutdown to re-detect the feedname on reloads (PR#2361 by @elsif2, fixes #2360).
+  - Switch to dynamic configuration to decouple report schema changes from IntelMQ releases.  Please see intelmq/bots/parsers/shadowserver/README.md for details. (PR#2372)
```
aaronkaplan: I think we should also add a section in the official intelmq docs for the shadowserver feeds, since they are so crucial.

intelmq/bots/parsers/shadowserver/README.md
aaronkaplan: Also - what happens when people don't update in air-gapped environments? Is the parsing fail-safe then ? Like HTML ignores fields and element attributes it does not understand?

intelmq/bots/parsers/shadowserver/_config.py
```diff
@@ -241,4125 +258,92 @@ def scan_exchange_identifier(field):
...
-        ('extra.', 'subject', validate_to_none),
-        ('malware.hash.md5', 'md5', validate_to_none),
-    ],
-    'constant_fields': {
-        'classification.taxonomy': 'abusive-content',
-        'classification.type': 'spam',
-        'classification.identifier': 'spam-url',
-    },
-}
+functions = {
```
aaronkaplan, line 293: I won't check this huge diff, but I assume that's valid since automatically generated stuff gets removed and moved to the dynamic confgen

intelmq/bots/parsers/shadowserver/_config.py
aaronkaplan: 4. and have a mechanism to *test* the new config automatically. If it fails, the admin needs to know. And the developers need to know.

intelmq/bots/parsers/shadowserver/README.md
sebix: if the schema file doesn't exist during bot initialization, the init() method could trigger the download

intelmq/tests/bots/parsers/shadowserver/test_report_smb.py
sebix: P.S.: `intelmq.lib.test.skip_internet`

sebix:
I haven't seen a discussion of this architectural change or and IEP yet. Or was it discussed on a non-public list?

intelmq/tests/bots/parsers/shadowserver/test_report_smb.py
elsif2: This does not require an internet connection.  Test cases use the static `schema.json.test` file.

intelmq/bots/parsers/shadowserver/README.md
aaronkaplan: agree with sebix here: this would be a good initial step which helps it to work implicitly.

intelmq/bots/parsers/shadowserver/README.md
```diff
@@ -7,3 +7,45 @@ This module is maintained by [The Shadowserver Foundation](https://www.shadowser
...
+  description: Our bot responsible for getting reports from Shadowserver
+  enabled: true
+  group: Collector
+  module: intelmq.bots.collectors.shadowserver.collector_reports_api
+  name: Shadowserver_Collector
+  parameters:
+    destination_queues:
+      _default: [shadowserver-parser-queue]
+    file_format: csv
+    api_key: "$API_KEY_received_from_the_shadowserver_foundation"
```
aaronkaplan: Is it possible to also to give a specific query (as in the shadowserver API ) here? like asn:12345?

intelmq/bots/collectors/shadowserver/collector_reports_api.py
elsif2: Updated: 
```
              raise ValueError("Invalid file_format '%s'. Must be 'csv'." % self.file_format)
```

intelmq/bots/collectors/shadowserver/collector_reports_api.py
elsif2: CSV is the native format for Shadowserver reports.

The `check` method has been added.

intelmq/bots/parsers/shadowserver/parser.py
elsif2: Fixed / replaced by _auto_update_.

intelmq/bots/parsers/shadowserver/parser.py
elsif2: Now controlled by the `auto_update` parameter.

intelmq/bots/parsers/shadowserver/README.md
elsif2: Now controlled by the _auto_update_ parameter.

intelmq/bots/parsers/shadowserver/_config.py
elsif2: Added.

intelmq/bots/parsers/shadowserver/_config.py
elsif2: Added.

intelmq/bots/parsers/shadowserver/update_schema.py
elsif2: Added.

intelmq/tests/bots/parsers/shadowserver/test_report_smb.py
elsif2: Added `test_download_schema.py`.

elsif2:
> In addition, I'd like to ask if you can publish a policy on changes in the schema, the most important for me is in which cases you can change classification config (e.g. identifier, taxonomy)

I started a *Schema contract* section to address concerns in the `bots/parsers/shadowserver/README.md`.

aaronkaplan:
> I haven't seen a discussion of this architectural change or and IEP yet. Or was it discussed on a non-public list?

@sebix I had a good chat with elsif2 today and we said 
- I try out the changes, get it to work for myself
- document things as I go
- cross check with elsif2 as I got and we change things iteratively
- we both start to write down sort of a "contract" / documentation on this . It's paramount to know for intelmq-users and -admins what the "contract" is: what fields may change dynamically, added fields will stay for ever, only new fields get added, how to deal with new fields? How often updates are needed? What happens when someone does not update? ETc. etc... If we write this down and make it clear, it is the basis for discussions and it helps also to improve the concept.

So, TL;DR: don't worry :) we are adressing concerns.

aaronkaplan (dismissed):
I share a lot of concerns with @kamil-certat here, but my main concern is: is there a way to be as little disruptive as possible? Or let me re-formulate: let's think hard how we can achieve this.

Because most of the time intelmq runs unattended, automatically, at night. 
People don't notice it in the logs if something dynamic (like a config update ) happened. 
Nor will they debug it.

We agreed in a chat today that 
- I would try out the change
- We iterate over it and improve it
- We document it
- We document a type of "contract" between users/-admins and the dynamic update service
- We need to have a opt-in/opt-out process for this.

I want to re-iterate that I think it is very very beneficial if intelmq users can easily get the latest and best shadowserver parsers. We just need to make sure it gets done right.

intelmq/bots/parsers/shadowserver/_config.py
elsif2: The behavior is unchanged from the existing release.  Errors are logged for new report types.  Fields added to existing reports are ignored.

intelmq/bots/parsers/shadowserver/README.md
elsif2: > Is it possible to also to give a specific query (as in the shadowserver API ) here? like asn:12345?

No.  Some organizations do have separate report sets that can be handled independently using the _reports_ option.

intelmq/bots/collectors/shadowserver/collector_reports_api.py
sebix: If there's only one choice left, why is it configurable at all?

elsif2: The file_format parameter has been removed.

aaronkaplan: makes sense if there is only one format :) thx

sebix:
> > I haven't seen a discussion of this architectural change or and IEP yet. Or was it discussed on a non-public list?
> 
>     * we both start to write down sort of a "contract" / documentation on this . It's paramount to know for intelmq-users and -admins what the "contract" is: what fields may change dynamically, added fields will stay for ever, only new fields get added, how to deal with new fields? How often updates are needed? What happens when someone does not update? ETc. etc... If we write this down and make it clear, it is the basis for discussions and it helps also to improve the concept.

This is a good step forward!

> So, TL;DR: don't worry :) we are adressing concerns.

I'm not worried personally, as I'm not directly affected, but many users are. I want to contribute my experience to get the change into operation smoothly.

Needless to say, the proposed approach has its advantages.

Here are some thoughts:
* The delegation of the mapping to a third party outside of the IntelMQ software also deprives the IntelMQ developers & users of the possibility of reviewing the field mappings before they become active.
* Until now, only collectors fetch the Shadowserver data via API or from local sources like IMAP, file, ticket system, etc. IntelMQ doesn't need to be connected to the internet to parse Shadowserver data. With this change merged, IntelMQ gains an additional requirement on the operation. Operators need to consider that and eventually adapt their installations or environments.
* Another issue is that it will take more work to parse older or historical data. Do you remember that a few years ago, we parsed Shadowserver data covering a few years of data for statistical purposes? We had the older Shadowserver feed mappings at hand, thus no issues parsing older data.

aaronkaplan:
Let me try to address some topics mentioned.

> * The delegation of the mapping to a third party outside of the IntelMQ software also deprives the IntelMQ developers & users of the possibility of reviewing the field mappings before they become active.

nope, since you can always fetch the newest mappings manually and review. So, I don't see a problem here. 
But yes, we could and should try out the "air-gapped" mode and see if it's actually feasible before merging. Agreed with that.

> * Until now, only collectors fetch the Shadowserver data via API or from local sources like IMAP, file, ticket system, etc. IntelMQ doesn't need to be connected to the internet to parse Shadowserver data. With this change merged, IntelMQ gains an additional requirement on the operation. Operators need to consider that and eventually adapt their installations or environments.

Again see above, it will also work in airgapped mode and pulling the latest mappings will be similar to doing an intelmq update manually.

> * Another issue is that it will take more work to parse older or historical data. Do you remember that a few years ago, we parsed Shadowserver data covering a few years of data for statistical purposes? We had the older Shadowserver feed mappings at hand, thus no issues parsing older data.

*This* is a very relevant point. I think this needs to be in the contract that historic fields don't get remapped or changed or deleted.

CHANGELOG.md
```diff
@@ -28,15 +28,21 @@ CHANGELOG
...
+<<<<<<< HEAD
 - `intelmq.bots.collector.rt`:
   - restrict `python-rt` to be below version 3.0 due to introduced breaking changes,
   - added support for `Subject NOT LIKE` queries,
   - added support for multiple values in ticket subject queries.
 - `intelmq.bots.collectors.rsync`: Support for optional private key, relative time parsing for the source path, extra rsync parameters and strict host key checking (PR#2241 by Mateo Durante).
+=======
+- `intelmq.bots.collectors.shadowserver.collector_reports_api`:
+  - The 'json' option is no longer supported as the 'csv' option provides better performance.  Please see intelmq/bots/parsers/shadowserver/README.md for a sample configuration. (PR#2372)
+>>>>>>> Documentation and style updates.
```
kamil-certat: Please fix leftovers from merge conflict

intelmq/bots/parsers/shadowserver/_config.py
```diff
@@ -241,4125 +264,109 @@ def scan_exchange_identifier(field):
...
+            return
+    else:
+        __config.logger.info("The schema file does not exist.")
+
+    if __config.schema_mtime == 0.0 and mtime == 0.0 and __config.auto_update:
+        update_schema()
+
+    __config.feedname_mapping.clear()
+    __config.filename_mapping.clear()
+    for schema_file in [__config.schema_file, __config.schema_base]:
```
kamil-certat: Could you explain a little why load `schema_base`, that looks to be a test schema, every time? If the only usage is for testing, I'd suggest loading it just during tests by specifying bot's parameters.

If this has more usage, I'd suggest loading base first. Otherwise, the base can override the updated schema.

intelmq/bots/parsers/shadowserver/_config.py
```diff
@@ -241,4125 +264,109 @@ def scan_exchange_identifier(field):
...
+                else:
+                    __config.feedname_mapping[schema[report]['feed_name']] = (schema[report]['feed_name'], schema[report])
+                    __config.filename_mapping[schema[report]['file_name']] = (schema[report]['feed_name'], schema[report])
+    __config.schema_mtime = mtime
+
+
+def update_schema():
+    """ download the latest configuration """
+    if os.environ.get('INTELMQ_SKIP_INTERNET'):
+        return False
```
kamil-certat: As being set, this is for tests only. Usage in production code is unpredictable. Instead, your test should skip executing when this env is enabled.

intelmq/bots/parsers/shadowserver/_config.py
```diff
@@ -241,4125 +264,109 @@ def scan_exchange_identifier(field):
...
-    'constant_fields': {
-        'classification.taxonomy': 'vulnerable',
-        'classification.type': 'vulnerable-system',
-        'protocol.application': 'xdmcp',
-        'classification.identifier': 'open-xdmcp',
-    },
-}
+    # download the schema to a temp file
+    (th, tmp) = tempfile.mkstemp(dir=VAR_STATE_PATH)
+    url = 'https://interchange.shadowserver.org/intelmq/v1/schema'
```
kamil-certat: Maybe export the URL as a constant / parameter?

intelmq/bots/parsers/shadowserver/_config.py
```diff
@@ -241,4125 +264,109 @@ def scan_exchange_identifier(field):
...
+    __config.logger.info("Attempting to download schema from %r" % url)
```
kamil-certat: NIT: When you use Python logger, do not format message strings. Instead, pass the template as the first argument, and values as the following arguments, in this case:

```python
__config.logger.info("Attempting to download schema from %r", url)
```

This small change is consistent with how Python logger works: if given log level isn't configured to follow, the logger will skip the message without executing formatting. So, it's nice-to-have optimalization :)

intelmq/bots/parsers/shadowserver/parser.py
```diff
@@ -209,5 +231,35 @@ def parse_line(self, row, report):
...
+                logger.setLevel('INFO')
+            else:
+                logger.setLevel('ERROR')
+            config.set_logger(logger)
+            if config.update_schema():
+                runtime_conf = utils.get_bots_settings()
+                try:
+                    ctl = IntelMQController()
+                    for bot in runtime_conf:
+                        if runtime_conf[bot]["module"] == __name__ and runtime_conf[bot]['parameters'].get('auto_update', True):
```
kamil-certat: I'm a little unsure as `auto_update` and reloading bots after update are two different things: you can manually or by cron update schema, and then reloading bots is expected. I'd suggest skipping `auto_update` check here. Or maybe your idea was different?

intelmq/bots/parsers/shadowserver/README.md
kamil-certat: I'm not sure which is better (starting schema vs. download when missing), but I can accept any of them.

kamil-certat:
Thanks for changes, @elsif2 I see most of my technical concerns have been addressed :)

About the general discussion:

re: delegation of schema outside IMQ

I see this as a trade-off problem, but I think ShadowServer has a very good reason to delegate it outside as their reports are developing much quicker than IntelMQ. In addition, I see a possibility for automated upgrades of this schema as a big facilitation for admins - even with a very quick adoption in IMQ packages, you still need to upgrade the whole IntelMQ instead of just the schema. Yes, it has some cons, but I think the pros are stronger in this case.

re: air-gaped systems

This is why I'd suggest keeping default, regularly updated (e.g. once a month) schema in the repo. Then your air-gaped system can still work after installation, just without additional configuration it would have a delay in adoption of schema changes.

re: historical schema

This is a good point. But I'd just suggest keeping the history of schemas, in to ways:
* do not delete old schema, create a `history` folder instead and store them all downloaded schemas, named after the schema version.
* For cases when you don't have the schema, @elsif2 would it be possible, that ShadowServer publish a public catalog with all historical schemas?

As I understand, the schema is versioned by time, so it should be easy to choose the right one if you need to perform a reparsing of old reports - assuming that ShadowServer would ensure that date of creation a schema means, that starting this or the next day the schema is used (please clarify it in the contract)

kamil-certat (changes requested):

elsif2:
The complete revision history of the schema files will be available at https://github.com/The-Shadowserver-Foundation/report_schema.

CHANGELOG.md
elsif2: Fixed.

intelmq/bots/parsers/shadowserver/_config.py
elsif2: Added a test_mode to switch between schema files.

intelmq/bots/parsers/shadowserver/_config.py
elsif2: Removed.

intelmq/bots/parsers/shadowserver/_config.py
elsif2: Added as a constant.

intelmq/bots/parsers/shadowserver/_config.py
elsif2: The logger parameters have been updated.

intelmq/bots/parsers/shadowserver/parser.py
elsif2: Removed auto_update check.

CHANGELOG.md
sebix: The README is not the right place to document the bot for the users.
user documentation is in `docs/user/bots.rst` (https://intelmq.readthedocs.io/en/latest/user/bots.html#shadowserver) and `intelmq/etc/feeds.yml` (https://intelmq.readthedocs.io/en/latest/user/feeds.html). For upgrade steps, use `NEWS.md`.
Please move the information from the README there.

CHANGELOG.md
```diff
@@ -62,15 +62,19 @@ CHANGELOG
 ### Bots
 
 #### Collectors
+<<<<<<< HEAD
```
sebix: merge leftover

intelmq/bots/parsers/shadowserver/README.md
```diff
@@ -7,3 +7,60 @@ This module is maintained by [The Shadowserver Foundation](https://www.shadowser
 
 Please contact intelmq@shadowserver.org with any issues or concerns.
 
+The report configuration is now stored in a _shadowserver-schema.json_ file downloaded from https://interchange.shadowserver.org/intelmq/v1/schema.
```
sebix: As explained above, the content here should be moved to the proper files

intelmq/bots/parsers/shadowserver/README.md
```diff
@@ -7,3 +7,60 @@ This module is maintained by [The Shadowserver Foundation](https://www.shadowser
...
+For air-gapped systems automation will be required to download and copy the file to VAR_STATE_PATH/shadowserver-schema.json.
+
+The parser will automatically reload the configuration when the file changes.
+
+
+## Schema contract
+
+Once set the `classification.identifier`, `classification.taxonomy`, and `classification.type` fields will remain static.
+
+Once set report fields will not be deleted.
```
sebix: I don't understand these sentences. Do you mean
> Once set (by the user?), the `classification.identifier`, `classification.taxonomy`, and `classification.type` fields will remain static.

or something else? Better be more verbose

intelmq/tests/bots/parsers/shadowserver/test_download_schema.py
```diff
@@ -0,0 +1,29 @@
...
+class TestShadowserverSchemaDownload(unittest.TestCase):
+
+    def test_download(self):
+        if os.path.isdir(VAR_STATE_PATH):
+            schema_file = os.path.join(VAR_STATE_PATH, 'shadowserver-schema.json')
+            config.set_logger(utils.log('test-bot', log_path=None))
+            if os.path.exists(schema_file):
+                os.unlink(schema_file)
+            self.assertEqual(True, config.update_schema())
+            self.assertEqual(True, os.path.exists(schema_file))
```
sebix: Tests must not make any changes in the directories of IntelMQ instances. Use a [temporary directory](https://docs.python.org/3/library/tempfile.html?highlight=tempdir#tempfile.tempdir) instead

sebix (changes requested):
I only reviewed some parts of the PR as @kamil-certat and @aaronkaplan already reviewed most parts.

CHANGELOG.md
elsif2: I moved the documentation from the readme to the bots documentation and removed the merge leftover.

intelmq/bots/parsers/shadowserver/README.md
elsif2: Updated in the documentation.  We won't change the classifications which would break event filtering for users.

intelmq/tests/bots/parsers/shadowserver/test_download_schema.py
elsif2: The test now uses a temporary directory.

intelmq/bots/parsers/shadowserver/README.md
elsif2: Moved to documentation.

intelmq/tests/bots/parsers/shadowserver/test_download_schema.py
elsif2: Need help here.  The test includes the `@test.skip_internet()` decorator yet the automated build fails to download.  Previously I used the `os.environ.get('INTELMQ_SKIP_INTERNET')` check to handle this.

sebix: The CI does not have `INTELMQ_SKIP_INTERNET` set because it has internet access, so there's no reason to skip these tests in the CI. In other environments, internet access is not available (and thus, the result is reproducible).

elsif2: Is the CI able to create a temporary directory?

sebix: Yes, other tests do that too

elsif2: It looks like the CI environment isn't complete:

`DEBUG    test-bot:_config.py:344 File not found: '/opt/intelmq/etc/runtime.yaml'.`

sebix: The unittest action calls the unittests, it does not install IntelMQ. The unittests work all without an installed IntelMQ. Otherwise, they wouldn't be unittests, but integration tests or something else. What is the issue your are running into?

elsif2: The code uses the create_request_session() function from intelmq.lib.utils to support sites that use proxies.  The create_request_session function calls load_configuration which throws the exception.

sebix: The bot may call this function. The unit test itself must not use it, except it is mocked properly (The methods to run bots in `intelmq.lib.test.TestCase` handles this). And you are missing this mocking.

What you could try is:
- create an instance of the bot with `prepare_bot` (equivalent to `run_bot`, but missing the call to `bot.proccess`)
- use the same mocking as `run_bot` to call the method you want to call:
  ```python
        with mock.patch('intelmq.lib.utils.load_configuration',
                        new=self.mocked_config):
            with mock.patch('intelmq.lib.utils.log', self.get_mocked_logger(self.logger)):
                # insert the call to the update method here
                self.bot.stop(exitcode=0)
```

Comment (elsif2) [reply intelmq/tests/bots/parsers/shadowserver/test_download_schema.py]: Thanks! I updated the test with mocking.  All checks passed.

File: intelmq/bots/parsers/shadowserver/parser.py [hunk 30]
```diff
@@ -265,5 +266,10 @@ def run(cls, parsed_args=None):
         else:
             super().run(parsed_args=parsed_args)
 
+    def test_update_schema(cls):
```
sebix, line 272: Please add a docstring explaining why this test method is needed in the bot's code.

elsif2: Added.

intelmq/tests/bots/parsers/shadowserver/test_download_schema.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def test_download(self):
+        self.prepare_bot(prepare_source_queue=False, parameters={'test_mode': True})
+        result = False
+        with mock.patch('intelmq.lib.utils.load_configuration', new=self.mocked_config):
+            with mock.patch('intelmq.lib.utils.log', self.get_mocked_logger(self.logger)):
+                self.log_stream.truncate(0)
+                result = self.bot.test_update_schema()
+                self.bot.stop(exitcode=0)
+        print(self.log_stream.getvalue())
```
kamil-certat: I think it was just for debugging, right? I'd remove this print.

intelmq/tests/bots/parsers/shadowserver/test_download_schema.py
```diff
@@ -0,0 +1,36 @@
...
+        self.assertEqual(True, result)
```
kamil-certat, line 34: I'd suggest adding some basic checks of the downloaded schema, like:

 * is it saved with the correct filename?
 * is it a correct JSON?
 * does it have the meta/version field?

So, basically I'd load the downloaded schema and checked one field. This would confirm, that this bot downloads the correct file.

kamil-certat (approved):
Hey, I generally feel satisfied with the current form :) For me, it's okay to merge it and test (after addressing any leftovers). Thanks for the work, and incorporating my comments.

To the schema contract, I'd ask to clarify if the _content_, _type_, or _mapping to IDF_ of fields can be changed later (and how we would be informed about that). Could you also specify, how we can be informed about new fields in reports and new report types? I assume that the primary canal is the Mattermost, isn't it? As discussed here, it would be nice to get some explanation how to select the right schema for a report (when parsing older data).

intelmq/tests/bots/parsers/shadowserver/test_download_schema.py
elsif2: For a test case I believe it to be helpful to provide output.

intelmq/tests/bots/parsers/shadowserver/test_download_schema.py
elsif2: The update_schema function in _config.py verifies that the downloaded file is valid JSON and contains a version identifier.  The filename is set by the _config.py module.

intelmq/tests/bots/parsers/shadowserver/test_download_schema.py
kamil-certat: I agree, but the logger output is collected by the test runner and automatically printed out when test fails, see e.g. output from pytest:

grafik

Thanks to this you don't need to print the log manually – and the output isn't making garbage on the screen if everything is okay. If you rely on manual reading the output to ensure it passed, then use our helpers like `assertLoglineMatches` or `assertLogMatches` to test the log output.

intelmq/tests/bots/parsers/shadowserver/test_download_schema.py
kamil-certat: Ah, right. I prefer to test such things in test code explicitly for readability and stability of tests, but it's opinionated, so I do not insist.

intelmq/tests/bots/parsers/shadowserver/test_download_schema.py
elsif2: Removed in 2f11b2a6667393c25eacd2584d14ade09065eb15.

sebix:
> I assume that the primary canal is the Mattermost, isn't it?

Is that a Mattermost channel that all Shadowserver partners know and regularly monitor? In the past, e-mail announcements were sent, notifying of new reports.

elsif2:
Report announcements are normally sent to the public@shadowserver.org mailing list as well as appearing on the website (https://www.shadowserver.org/news-insights/).  Updates may also be posted on the Mattermost channel, but that is only accessible by Alliance members.

kamil-certat:
So, the report fields can be removed? If so, please clarify an appropriate adoption window & warning before :)

sebix (approved):

sebix:
In-code issues that I have found are addressed, so I (formally) added my approval. Please note, that I limited my review to only shallow aspects of documentation, code architecture, and did not go into detail at all, as I feel unqualified not to be a current user of this bot.

It's up to @kamil-certat or @aaronkaplan to get this PR reviewed/merged.

aaronkaplan:
Will do. As always, will do an end-to-end test first and check with @elsif2 in case I see issues.
Also, @elsif2 and me would like to write a "contract" document (initial text for that already in the PR) which allows people to know how to use and when to use the dynamic config safely.

elsif2:
> So, the report fields can be removed? If so, please clarify an appropriate adoption window & warning before :)

Removing fields is a rare occurrence, more often they will be renamed to make the field names standardized across reports.  There will be at least 30 days notice on any field changes.

kamil-certat:
Great, thanks for the clarification!

kamil-certat:
@aaronkaplan @elsif2 Hey guys, how the work is going? 

I think I found a one edge-case that would be nice to handle: I'm going to propose new fields in the IntelMQ Data Format. I think the dynamic configured parser may have problems, if a new schema decides to use some field added in a new IDF, which is not yet in the local harmonization file. This would limit ShadowServer not to use any new field - maybe you want (not necessary in this PR, maybe later or so) to cover this case? E.g. have a fallback field for potentially problematic fields, catch the harmonization error and try to use the fallback field if the first-choice isn't available?

aaronkaplan:
Not sure if I fully understand your point @kamil-certat . Are you saying that if you introduce a field name, then it might be reserved for shadowserver dynamic config changes? I think that's something we could solve with a procedure at shadowserver before a new field gets added there on that side. No?

sebix:
> Not sure if I fully understand your point @kamil-certat . Are you saying that if you introduce a field name, then it might be reserved for shadowserver dynamic config changes? I think that's something we could solve with a procedure at shadowserver before a new field gets added there on that side. No?

The other way round: If IntelMQ IDH introduces a new field, and the shadowserver dynamic config starts using it, all shadowserver bots on IntelMQ instances that are on an older version (that does not have the new field) will throw errors.

aaronkaplan:
Good point, will circle this with shadowserver

aaronkaplan:
So , @elsif2 and me discussed and we propose this approach:

The example was: lets say a `severity` field is added to the IDF.  If an exception is thrown on that field, I would just set it as `extra.severity` instead?

Resulting solution propos:
for a dynamically updated shaodwserver config/bot, iff the bot does not find a field (in an old installation) ,it will
1. log that it did not find the field
2. ignore it

This approach has the benefit that it keeps the assumption that the field does not exist in this (old) intelmq installation.

Another option could be to move writes to a field `X` which does not exist in an old version to `extra.X`. This also has some downsides for bots downstream which then have to look for `X` as well as for `extra.X`.

Please comment @kamil-certat , @sebix

elsif2:
I submitted a small update that logs a warning for any field that is not in the IDF.

kamil-certat:
@aaronkaplan The proposed behaviour sounds okay for me

kamil-certat (commented):
I think it looks good now. I would try to test it in our test instance in the next two weeks.

Could you please take a look at the commits? It looks like Github and/or git failed to properly discover the diff. Could you check if you shouldn't merge or rebase your branch to prevent mess on the `develop`?

kamil-certat (approved):

aaronkaplan:
Yup, @kamil-certat if it works for your setup I think we should merge (finally!!!!) :)

kamil-certat:
Okay, I'll let you know when I test it

kamil-certat (approved):

kamil-certat:
Notes from testing on our instance:

1. Please let configure the schema URL. As the ShadowServer infrastructure is currently down, I needed to switch the configuration to use the schema from Github. The only way to do is currently editing the source code.
   * Suggestion: Set it as a parser config value, and then schema loader should use the first url found in active parsers' runtime configuration. The [`Bot.check()`](https://github.com/certtools/intelmq/blob/d624988693894e71defdd8b18618f7af98be852c/intelmq/lib/bot.py#L963) method should report if different bots have different schema URLs, unless you want to support such a case.
3. As the parser do not provide any initial schema and auto updates are disabled by default, the first run on a new installation/update from previous parser has no chance to work. But nothing says about that, because `_config.reload()` does nothing if the schema file does not exist.
   * Suggestion: `_config.reload()` should fail if schema file does not exist, and suggest downloading it or enabling automated downloads. In addition, the `Bot.check()` method should report error if neither the schema file exists nor auto updates are enabled.
2. The proposed cron scheduler should ensure the user, or mention it's important to run it as `intelmq`. Otherwise, the schema is created with e.g. root as owner, and bot cannot access it (tested).
   * Suggestion: add `intelmq` between the schedule and command in the crontab line in `bots.rst`. In addition, as mentioned previously, I suggest adding this line to the IntelMQ crontab for deb packages: https://github.com/certtools/intelmq/blob/develop/debian/cron.d/intelmq-update-database

But otherwise - it looks working :white_check_mark: It processed well sample data I have on the staging. I'd suggest doing those small changes and merging the code (cc: @aaronkaplan @sebix)

elsif2:
1. The interchange URL should be the primary URL.  We are working on obtaining redundancy.

2. I changed the error from a log message to an exception.

3. Updated bots.rst that the cronjob should be run as `intelmq`.

kamil-certat:
Thank you!

aaronkaplan:
looks like my requests have been met. 
Testing again now.

docs/user/bots.md
```diff
@@ -956,7 +951,27 @@ The resulting reports contain the following special field:
...
+  shadowserver-collector:
+    description: Our bot responsible for getting reports from Shadowserver
+    enabled: true
+    group: Collector
+    module: intelmq.bots.collectors.shadowserver.collector_reports_api
+    name: Shadowserver_Collector
+    parameters:
+      destination_queues:
+        _default: [shadowserver-parser-queue]
+      file_format: csv
```
aaronkaplan, line 969: by now, we get a warning when running `intelmqctl check`:

```
Bot 'shadowserver-collector': The file_format parameter is no longer supported. All reports are CSV.
```

sebix:
I wanted to cancel the pending review for myself requested by Aaron, but it looks like I instead canceled one of Aaron's reviews :thinking: 

Anyway, I won't review this PR again, as I already reviewed some parts of it (core, docs). The major and more relevant parts have already been reviewed by Kamil and Aaron. I don't want to duplicate their efforts. I shared my thoughts on the change in general already above (https://github.com/certtools/intelmq/pull/2372#issuecomment-1662214124)

aaronkaplan:
@sebix all good, I am still looking at it. It's a mega mega mega commit. Need to do a smaller one the next time to make the review process faster. Most of us have other duties / daytime jobs etc.

kamil-certat:
This branch is currently in production usage on our instance, so far I do not see any issues.

DigitalTrustCenter:
We would really like to see this feature merged in IntelMQ. We know the pain of manually merging the _config.py whenever new reports are added too well. The changes made to host(name) on the 18th of November showed the problem where we would require a different config from one day to another. The dynamic config would fix those kinds of issues as well.

We do however see a problem. We run a custom config with a couple of changes compared to the upstream config:
-	We changed the classification.identifer of scan_http_vulnerable from accessible-http to vulnerable-http. accessible-http is also used by scan_http, which is a completely different report.
-	event_honeypot_brute_force, event_honeypot_darknet, event_sinkhole_http and event_sinkhole map fields like tag to classification.identifer. We would like a constant value for classification.identifer, so we moved it to the constant_fields section with values honeypot-brute-force, darknet, sinkhole-http and sinkhole.
-	sandbox_url and sandbox_conn (unnecessary in our opinion) mix source and destination fields. We changed destination.fqdn and destination.to source.fqdn and source.url.

We would not be surprised if other organizations have their own changes that they made to the config.
A solution would be to allow us to configure a path to a JSON patch document (RFC 6902) in the config. This patch is applied to the schema JSON after a new version is downloaded in the update process.

kamil-certat:
Hey, we indeed have a couple of custom changes (e.g. filters by tags, own identifiers etc.), but we use a sieve bot for that - and as long as the `feed.code` is stable, our bot handles that independently. I would recommend this approach instead of modifying ´_config.py` or patching schema

DigitalTrustCenter:
Thanks for the pointer. We are now looking into improving our setup using the sieve bot. The only thing it cannot solve is the change of the fields for sandbox_url (you cannot add a new field with a value from another field). Maybe those fields should be changed in the upstream schema itself. However, that is not relevant for this PR.

elsif2:
The change occurred between 3.0.2 and 3.1.0.  I will correct this in the schema.

## 3.0.2

```
sandbox_url = {
    'required_fields': [
        ('time.source', 'timestamp', add_UTC_to_timestamp),
        ('source.ip', 'ip'),
    ],
    'optional_fields': [
        ('source.asn', 'asn', invalidate_zero),
        ('malware.hash.md5', 'md5hash'),
        ('source.url', 'url'),
        ('user_agent', 'user_agent', validate_to_none),
        ('source.fqdn', 'host', validate_fqdn),
        ('extra.', 'method', validate_to_none),
    ],
    'constant_fields': {
        'classification.taxonomy': 'malicious-code',
        'classification.type': 'malware-distribution',
        'classification.identifier': 'sandbox-url',
    },
}
```

## 3.1.0

```
sandbox_url = {
    'required_fields': [
        ('time.source', 'timestamp', add_UTC_to_timestamp),
        ('source.ip', 'ip', validate_ip),
    ],
    'optional_fields': [
        ('destination.fqdn', 'host', validate_fqdn),
        ('extra.http_request_method', 'method', validate_to_none),
        ('source.asn', 'asn', invalidate_zero),
        ('malware.hash.md5', 'md5', validate_to_none),
        ('destination.url', 'url', convert_http_host_and_url, True),
        ('user_agent', 'user_agent', validate_to_none),
    ],
    'constant_fields': {
        'classification.taxonomy': 'malicious-code',
        'classification.type': 'malware-distribution',
        'classification.identifier': 'sandbox-url',
    },
}

sandbox_conn = {
    'required_fields': [
        ('time.source', 'timestamp', add_UTC_to_timestamp),
        ('source.ip', 'ip', validate_ip),
        ('source.port', 'port', convert_int),
    ],
    'optional_fields': [
        ('destination.fqdn', 'host', validate_fqdn),
        ('source.asn', 'asn', invalidate_zero),
        ('malware.hash.md5', 'md5', validate_to_none),
        ('protocol.transport', 'protocol'),
        ('extra.', 'bytes_in', validate_to_none),
        ('extra.', 'bytes_out', validate_to_none),
    ],
    'constant_fields': {
        'classification.taxonomy': 'malicious-code',
        'classification.type': 'malware-distribution',
        'classification.identifier': 'sandbox-conn',
    },
}
```

sebix:
> Thanks for the pointer. We are now looking into improving our setup using the sieve bot. The only thing it cannot solve is the change of the fields for sandbox_url (you cannot add a new field with a value from another field).

That is true. As a workaround, you could use the [Jinja2 Template Expert](https://intelmq.readthedocs.io/en/latest/user/bots.html#jinja2-template-expert) or the [Modify Expert](https://intelmq.readthedocs.io/en/latest/user/bots.html#modify)

aaronkaplan (approved):
This has been possibly the longest running PR in intelmq.
But it's also the most complex one I have ever seen.
I think @elsif2 managed to stun us all.
Checking and testing this rigorously is pretty much out of the question. 
But what I tested and was running on my setup, it looks good.

Thanks *a ton* for this important PR and sorry that it took months. Godspeed, shadowserver!

docs/user/bots.md
aaronkaplan: it's a warning.

aaronkaplan:
And all checks have passed :) merging in...

CHANGELOG.md
```diff
@@ -156,11 +156,13 @@
...
   - added support for multiple values in ticket subject queries.
 - `intelmq.bots.collectors.rsync`: Support for optional private key, relative time parsing for the source path, extra rsync parameters and strict host key checking (PR#2241 by Mateo Durante).
+- `intelmq.bots.collectors.shadowserver.collector_reports_api`:
+  - The 'json' option is no longer supported as the 'csv' option provides better performance.
 
 #### Parsers
 - `intelmq.bots.parsers.shadowserver._config`:
   - Reset detected `feedname` at shutdown to re-detect the feedname on reloads (PR#2361 by @elsif2, fixes #2360).
-- `intelmq.bots.parsers.shadowserver._config`:
+  - Switch to dynamic configuration to decouple report schema changes from IntelMQ releases.
```
sebix, line 165: The changelog entry is in the wrong section (3.2.0) instead of 3.2.2

sebix: can be fixed with https://github.com/certtools/intelmq/pull/2447

Issue #1523, opened 2020-05-06
Relative time parsing in Filter Expert Bot


kbrajneesh:
For relative filter, if param `not_after: x hours`  (**x < 24**) is given, instead of being parsed as **datetime.timedelta** object by
>             relative = timedelta(minutes=parse_relative(time_attr)) [Line 22]
currently it is being parsed as **datetime.datetime** object by
>             absolute = parser.parse(time_attr) [Line 20]
Leading to error
**TypeError: can't compare offset-naive and offset-aware datetimes**

wagner-certat:
Thanks for reporting!

PR #2374, opened 2023-06-09
sieve bot gets :before and :after keywords

e3rd:
We face an interesting issue. One of our sources takes a lot of time to investigate a webpage before reporting them sometimes. Thus, its get fresh `extra.verification_time` but very old `time.source`. Filtering with the expert filter is only available for the `time.source`. Hence, up to 7 % of the webpages from this source did not make it through our `time.source` expert filter, despite they are completely new for us.

The hard but robust solution was to update the sieve bot so that it could handle dates.

Instead of `dateutil.parser` I used `pendulum` library which is just great. No more infinite messing with the timezones, better parsing dates ( https://github.com/certtools/intelmq/issues/1523 will not occur here).


sebix (commented):
Nice feature!

What is the major benefit of `pendulum` over the existing `intelmq.lib.utils.parse_relative` function?

e3rd:
`pendulum.parse` replaces here external `dateutil.parser.parse`, not the great internal `parse_relative` :)

sebix:
What does `pendulum.parse` do better than `dateutil.parser.parse`? The timestamp data format in events is already harmonized at this processing stage (after the parser, in the expert), therefore `intelmq.lib.utils.DateTime.from_isoformat` should suffice to parse the timestamp and is _way_ faster.

e3rd:
Oh, is it? Nice for the event. I use it for the string from the `.sieve` file too. I was a bit unhappy about trimming timezones when using dateutil `.replace(tzinfo=timezone.utc)` used in the `filter.expert` – or is that fold proof? Timezones danced ahead my eyes... 

Also, dateutil enforces to use `TIMESPANS` in order to mitigate #1523. Pendulum does not need that, hence the code is a little shorter.

But I did not think about performance, if you say pendulum is much slower than dateutil, I'll refactor that back next week.

sebix:
Sorry for being unprecise. I didn't mean that dateutil is faster than pendulum (I didn't check that) but both are slower than `intelmq.lib.utils.DateTime.from_isoformat`.

e3rd:
Hi, you are right, `DateTime.from_isoformat` is the right way to parse a key since intelmq message is harmonized. (I did this originally because my model example intelmq/bots/experts/filter/expert.py#L81 does not use DateTime.from_isoformat neither.)

Now I used DateTime.from_isoformat to parse the intelmq-message value.

I did some naive measuring pendulum vs dateutil.parser and pendulum is 2× – 8× quicker in every example I made up.

```python
from pendulum import parse as pd
from dateutil.parser import parse as dt

%timeit -n 100 dt("2016-05-15")
%timeit -n 100 pd("2016-05-15")
%timeit -n 100 dt("2016-05-15 12:12")
%timeit -n 100 pd("2016-05-15 12:12")
%timeit -n 100 dt("12:12")
%timeit -n 100 pd("12:12")
%timeit -n 100 dt("2015-09-12T06:22:11+00:00")
%timeit -n 100 pd("2015-09-12T06:22:11+00:00")

13.1 µs ± 1.96 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
2.59 µs ± 121 ns per loop (mean ± std. dev. of 7 runs, 100 loops each)
20.3 µs ± 1.42 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
2.68 µs ± 504 ns per loop (mean ± std. dev. of 7 runs, 100 loops each)
8.81 µs ± 221 ns per loop (mean ± std. dev. of 7 runs, 100 loops each)
4.19 µs ± 87.5 ns per loop (mean ± std. dev. of 7 runs, 100 loops each)
26.8 µs ± 266 ns per loop (mean ± std. dev. of 7 runs, 100 loops each)
3.07 µs ± 160 ns per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

However, it always takes around 4 µs to parse the `:before`/`:after` keyword time record with pendulum, for every processed line. If you think this is too much, I'll implement a caching or something.

sebix:
The speed advantage of pendulum comes from CPython, which neither dateutil nor Python's `datetime.strptime` uses. Even strptime is 50% slower than pendulum, but we are discussing only microseconds. We probably lose more time in other places, but every microsecond counts with large input data (parsing).

pendulum's speed advantage makes it attractive, and we can address this topic on a broader basis as a candidate to replace dateutil IntelMQ-wide instead of switching just one bot to another parser, making the overall user experience less consistent. E.g. providing packages and requiring it by default in `setup.py` would install it automatically.

intelmq/tests/bots/experts/sieve/test_sieve_files/test_date_match.sieve
```diff
@@ -0,0 +1,28 @@
...
+	append extra.list 'before 2023-06-01 15:00'
+}
+
+if time.observation :after '2023-06-01 15:00' {
+	append extra.list 'after 2023-06-01 15:00'
+}
+
+if extra.str == 'few-hours' && time.observation :after '10 h' {
+	append extra.list 'after 10 h'
+}
```
sebix: Missing newline

sebix (changes requested):
The speed advantage of pendulum comes from CPython, which neither dateutil nor Python's `datetime.strptime` uses. Even strptime is 50% slower than pendulum, but we are discussing only microseconds. We probably lose more time in other places, but every microsecond counts with large input data (parsing).

pendulum's speed advantage makes it attractive, and we can address this topic on a broader basis as a candidate to replace dateutil IntelMQ-wide. E.g. providing packages and requiring it by default in `setup.py` would install it automatically.

intelmq/tests/bots/experts/sieve/test_sieve_files/test_date_match.sieve
e3rd: oki

sebix (approved):

sebix:
@aaronkaplan wants to review this PR too, so waiting for him

gethvi (approved):

PR #2375, opened 2023-06-13
FIX: Updates Turris Greylist parser and feed URL

### LINKED ISSUE #2167: Turris greylist has been updated
It seems that this morning the Turris graylist has been updated. The CSV file location has been changed from `https://project.turris.cz/greylist-data/greylist-latest.csv` to `https://view.sentinel.turris.cz/greylist-data/greylist-latest.csv`.

`http_url` parameter in `bots.collectors.http.collector_http` bot should be changed to `https://view.sentinel.turris.cz/greylist-data/greylist-latest.csv` to receive the correct input for the parser bot.

They also removed IP lookup data from the CSV file. (CSV header: `Address,Country,Tags,ASN` -> `Address,Tags`)
So you would also need to modify the `bots.parsers.turris.parser` file and remove `source.geolocation.cc` and `source.asn` from the `columns` list.

gethvi:
Fixes #2167.

### DIFF
--- a/intelmq/lib/upgrades.py
+++ b/intelmq/lib/upgrades.py
@@ -38,6 +38,7 @@ __all__ = ['v100_dev7_modify_syntax',
            'v301_deprecations',
            'v310_feed_changes',
            'v310_shadowserver_feednames',
+           'v320_update_turris_greylist_url',
            ]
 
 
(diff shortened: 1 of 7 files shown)


sebix:
```
____________________ TestUpgradeLib.test_all_functions_used ____________________

self = <intelmq.tests.lib.test_upgrades.TestUpgradeLib testMethod=test_all_functions_used>

    def test_all_functions_used(self):
        self.assertEqual(len(self.mapping_list_name),
                         len(set(self.mapping_list_name)),
                         msg='Some function is assigned to multiple versions.')
        self.assertEqual(set(self.allfs),
                         set(self.mapping_list_name),
>                        msg='v* functions in the module do not '
                             'match the mapping.')
E       AssertionError: Items in the second set but not the first:
E       'v320_update_turris_greylist_url' : v* functions in the module do not match the mapping.

intelmq/tests/lib/test_upgrades.py:579: AssertionError
```
The function is not listed in `__all__`

intelmq/lib/upgrades.py
```diff
@@ -38,6 +38,7 @@
            'v301_deprecations',
            'v310_feed_changes',
            'v310_shadowserver_feednames',
+           'v320_update_turris_greylist_url'
```
sebix:
```suggestion
           'v320_update_turris_greylist_url',
```
sorry for bothering again

sebix (changes requested):

intelmq/lib/upgrades.py
gethvi: No problem. :)

PR #2376, opened 2023-06-14
ENH: RTCollector: Exclude by subject and multiple values

kamil-certat:
Added support for filtering out tickets based on
given subject pattern. In addition, the subject
patterns can now contains multiple value. Because
this isn't currently supported by the python-rt library,
the query is converted to the RT-native format
if needed.


sebix (approved):

kamil-certat:
@sebix Do you mind merging? :)

PR #2378, opened 2023-06-14
FIX: Ensure rejecting URLs starting with space

### LINKED ISSUE #2377: Regression due to Python change in urllib
Python [security update](https://docs.python.org/release/3.8.17/whatsnew/changelog.html) changed how the `urlib.parse` treats the leading spaces in URLs. They are currently stripped, and the URL is successfully processed: https://github.com/python/cpython/pull/102508/files#

This breaks our test cases if running on the system with patched Python, e.g. https://github.com/certtools/intelmq/actions/runs/5266316737/jobs/9520090889

Broken test's assert:

https://github.com/certtools/intelmq/blob/6d4d8894d296159b9a14c8e0d0cd25fd9ad9e70e/intelmq/tests/lib/test_harmonization.py#L414

In addition, the official documentation now clearly says, that `parse` is not intended to be used as validation.

I think that we should keep the previous validation by rejecting URLs starting from empty chars as it may be important for some services.

kamil-certat:
The gh-102153 change in CPython modified how
urllib.pare handles URLs with the leading
spaces. To ensure previous behaviour, additional
check was added.

Closes: #2377

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,9 @@ CHANGELOG
 - `intelmq.lib.upgrages`: Fix a bug in the upgrade function for version 3.1.0 which caused an exception if a generic csv parser instance had no parameter `type` (PR#2319 by Filip Pokorný).
 - `intelmq.lib.datatypes`: Adds `TimeFormat` class to be used for the `time_format` bot parameter (PR#2329 by Filip Pokorný).
 - `intelmq.lib.exceptions`: Fixes a bug in `InvalidArgument` exception (PR#2329 by Filip Pokorný).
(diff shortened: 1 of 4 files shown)


sebix (approved):

kamil-certat:
@sebix Could you take a look, if after my last changes you still approve the PR?

CHANGELOG.md
```diff
@@ -39,6 +42,7 @@ CHANGELOG
 - `intelmq.bots.parsers.html_table.parser`: Changes `time_format` parameter to use new `TimeFormat` class (PR#2329 by Filip Pokorný).
 - `intelmq.bots.parsers.turris.parser.py` Updated to the latest data format (issue #2167). (PR#2373 by Filip Pokorný).
 
+
```
sebix:
```suggestion
```

sebix (changes requested):

PR #2379, opened 2023-06-21
FIX: Pin codespell version

kamil-certat:
Without pinning the version, codespell is regulary
failing after releasing their new version with
new dictionary.


sebix (approved):

kamil-certat:
Ugh, it won't pass unittests here. I'll cherry-pick changes to the PR with urllib fix

PR #2385, opened 2023-07-13
Added parameter for outputfile in initdb script

sebkuf:
Now you can define where to save the outputfile of the intelmq_psql_initdb script.

### DIFF
--- a/intelmq/bin/intelmq_psql_initdb.py
+++ b/intelmq/bin/intelmq_psql_initdb.py
@@ -14,6 +14,7 @@ import json
 import os
 import sys
 import tempfile
+import argparse
 
 from intelmq import HARMONIZATION_CONF_FILE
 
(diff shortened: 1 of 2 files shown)


intelmq/bin/intelmq_psql_initdb.py
```diff
@@ -80,7 +95,19 @@ def generate(harmonization_file=HARMONIZATION_CONF_FILE):
 
 
 def main():
-    OUTPUTFILE = "/tmp/initdb.sql"
+    parser = argparse.ArgumentParser(
+        prog=APPNAME,
+        description=DESCRIPTION,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        usage=USAGE
```
kamil-certat: You don't need to specify usage manually: 
https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser

> [usage](https://docs.python.org/3/library/argparse.html#usage) - The string describing the program usage (default: generated from arguments added to parser)

intelmq/bin/intelmq_psql_initdb.py
```diff
@@ -80,7 +95,19 @@ def generate(harmonization_file=HARMONIZATION_CONF_FILE):
 
 
 def main():
-    OUTPUTFILE = "/tmp/initdb.sql"
+    parser = argparse.ArgumentParser(
+        prog=APPNAME,
```
kamil-certat: You don't need to specify usage manually:
https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser

> [prog](https://docs.python.org/3/library/argparse.html#prog) - The name of the program (default: os.path.basename(sys.argv[0]))

kamil-certat (commented):
It's better to leave default generated help messages – then noone needs to manually adjust them when adding new command or changing filename.

kamil-certat (approved):

PR #2387, opened 2023-07-18
Increased the performance of the intelmqctl script

sebkuf:
It should improve the performance of the check method of the intelmqctl script. 
The global settings were read from a file with each loop pass, now the settings are read only once and then only copied.

### DIFF
--- a/intelmq/bin/intelmqctl.py
+++ b/intelmq/bin/intelmqctl.py
@@ -4,6 +4,7 @@
 
 # -*- coding: utf-8 -*-
 import argparse
+import copy
 import datetime
 import importlib
 import json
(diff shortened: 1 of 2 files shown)


intelmq/bin/intelmqctl.py
```diff
@@ -925,6 +926,7 @@ def check(self, no_connections=False, check_executables=True):
                                  "intelmqctl upgrade-config.")
 
         check_logger.info('Checking for bots.')
+        global_settings = files[RUNTIME_CONF_FILE]
```
kamil-certat:
```suggestion
        global_settings = files[RUNTIME_CONF_FILE]['global']
```
Only the `global` key are global settings, not the whole file ;)

kamil-certat (approved):

sebix (approved):

Issue #2350, opened 2023-04-06
Changing IntelMQ to Namespace Package


gethvi:
After some discussion and testing, the following steps are necessary to turn IntelMQ package into a [namespace package](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/). This is supported since Python 3.3.

This is for allowing other third party (namespace) packages (such as the discussed `intelmq-shadowserver`) to tap into `intelmq` namespace and provide bots.

1. Change the line of IntelMQ setup.py  `packages=find_packages()` to `packages=find_namespace_packages()`
2. Remove following `__init__.py` files:
    - `intelmq/__init__.py`
    - `intelmq/bots/__init__.py`
    - `intelmq/bots/*/__init__.py`
    - `intelmq/bots/*/*/__init__.py`

Basically any subpackage that is possibly sourced from multiple dirs needs to have it's `__init__.py` removed.

This is how the experimental namespace package `intelmq` dir tree looks (note the removed `__init__.py` files, other files are omitted):
```
intelmq
├── intelmq
│   └── bots
│       └── parsers
│           └── shadowserver
│               ├── _config.py
│               ├── parser_json.py
│               ├── parser.py
│               └── README.md
└── setup.py
```

This is how the dummy namespace package `intelmq-shadowserver` dir tree looks like (no `__init__.py` files either):
```
intelmq-shadowserver
├── intelmq
│   └── bots
│       └── parsers
│           └── shadowserver
│               └── external_parser.py
└── setup.py
```

Having both installed it is possible to do:
```python
from intelmq.bots.parsers.shadwoserver import parser, external_parser
```

This feature is still under consideration, however I wanted to note down my initial experiment and what would need to be done.

sebix:
As I said in another communication channel: Thank you for investigating this topic, it was on the wishlist for many years. Previously the blocker were deficiencies in Python packaging tools (the namespace packages themselves weren't sufficient, I don't remember what exactly was the culprit).

Do we want to allow one bot to be composed of multiple packages, or do we require that every bot must be at most in one package? Latter would guarantee that no other package puts data in a bot's own directory? It could reduce the complexity. For developers it could reduce the complexity of maintenance, and for the users/admins the complexity of usage and understanding the system.

sebix:
Is that solved by #2413 ("Extension packages")?

kamil-certat:
I belive so

PR #2393, opened 2023-07-24
Change to namespace package

sebkuf:
It should change the ``intelmq/bots`` folder to a namespace package. Now it should be easier to add new Bots. The ``intelmqctl list bots`` functionality now also works with namespaces. Added a documentation how to add a new bot.
Implements #2350


intelmq/lib/utils.py
```diff
@@ -860,13 +862,22 @@ def list_all_bots() -> dict:
...
+    ns_pkg = intelmq.bots
 
-    botfiles = [botfile for botfile in pathlib.Path(base_path).glob('**/*.py') if botfile.is_file() and botfile.name != '__init__.py']
-    for file in botfiles:
-        file = Path(file.as_posix().replace(base_path, 'intelmq/bots'))
+    def iter_ns(pkg):
+        botmodules = []
+        for path, name, is_pkg in iter_modules(pkg.__path__, pkg.__name__ + '.'):
+            if is_pkg:
+                botmodules = botmodules + iter_ns(importlib.import_module(name))
```
sebix, line 871: This line causes the current unit test failures (import of a SyntaxError-throwing-bot fails with a SyntaxError, which is not handled correctly here). An example of catching the exception is nine lines below.

sebix, line 871: This line causes the current unit test failures (import of a SyntaxError-throwing-bot fails with a SyntaxError, which is not handled correctly here). An example of catching the exception is nine lines below.

intelmq/bots/__init__.py
```diff
@@ -0,0 +1 @@
+__path__ = __import__('pkgutil').extend_path(__path__, __name__)
```
sebix, line 1: trailing newlines missing here and in the other similar files

aTestNamespace/intelmqTest.egg-info/PKG-INFO
```diff
@@ -0,0 +1,33 @@
+Metadata-Version: 2.1
```
sebix, line 1: I think these files should be part of the repository, should they?

contrib/examples/example_add_bot/setup.py
```diff
@@ -0,0 +1,39 @@
+from setuptools import find_packages, setup
+
+setup(
+    name='intelmqExampleAddBot',
+    version=3.1,  # noqa: F821
```
sebix, line 5: should be a string AFAIK. And let's use semantic versioning
```suggestion
    version='3.1.0',
```

contrib/examples/example_add_bot/setup.py
```diff
@@ -0,0 +1,39 @@
...
+    maintainer='Sebastian Wagner',
+    maintainer_email='intelmq-dev@lists.cert.at',
```
sebix, line 7: I don't want to be responsible for code which I never wrote and never saw =)
```suggestion
    maintainer='Your Name',
    maintainer_email='you@example.com',
```

contrib/examples/example_add_bot/setup.py
```diff
@@ -0,0 +1,39 @@
...
+    python_requires='>=3.7',
```
sebix, line 8:
```suggestion
```
intelmq itself requires python >= 3.7 currently, that may not be true for bots. as the bots require the core library anyway, the requirement is active too without listing it here

docs/dev/guide.rst
```diff
@@ -891,6 +891,50 @@ The databases `<` 10 are reserved for the IntelMQ core:
...
+   ├── setup.py
+   └── intelmq
+        ├── __init__.py
+        └── bots
+             └── collectors
+                  └── your_bot
+                       └── your_bot.py
+
+The ``additional_bots`` folder needs to be defined e.g. with a ``setup.py``. Only the intelmq folder contains a ``__init__.py`` 
+file and it must contain only this line. Any additional code in ``__init__.py`` after this line will be inaccessible.
```
sebix, line 912:
```suggestion
file, which must have only the following line. Any additional code in ``__init__.py`` after this line will be inaccessible.
```

intelmq/lib/utils.py
```diff
@@ -860,13 +862,22 @@ def list_all_bots() -> dict:
     from intelmq.lib.bot import Bot  # noqa: prevents circular import
     bot_parameters = dir(Bot)
 
-    base_path = resource_filename('intelmq', 'bots')
+    ns_pkg = intelmq.bots
 
-    botfiles = [botfile for botfile in pathlib.Path(base_path).glob('**/*.py') if botfile.is_file() and botfile.name != '__init__.py']
-    for file in botfiles:
-        file = Path(file.as_posix().replace(base_path, 'intelmq/bots'))
+    def iter_ns(pkg):
```
sebix, line 867: Could be move the function to the outer scope?

sebix (commented):
Thank you for your work so far. Would be very cool if we can get this feature done :)

kamil-certat:
I'll take over the change and create a new PR as sebkuf is no longer available.

PR #2398, opened 2023-08-16
ENH: message: show reason for InvalidKey exception

wagner-intevation:
if the user wants to add a field with an invalid key, the reason was not clear
now the error message says that either
- the key is not allowed in the harmonization configuration or
- the key does not match the regular expression (for extra keys), e.g. is not lower case etc.

### DIFF
--- a/intelmq/lib/exceptions.py
+++ b/intelmq/lib/exceptions.py
@@ -88,8 +88,8 @@ class InvalidValue(IntelMQHarmonizationException):
 
 class InvalidKey(IntelMQHarmonizationException, KeyError):
 
-    def __init__(self, key: str):
-        message = "invalid key %s" % repr(key)
+    def __init__(self, key: str, additional_text: Optional[str] = None):
+        message = f"invalid key {key!r} {additional_text or ''}".strip()  # remove trailing whitespace if additional_text is not given
         super().__init__(message)
 
 
--- a/intelmq/lib/message.py
+++ b/intelmq/lib/message.py
@@ -13,7 +13,7 @@ import json
 import re
 import warnings
 from collections import defaultdict
-from typing import Any, Dict, Iterable, Optional, Sequence, Union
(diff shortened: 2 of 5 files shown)


intelmq/lib/exceptions.py
```diff
@@ -88,8 +88,9 @@ def __init__(self, key: str, value: str, reason: Any = None, object: bytes = Non
 
 class InvalidKey(IntelMQHarmonizationException, KeyError):
 
-    def __init__(self, key: str):
-        message = "invalid key %s" % repr(key)
+    def __init__(self, key: str, additional_text: Optional[str] = None):
+        postfix = f' {additional_text}' if additional_text else ''
+        message = f"invalid key {key!r}{postfix}"
```
kamil-certat: Why not just: 
```python
message = f"invalid key {key!r} {additional_text or ''}"
```

intelmq/lib/message.py
```diff
@@ -330,15 +332,15 @@ def unserialize(message_string: str):
...
         except KeyError:
-            return False
+            return False, 'This key is not allowed by the harmonization configuration'
         if key in self.harmonization_config or key == '__type':
-            return True
+            return True, None
         if subitem:
-            return HARMONIZATION_KEY_FORMAT.match(key)
+            return HARMONIZATION_KEY_FORMAT.match(key), f'Does not match regular expression {HARMONIZATION_KEY_FORMAT.pattern}'
         return False
```
kamil-certat: Cation: Default case return just one item instead of expected tuple

kamil-certat (approved):
Approved assuming you fix the last-resort return from validation function

intelmq/lib/exceptions.py
wagner-intevation: because of the missing whitespace before `additional_text` to separate the two text parts

PR #2399, opened 2023-08-24
FIX: Keep standalone flag after reloading

kamil-certat:
Library mode introduced 'standalone' flag for
bots running as a process. This flag instruct
the bot to handle signals, including SIGTERM.
 Unfortuneatly, this flag weren't preserved
when reloading the bot, what caused not being
able to gracefully stop reloaded bots.

Reloading is used when logrotate with recommended
configuration is enabled, what means that
this bug affects all 3.2.0 installations
from native packages after first logrotate.

### DIFF
--- a/intelmq/lib/bot.py
+++ b/intelmq/lib/bot.py
@@ -183,7 +183,7 @@ class Bot:
 
         try:
             self.logger.info('Bot is starting.')
-            self.logger.debug("Standalone mode: %s", self._standalone)
+            self.logger.debug("Standalone mode: %s.", self._standalone)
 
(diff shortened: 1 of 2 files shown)


intelmq/lib/bot.py
```diff
@@ -183,6 +183,7 @@ def __init__(self, bot_id: str, start: bool = False, sighup_event=None,
 
         try:
             self.logger.info('Bot is starting.')
+            self.logger.debug("Standalone mode: %s.", self._standalone)
```
sebix: Can we add this info to the `... initialized with ...` log line?

sebix (approved):

PR #2407, opened 2023-09-06
Docs: Update support for Ubuntu

sebix:
Ubuntu 20.04 is no longer supported because it only has dnspython 1.16.0, but intelmq requires >= 2.0.0
On Ubuntu 22.04, intelmq is available, but not intelmq-api

### DIFF
--- a/docs/user/installation.rst
+++ b/docs/user/installation.rst
@@ -30,9 +30,9 @@ The following instructions assume the following requirements. Python versions >=
 
 Supported and recommended operating systems are:
 
-* Debian Debian 11 Bullseye, Debian 12 Bookworm
+* Debian 11 Bullseye, Debian 12 Bookworm
 * openSUSE Tumbleweed, Leap 15.5
-* Ubuntu: 20.04 Focal, 22.04 Jammy
+* Ubuntu: 22.04 Jammy
 * For the Docker-installation: Docker Engine: 18.x and higher
 
 Other distributions which are (most probably) supported include AlmaLinux, CentOS, Fedora, FreeBSD 12, RHEL and RockyLinux.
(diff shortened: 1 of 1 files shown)


aaronkaplan:
Hm... what to do? Still plenty of 20.04 installations around.... and officially still supported.

"For each Ubuntu LTS release, Canonical maintains the Base Packages and provides security updates, including kernel livepatching, for a period of ten years. The lifecycle consists of an initial five-year standard security maintenance period, during which maintenance updates are publicly available without a subscription"
(source: https://ubuntu.com/about/release-cycle).

So for 20.04 this would mean that it's supported without ESM until April 2025 (!)

sebix:
Actually with the backport of dnspython in the repos, it should still work, changed the text accordingly. intelmq-api is an issue with ubuntu and Debian 10.

docs/user/installation.rst
```diff
@@ -30,9 +30,9 @@ The following instructions assume the following requirements. Python versions >=
 
 Supported and recommended operating systems are:
 
-* Debian Debian 11 Bullseye, Debian 12 Bookworm
+* Debian 11 Bullseye, Debian 12 Bookworm
 * openSUSE Tumbleweed, Leap 15.5
-* Ubuntu: 20.04 Focal, 22.04 Jammy
+* Ubuntu: 22.04 Jammy
```
kamil-certat, line 35: If deb still supports Ubuntu 20, shouldn't it be listed here as well? Or you wanted to make a difference between recommended and minimally supported?

aaronkaplan:
I would love to have that still listed. Reasoning: as mentioned before - it's going to be officially supported until 2025 at least and larger orgs don't change soo often if the bugfix support is given.

docs/user/installation.rst
sebix: > Or you wanted to make a difference between recommended and minimally supported?

Not sure how to describe the situation in the docs ideally. Buster and Focal work, but only with upgrading packages (either custom repo or pip3). Maybe making a separated list "Supported, but not recommended for new installations"?

sebix:
> and larger orgs don't change soo often if the bugfix support is given.

These larger orgs also have the resources. They just lack the will to contribute to IntelMQ. </Rant>

kamil-certat:
I'd suggest two sections:

- Recommended setup – only the newest LTS/active, described as where we put the main effort
- Supported setup – here all current LTS/active, described as where we want IntelMQ to work on and are open for any bug reports

EDIT: And a note that we are open for every contribution, also for not officially supported setups

aaronkaplan:
> > and larger orgs don't change soo often if the bugfix support is given.
> 
> These larger orgs also have the resources. They just lack the will to contribute to IntelMQ.

Well, working at one, I know that they have all kinds of other topics which often stop them from even being able to do so.
But that's another discussion and won't bring us forward here.

aaronkaplan:
I am fine with the proposed changes in this PR as is.

docs/user/installation.rst
aaronkaplan: okay, I'll create an issue for this topic and we can revisit the documentation on that topic later. See #2451

aaronkaplan:
Merging in, since we moved the documentation change to later . See #2451

aaronkaplan (approved):

aaronkaplan:
Closing as discussed in the chat.... We may re-open it if needed and wade through the mud of merging this one :)

PR #2414, opened 2023-10-07
STOMP+SSL-related changes: security fixes + convenience enhancement (+ doc updates/improvements)

### LINKED ISSUE #2409: STOMP collector and output bots: consider restricting the version of `stomp.py` dependency to `>=8.1.0`
Rationale:
* [security] `stomp.py` in versions older than `4.1.12` always uses the old `ssl.wrap_socket()` API which has no support for server name indication (SNI) and hostname matching (newer versions of `stomp.py` use more modern `ssl.SSLContext`-related APIs if possible).
* [security] By default, `stomp.py` in versions `8.0.0.` and `8.0.1` (`8.x.x` older than `8.1.0`) mistakenly does *not* perform hostname matching (an `ssl.SSLContext` in constructed manually, with *protocol* set to `ssl.PROTOCOL_TLS` which does *not* provide automatic hostname matching -- contrary to `ssl.PROTOCOL_TLS_CLIENT` which is used in the version `8.1.0` of `stomp.py`).
* In the implementation of the aforementioned bots, various versions of `stomp.py` require different code paths  -- see the `if` statements checking for (or related in other ways with) versions: `<4.1.20`, `>=4.1.21`, `>=5.0.0`, `>=6.1.0`. Restricting the version of `stomp.py` to a newer one would make it possible to simplify the code (and probably also make it more correct/reliable...[^1]).

[^1]: In particular, after some superficial tests it seems to me that disconnects/reconnects are not handled properly by the collector bot when the version of `stomp.py` is older than ~~`4.1.20`~~ `4.1.21` (although I admit I haven't investigated it more deeply...).

zuo:
The main purpose of this PR is to fix the following two problems -- both related to how the `stomp.py` library deals with SSL:

* (1) **[ad security]** Certain versions of `stomp.py`, that IntelMQ needs to be compatible with, use the `ssl` module's tools in ways that suffers from certain security weaknesses. *In particular,* ~~`stomp.py < 4.1.12`~~ [not supported anymore] ~~uses a deprecated helper: `ssl.wrap_socket()`~~ `stomp.py >=8.0, <8.1` mistakenly creates an `SSLContext` instance with the `check_hostname` flag unset; an important negative result is that the hostname of the STOMP server is *not* checked during the TLS handshake (!). *Also,* there are weaknesses (caused either by `stomp.py` or by older Python versions) of using too old versions of TLS (namely: 1.0 and 1.1 -- today considered insecure).

* (2) **[ad admin convenience]** No version of `stomp.py` makes it possible to load the *system's default CA certificates* -- condemning administrators to bother with manual updates of the CA certificate(s) file, even if the certificate of the STOMP server the bots connect to could be verified using some of the publicly available CA certificates which are part of nearly all mainstream operating system distributions... *Note:* this is the case with the upcoming new *n6* Stream API server's certificate.

An important part of the changes to fix these problems is a non-public helper class, `intelmq.lib.mixins.stomp._StompPyDedicatedSSLProxy`, which implements a kind of transparent proxy object: it wraps the `ssl` attribute of the `stomp.transport` module (originaly set to the `ssl` module object) replacing some of the `ssl` module's tools with their patched variants. *Note:* the `ssl` module itself and all its members are left untouched -- only the `ssl` member of the `stomp.transport` module is replaced with an instance of that class.

For more details -- see the docstrings and comments in the code. [*See the 4th commit...*]

Apart from that, versions of `stomp.py >=4.1.8, <4.1.12` are no longer supported + some details in `debian/control` have been fixed. [*See the 3rd commit...*]

Together with the aforementioned changes, relevant parts of the documentation (including also the feeds spec and the changelog) have been appropriately updated + also improved by including some extra details concerning the upcoming switch to the new variant of *n6* Stream API (the one with username+password-based authentication). [*Also, see the 3rd and 4th commit...*]

Also, an additional minor fix has been applied to the imports part of the STOMP collector's module, concerning the `stomp.py`-imports-related logic. [*See the 5th commit...*]

Plus, apart from all that, several additional (also STOMP-or-*n6*-integration-related) doc updates/improvements as well as very minor/cosmetic code tweaks have been made. [*See the 1st, 2nd and 6th commit...*]

### DIFF
--- a/intelmq/bots/outputs/stomp/output.py
+++ b/intelmq/bots/outputs/stomp/output.py
@@ -4,34 +4,51 @@
 
 # -*- coding: utf-8 -*-
 
-from intelmq.lib.bot import OutputBot
-from intelmq.lib.mixins import StompMixin
-
 try:
     import stomp
 except ImportError:
     stomp = None
 
+from intelmq.lib.bot import OutputBot
+from intelmq.lib.mixins import StompMixin
+
 
 class StompOutputBot(OutputBot, StompMixin):
     """Send events to a STMOP server"""
     """ main class for the STOMP protocol output bot """
-    exchange: str = "/exchange/_push"
-    heartbeat: int = 60000
+
     http_verify_cert = True
     keep_raw_field: bool = False
     message_hierarchical_output: bool = False
     message_jsondict_as_string: bool = False
     message_with_type: bool = False
-    port: int = 61614
-    server: str = "127.0.0.1"  # TODO: could be ip address
     single_key: bool = False
+
+    server: str = '127.0.0.1'  # <- TODO: change to 'n6stream.cert.pl' (==StompCollectorBot.server)
+    port: int = 61614
+    exchange: str = '/exchange/_push'
+    heartbeat: int = 60000
+
+    # Note: the `ssl_ca_certificate` configuration parameter must be set:
+    # * *either* to the server's CA certificate(s) file path,
+    # * *or* to an empty string -- dictating that the SSL tools employed
+    #   by the `stomp.py`'s machinery will attempt to load the system’s
+    #   default CA certificates.
+    # The latter, if applicable, is more convenient -- by avoiding the
+    # need to manually update the CA certificate(s) file.
+    ssl_ca_certificate: str = 'ca.pem'  # <- TODO: change to ''
+    # (^ TODO: could also be pathlib.Path)
+
     auth_by_ssl_client_certificate: bool = True
-    username: str = 'guest'  # ignored if `auth_by_ssl_client_certificate` is true
-    password: str = 'guest'  # ignored if `auth_by_ssl_client_certificate` is true
-    ssl_ca_certificate: str = 'ca.pem'  # TODO: could be pathlib.Path
-    ssl_client_certificate: str = 'client.pem'  # TODO: pathlib.Path
-    ssl_client_certificate_key: str = 'client.key'  # TODO: patlib.Path
+
+    # Used if `auth_by_ssl_client_certificate` is true (otherwise ignored):
+    ssl_client_certificate: str = 'client.pem'       # (cert file path)
+    ssl_client_certificate_key: str = 'client.key'   # (cert's key file path)
+    # (^ TODO: could also be pathlib.Path)
+
+    # Used if `auth_by_ssl_client_certificate` is false (otherwise ignored):
+    username: str = 'guest'   # (STOMP auth *login*)
+    password: str = 'guest'   # (STOMP auth *passcode*)
 
     _conn = None
 
--- a/intelmq/lib/mixins/stomp.py
+++ b/intelmq/lib/mixins/stomp.py
@@ -4,18 +4,25 @@ SPDX-FileCopyrightText: 2017 Sebastian Wagner, 2023 NASK
 SPDX-License-Identifier: AGPL-3.0-or-later
 """
 
+import enum
+import os
+import ssl
+import sys
 from typing import (
     Any,
     Callable,
     List,
     NoReturn,
     Tuple,
+    Union,
 )
 
 try:
     import stomp
 except ImportError:
     stomp = None
+else:
+    import stomp.transport
 
 from intelmq.lib.exceptions import MissingDependencyError
 
@@ -31,14 +38,26 @@ class StompMixin:
     port: int
     heartbeat: int
 
+    # Note: the `ssl_ca_certificate` configuration parameter must be set:
+    # * *either* to the server's CA certificate(s) file path,
+    # * *or* to an empty string -- dictating that the SSL tools employed
+    #   by the `stomp.py`'s machinery will attempt to load the system’s
+    #   default CA certificates.
+    # The latter, if applicable, is more convenient -- by avoiding the
+    # need to manually update the CA certificate(s) file.
+    ssl_ca_certificate: str
+    # (^ TODO: could also be pathlib.Path)
+
     auth_by_ssl_client_certificate: bool
 
-    username: str  # to be ignored if `auth_by_ssl_client_certificate` is true
-    password: str  # to be ignored if `auth_by_ssl_client_certificate` is true
+    # Used if `auth_by_ssl_client_certificate` is true (otherwise ignored):
+    ssl_client_certificate: str       # (cert file path)
+    ssl_client_certificate_key: str   # (cert's key file path)
+    # (^ TODO: could also be pathlib.Path)
 
-    ssl_ca_certificate: str  # TODO: could be pathlib.Path
-    ssl_client_certificate: str  # TODO: could be pathlib.Path
-    ssl_client_certificate_key: str  # TODO: could be patlib.Path
+    # Used if `auth_by_ssl_client_certificate` is false (otherwise ignored):
+    username: str   # (STOMP auth *login*)
+    password: str   # (STOMP auth *passcode*)
 
     #
     # Helper methods intended to be used in subclasses
@@ -73,7 +92,11 @@ class StompMixin:
           to be passed to the `connect()` method of the aforementioned
           `<STOMP connection>` object.
         """
+        _StompPyDedicatedSSLProxy.patch_stomp_transport_ssl()
         ssl_kwargs, connect_kwargs = self.__get_ssl_and_connect_kwargs()
+        # Note: here we coerce `port` to int just to be on the safe
+        # side, as some historical versions of `etc/feeds.yaml` used
+        # to set it to a string.
         host_and_ports = [(self.server, int(self.port))]
         stomp_connection = stomp.Connection(host_and_ports=host_and_ports,
                                             heartbeats=(self.heartbeat,
@@ -98,8 +121,8 @@ class StompMixin:
         if stomp is None:
             raise MissingDependencyError('stomp',
                                          additional_text=cls._DEPENDENCY_NAME_REMARK)
-        if stomp.__version__ < (4, 1, 8):
-            raise MissingDependencyError('stomp', version="4.1.8",
+        if stomp.__version__ < (4, 1, 12):
+            raise MissingDependencyError('stomp', version="4.1.12",
                                          installed=stomp.__version__,
                                          additional_text=cls._DEPENDENCY_NAME_REMARK)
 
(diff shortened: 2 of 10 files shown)


intelmq/lib/mixins/stomp.py
```diff
@@ -170,3 +197,320 @@
...
+            if not certfile:
+                raise NotImplementedError('`certfile` unexpectedly empty/None '
+                                          'whereas `keyfile` is not None')
+
+        ssl_context = self.__patched__create_default_context(cafile=ca_certs)
+        if certfile:
+            ssl_context.load_cert_chain(certfile, keyfile)
+
+        # (note: this is the only place where `self.__server_hostname` is needed)
+        return ssl_context.wrap_socket(sock, server_hostname=self.__server_hostname)
```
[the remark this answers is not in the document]
zuo: In a newer version of this commit (see below...), the code enforces TLSv1_2 as the `minimum_version`.

intelmq/bots/outputs/stomp/output.py
```diff
@@ -4,34 +4,51 @@
...
-    ssl_client_certificate_key: str = 'client.key'  # TODO: patlib.Path
+
+    # Used if `auth_by_ssl_client_certificate` is true (otherwise ignored):
+    ssl_client_certificate: str = 'client.pem'       # (cert file path)
+    ssl_client_certificate_key: str = 'client.key'   # (cert's key file path)
+    # (^ TODO: could also be pathlib.Path)
+
+    # Used if `auth_by_ssl_client_certificate` is false (otherwise ignored):
+    username: str = 'guest'   # (STOMP auth *login*)
+    password: str = 'guest'   # (STOMP auth *passcode*)
```
kamil-certat, line 51: More just a FYI, but basically, if parameters are set by StompMixIn, it shouldn't be necessary to repeat them in bot's class. As there is quite a long part with parameters' explanation, I'd put it in the Mixin class, and just leave a note in bots' classes where to look for them. On the other hand, here they are easier accessible. What do you think, @zuo & @sebix?

intelmq/lib/mixins/stomp.py
```diff
@@ -170,3 +197,328 @@ def __get_ssl_and_connect_kwargs(self) -> Tuple[dict, dict]:
                 passcode=self.password,
             )
         return ssl_kwargs, connect_kwargs
+
+    def __patch_ssl_related_stomp_stuff(self,
+                                        host_and_ports: List[Tuple[str, int]]) -> None:
+        assert len(host_and_ports) == 1
```
kamil-certat: Do not rely on assertions in non-test code, they are intended for the debug purposes only, and can be easily [disabled with optimizations](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement).

docs/user/bots.rst
```diff
@@ -947,7 +947,7 @@ Install the `stomp.py` library from PyPI:
 * **Feed parameters** (see above)
 * `server`: STOMP server's hostname or IP, e.g. "n6stream.cert.pl" (which is default)
 * `port`: STOMP server's port number (default: 61614)
-* `exchange`: STOMP *destination* to subscribe to, e.g. "/exchange/my.example.org/*.*.*.*"
+* `exchange`: STOMP *destination* to subscribe to, e.g. ``"/exchange/my.example.org/*.*.*.*"``
```
zuo: Without the ``` `` ``` marks rendering of the `*.*.*.*` fragment was wrong.

sebix: the backtick markup is correct, to display it as code/preformatted text

intelmq/lib/mixins/stomp.py
```diff
@@ -74,7 +93,11 @@ def prepare_stomp_connection(self) -> Tuple['stomp.Connection', dict]:
           `<STOMP connection>` object.
         """
         ssl_kwargs, connect_kwargs = self.__get_ssl_and_connect_kwargs()
+        # Note: here we coerce `port` to int just to be on the safe
+        # side, as some historical versions of `etc/feeds.yaml` used
+        # to set it to a string.
         host_and_ports = [(self.server, int(self.port))]
+        self.__patch_ssl_related_stomp_stuff(host_and_ports)
```
kamil-certat:
```suggestion
        self.__patch_ssl_in_stomp(host_and_ports)
```

Or so, to make it look better ;)

intelmq/lib/mixins/stomp.py
```diff
@@ -170,3 +197,328 @@ def __get_ssl_and_connect_kwargs(self) -> Tuple[dict, dict]:
                 passcode=self.password,
             )
         return ssl_kwargs, connect_kwargs
+
+    def __patch_ssl_related_stomp_stuff(self,
+                                        host_and_ports: List[Tuple[str, int]]) -> None:
+        assert len(host_and_ports) == 1
+        [(server_hostname, _)] = host_and_ports
```
kamil-certat: Instead of _pack (L99), call, assert, unpack, ignore one, call_ please just write in L100:

```
_StompPyDedicatedSSLProxy.patch_stomp_transport_ssl(self.server)
```

and remove this function.

intelmq/lib/mixins/stomp.py
```diff
@@ -170,3 +197,328 @@ def __get_ssl_and_connect_kwargs(self) -> Tuple[dict, dict]:
...
+    #
+    # Checking and replacing `stomp.transport` module's `ssl` member
+
+    @classmethod
+    def patch_stomp_transport_ssl(cls, server_hostname: str) -> None:
+        # (the following invariants are considered already guaranteed)
+        assert sys.version_info[:2] >= (3, 7)
+        assert stomp is not None
+        assert stomp.__version__ >= (4, 1, 8)
+        assert getattr(stomp, 'transport', None) is not None
```
kamil-certat: Please remove. Those conditions _are considered already guaranteed_, so verifying them again with the debug-only asserts should be considered unnecessary - it would make sense in a mathematical proof, but not in a service. It introduces an additional place, where we possibly should maintenance the check - e.g., for the consistency, if we drop the Python 3.7 support, this place should also be updated.

kamil-certat (changes requested):
I have started the review, but I after a while of analysing it, I'd like to start with challenging some architectural decisions you've made.

I love the improvements in the security, however please see that the solution is really crafted for a very specific conditions, and can easily break with any upstream changes, as well as duplicate a lot of checks. As we support a library mode, I can imagine creating two objects of bots connecting to two different hosts in a one process - and then the Proxy will keep the first hostname and raise an exception.

Consider we re-design the solution in following way:

1. Do not check things that are already granted, like minimum version of Python, if it doesn't change the code flow.
2. Create a subclass of SSLContext, and override related methods there, instead of patching them. Keep the default behaviour any time the call isn't the one marking a special behaviour, but just use the `super()....` for that.
3. The proxy should override only the expected objects (SSLContext class, create_default_context etc.), and do not change the `ssl` behaviour in any other way.

If the conditions doesn't look like a call from expected stomp, do not raise - assume it was another legit use of the stomp etc., just not the one we want to secure.

zuo:
> I have started the review, but I after a while of analysing it, I'd like to start with challenging some architectural decisions you've made.
> 
> I love the improvements in the security, however please see that the solution is really crafted for a very specific conditions, and can easily break with any upstream changes, as well as duplicate a lot of checks. As we support a library mode, I can imagine creating two objects of bots connecting to two different hosts in a one process - and then the Proxy will keep the first hostname and raise an exception.
[...]

OK, I will reconsider the design...

One question: could we drop support for `stomp.py < 4.1.12`? The hostname-related problem you mentioned above would disappear (as versions `>= 4.1.12` do not use `ssl.wrap_socket()` [at least not for `Python >= 3.7`], so the patched variant of it -- which is the only place where that early-stored `hostname` is used -- could be entirely removed).

sebix:
> One question: could we drop support for `stomp.py < 4.1.12`? The hostname-related problem you mentioned above would disappear (as versions `>= 4.1.12` do not use `ssl.wrap_socket()` [at least not for `Python >= 3.7`], so the patched variant of it -- which is the only place where that early-stored `hostname` is used -- could be entirely removed).


Yes, absolutely.
By way of comparison: Debian 10 buster and Ubuntu 20.04 focal have python3-stomp == 4.1.19.

We can also consider dropping support for < 6.
We seek for compatibility and don't want to enforce that users strictly need the absolute latest version of a library (makes sense in very controlled and similar environments), but if the required effort is high, the reward is low (it affects only one single bot, not the core), dropping compatibility is reasonable.

intelmq/bots/outputs/stomp/output.py
sebix: Yes, re-defining them is not necessary

sebix (approved):
The more important review comes from @kamil-certat as he is a user/operator of the bot and knows more details, also from a user- and admin perspective.

zuo:
@sebix @kamil-certat 

OK, this is a redesigned version + with a few additional minor improvements/fixes/updates -- in particular, minimum `stomp.py` version has been changed from `4.1.8` to `4.1.12`, and a package name has been fixed in `debian/control` (see the relevant commit...).

kamil-certat (approved):
I think it looks good. Thanks for the change!

kamil-certat:
@sebix @aaronkaplan Could you please merge this change? It's quite important for changes in n6 (using default CAs).

kamil-certat (approved):

kamil-certat:
@zuo We have merged the new docs system, may you merge `develop` and copy your changes in documentation to `docs/user/bots.md`?

zuo:
> @zuo We have merged the new docs system, may you merge `develop` and copy your changes in documentation to `docs/user/bots.md`?

OK, I will.

sebix:
I rebased the PR on the branch 'develop' and moved all docs to the new files. As the documentation changes were spread over five commits, I squashed them all together, because keeping the changes apart in separate commits, and moving every single change manually to the new format would be a heck of a lot of effort.

As @kamil-certat approved the PR, I will merge it after the CI checks pass.

zuo:
@sebix @kamil-certat Oh, thank you for the rebase and merge! And I am sorry I haven't managed to do that promptly!

sebix:
@zuo Thank you for improving, maintaining and contributing to IntelMQ!

PR #2419, opened 2023-10-31
DOC: Changes documentation to mkdocs

gethvi:
Live version: https://gethvi.github.io/intelmq/mkdocs/


setup.py
```diff
@@ -35,7 +47,7 @@
     module = '.'.join(file.with_suffix('').parts)
     BOTS.append('{0} = {0}:BOT.run'.format(module))
 
-with open(os.path.join(os.path.dirname(__file__), 'README.rst')) as handle:
+with open(os.path.join(os.path.dirname(__file__), 'README.md')) as handle:
```
sebix, line 50: Please set `long_description_content_type='text/markdown'` (see https://packaging.python.org/en/latest/guides/making-a-pypi-friendly-readme/)

sebix (changes requested):

setup.py
gethvi: Thanks. Done.

.github/workflows/build-docs.yml
```diff
@@ -1,39 +1,61 @@
...
+        uses: actions/checkout@v3
+
+      - name: "Setup python"
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.x
+
+      - name: "Install build dependencies"
+        run: |
+          pip install mkdocs-material mike lunr pygments mkdocstrings[python] mkdocs-material mkdocs-glightbox mkdocs-redirects mkdocs-minify-plugin
```
kamil-certat, line 40: Wouldn't it be possible to reuse the list from `setup.py`? Maybe by creating a new extra? Otherwise, we need to again maintain two separated dependency lists.

Makefile
```diff
@@ -0,0 +1,30 @@
+# SPDX-FileCopyrightText: 2023 Filip Pokorný
+# SPDX-License-Identifier: AGPL-3.0-or-later
```
kamil-certat, line 2: +1 for Makefile
+1 for using `pip-compile`

-1 for pushing Makefile with tools that aren't installed. `pip-compile` isn't so far mentioned anywhere in the IntelMQ, so I'd suggest leaving this Makefile with `docs` only (for now)

mkdocs.yml
```diff
@@ -0,0 +1,154 @@
+# SPDX-FileCopyrightText: 2023 Filip Pokorný
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+site_name: IntelMQ
+site_url: https://intelmq.org
```
kamil-certat: Should it be `intelmq.org` or `docs.intelmq.org`?

kamil-certat (commented):
I haven't checked all changes in the docs (this will come with the usage, but I feel you also addressed some my concerns about the docs structure :D). I love the change, and want to see it online <3

However, could you please bring back to the README two things:
- IntelMQ logo (because it looks sad on GitHub without it)
- European funding info (because I'm pretty sure there are still active legal requirements to say it clearly).

I'd also add a link to docs in the README - even if it would be redundant on the docs website, it's important in GitHub

Makefile
gethvi: Removed.

mkdocs.yml
gethvi: Fixed.

.github/workflows/build-docs.yml
aaronkaplan: @kamil-certat we'll do that in a second step, ok? 
I have @gethvi 's attention briefly now. And this does not seem like it;'s critical. Will merge now into develop and then fix broken things

aaronkaplan: I.e. I'll take the blame to clean it up in case it's broken

PR #2420, opened 2023-11-01
FIX: prevent jinja expert bot crash on logrotate

monoidic:
This PR prevents this error on logrotate, which appears to be caused by `init` getting rerun after `self.fields` has already been updated by `init`?
```
Jinja-Expert - ERROR - Bot initialization failed.
  File "/usr/lib/python3/dist-packages/intelmq/lib/bot.py", line 241, in __init__
    self.init()
  File "/usr/lib/python3/dist-packages/intelmq/bots/experts/jinja/expert.py", line 38, in init
    if not template.startswith("file:///"):
AttributeError: 'Template' object has no attribute 'startswith'
```

### DIFF
--- a/intelmq/bots/experts/jinja/expert.py
+++ b/intelmq/bots/experts/jinja/expert.py
@@ -27,7 +27,8 @@ class JinjaExpertBot(ExpertBot):
           extra.somejinjaoutput: file:///etc/intelmq/somejinjatemplate.j2
     """
 
-    fields: Dict[str, Union[str, Template]] = {}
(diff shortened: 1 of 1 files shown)


kamil-certat (approved):
Thanks for the change, LGTM!

> which appears to be caused by init getting rerun after self.fields has already been updated by init?

This is typical situation on reloading the bot, when the process isn't stopped (as during the restart), and `init()` is expected to refresh the configuration

sebix:
Merged, based on kamil's review.

PR #2427, opened 2023-11-22
doc: fix logo display in root repository

bernhardreiter:
* Use an absolute link from docs/index.md to make sure that the svg logo is displayed from the root repository as well as when viewing index.md directly.


gethvi:
Thanks!

PR #2428, opened 2023-11-22
WIP: Merging of API and Manager to IntelMQ project

gethvi:
UPDATED.

Work in progress.

This PR merges API and Manager to the main IntelMQ repository.

Pros:
- Users only need to install one piece of software.
- No need to juggle with versions of three components.
- We only need to make sure that the code works together within each commit.
- Less packaging.
- Easier implementation when pieces of code are aware of each other and you can include/import them.
- Users can disable API and/or Manager easily.
- One unified configuration file.

Cons:
- The additional code and requirements take few more kB of disk space.

---

Updated project structure:

```tree
intelmq
├── app
│   ├── api       # <-- updated sources of intelmq-api
│   └── webgui    # <-- updated sources of intelmq-manager
├── bin
├── bots
├── etc
├── lib
└── tests
    └── app
        └── api   # <-- updated tests of intelmq-api
```

---

1. This PR adds `intelmq.yaml` configuration file with `server` block and adds new options, example:
```yaml
server:
  host: 127.0.0.1
  port: 8080
  workers: 3
  debug: false
  enable_webgui: true
  session_store: /var/lib/intelmq/server/sessions.sqlite
  access_log: /var/log/intelmq/access.log
  allowed_path: /var/lib/intelmq/bots/
  intelmq_ctl_cmd:
    - /usr/bin/intelmqctl
```

---
2. This PR changes of Manager (webgui):
- Mako templates changed to jinja templates.
- Serves the website using FastAPI (used for the API as well).
- All links are dynamically generated.
- Links changed from `/management.html` to nicer `/management` etc.
- Removes rendering of login form from HTML site when session_store is disabled.
- Removes problem with double slash URL.
---

3. This PR integrates `intelmq-add-user` into one consistent CLI:

```
intelmq
intelmq server start --debug --port 8080 --host 127.0.0.1
intelmq server adduser --username intelmq --password secret
```

It is possible to run either:
- production ready gunicorn server with `intelmq server start`
- development uvicorn server with `intelmq server start --debug`.

---

4. Cleans up and updates intelmqsetup from code no longer needed.

---

5. Changes positions.conf file path to /var/lib/intelmq/server/positions.json. This file is hardly ever edited manually by the user, no need to put it in etc.


kamil-certat (changes requested):
This PR is much much more than merging all projects in one, it also rewrites a number of IMQ parts, introduce new dokerization, rewrites templating... Although I appreciate the direction, it's not possible to review properly such a change. Please split it into a few PRs, my suggestion is:

1. Merge in API (*)
2. Merge in Manager
3. Unify configuration file (*)
4. Rewrite `intelmqsetup` (*)
5. Introduce new docker image

(*) A few questions I have already:
1. You have introduced the `intelmq [serve/setup]` command. I support it, but it's inconsistent with `intelmqctl` - we should discuss, if we need both (my first idea: deprecate `intelmqctl` in favour of `intelmq` - but even if so, please do not migrate and introduce new stuff at once ;))
2. Code responsible for handling `intelmq serve` should be in my opinion moved to `web/` and just imported in the main file.
  - side note: if we touch commands so big, we may consider moving to Click in the future (it should be also available as DEB package as well)
3. Do I see correctly, that you removed Gunicorn and systemd units at all? Units are for DEB packages only, but using only `uvicorn.run` isn't a production-grade solution: https://www.uvicorn.org/deployment/
4. I support unification of configuration file, but I'm against using `global` in the runtime config. The technical issue here is that this `global` is then propagated to every bot (so basically, if any bot has a `server` config field, we just override it silently. I would suggest introducing a separated `config.yaml` for IntelMQ configuration, and leave `runtime.yaml` for workflow-only config. This would also be a good move if we eventually introduce supoort for multiple workflows.

gethvi:
Thank you for the review. I understand the concerns, it's just easier to push more changes at once instead spending time to make things work in one way just to open another PR to make them work another way.

So at least the first three steps (merging of API, Manager and unify the configuration) seems like logical steps to put together. The rest just snowballed from there.

> You have introduced the intelmq [serve/setup] command. I support it, but it's inconsistent with intelmqctl - we should discuss, if we need both (my first idea: deprecate intelmqctl in favour of intelmq - but even if so, please do not migrate and introduce new stuff at once ;))

It might be inconsistent with `intelmqctl`, but it's more consistent than a bunch of separate script callings such as:
- `intelmq-api-adduser`
- `intelmqsetup` (which I included just because the merging of projects allowed to remove some of it's now unnecessary code, so really it's just moved somewhere else and it's backwards compatible, you can still call `intelmqsetup`)
- `hug -m intelmq_api.serve` (old I know but still)
- `/usr/bin/gunicorn intelmq_api.main:app --workers 4 --worker-class uvicorn.workers.UvicornWorker --bind unix:intelmq_api.sock`.

My current solution was to just allow user easier and consistent interface to all these individual scripts. It doesn't intersect with `intelmqctl` functionality. The server options should be in my opinion included in the intelmq config (and my plan was to add more later such as workers count). User shouldn't need to know any python, any uvicorn, gunicorn or any other underlaying technology. All the user needs to do in my book is to edit the config file and run a command like `intelmq server start` which I introduced. Make it simple.

> Code responsible for handling intelmq serve should be in my opinion moved to web/ and just imported in the main file.
> *    side note: if we touch commands so big, we may consider moving to Click in the future (it should be also available as DEB package as well)

Not sure I totally agree with moving the server.py into web. Web module deals only with Manager code and the user has the option to not even enable it by `enable_webgui: false`. The server.py only imports the APIRouter object from web.

On the other hand I do agree that Click might be better option.

> Do I see correctly, that you removed Gunicorn and systemd units at all? Units are for DEB packages only, but using only uvicorn.run isn't a production-grade solution: https://www.uvicorn.org/deployment/

Thank you, I just simply forgot about the unit files and Gunicorn. I will fix this.

> I support unification of configuration file, but I'm against using global in the runtime config. The technical issue here is that this global is then propagated to every bot (so basically, if any bot has a server config field, we just override it silently. I would suggest introducing a separated config.yaml for IntelMQ configuration, and leave runtime.yaml for workflow-only config. This would also be a good move if we eventually introduce supoort for multiple workflows.

I agree. I actually tried to put it on the same level as `global`, but it broke a lot of things, because other parts of the code thought `server` is just another bot. :smile: So I went with the easier way (and less changes).

 I also agree with separation of intelmq config and workflow config(s).

kamil-certat:
Quick reply: ;)

> I understand the concerns, it's just easier to push more changes at once instead spending time to make things work in one way just to open another PR to make them work another way.

I fully agree if it requires making things in another way. If the change small, then I think we should separate it (e.g. - API can work without manager, so we can first clean up the API, and then add manager on top of it, without changing any code, just adding new).

My trouble is, that I need to put reviewing IntelMQ between other things, and it's hard to do so with a big PR ;)

> Not sure I totally agree with moving the server.py into web. Web module deals only with Manager code [...]

Then I probably missed that web is for Manager only - I'd like to just move the `server` from the `__main__.py`, my concern here is that we have a history of big scripts doing many things, so I'd like to logically separate subcommands. 

> It might be inconsistent with intelmqctl, but it's more consistent than a bunch of separate script callings such as:

I agree, and my goal was to start discussion about moving all into `intelmq xxx`, and deprecate `intelmqctl` as well.

> All the user needs to do in my book is to edit the config file and run a command like intelmq server start which I introduced. Make it simple.

I definitely agree, and I like it. I just belive it should be a wrapper around qunicorn (but maybe something has changed, and Uvicorn is producion-ready now?). And `uvicorn`, the same as `hug`, should never be visible to user - they are development-only servers.

contrib/app/api/intelmq-api.service
```diff
@@ -0,0 +1,26 @@
...
+[Service]
+
+# To override settings path, use e.g.:
+# Environment="INTELMQ_API_CONFIG=/etc/intelmq/api-config.json"
+
+Environment="ROOT_PATH=/intelmq"
+User=www-data
+Group=www-data
+RuntimeDirectory=gunicorn
+WorkingDirectory=/usr/lib/python3/dist-packages/intelmq_api/
```
kamil-certat, line 18: Here should be a valid path to the IntelMQ package (if installed from deb)

contrib/app/api/intelmq-api.service
```diff
@@ -0,0 +1,26 @@
...
+ExecStart=/usr/bin/gunicorn intelmq_api.main:app --workers 4 --worker-class uvicorn.workers.UvicornWorker --bind unix:intelmq_api.sock
```
kamil-certat, line 19: This should be updated to use new running method

contrib/app/api/intelmq-api.socket
```diff
@@ -0,0 +1,12 @@
+# SPDX-FileCopyrightText: 2022 CERT.at GmbH <https://cert.at/>
+# SPDX-License-Identifier: CC0-1.0
+
+[Unit]
+Description=The socket to handle IntelMQ API requests
+
+[Socket]
+ListenStream=/usr/lib/python3/dist-packages/intelmq_api/intelmq_api.sock
```
kamil-certat, line 8: This path should be updated

contrib/app/webgui/manager-apache.conf
```diff
@@ -0,0 +1,13 @@
+# SPDX-FileCopyrightText: 2020 IntelMQ Team
+#
+# SPDX-License-Identifier: CC0-1.0
+
+Alias /intelmq-manager /usr/share/intelmq_manager/html/
```
kamil-certat, line 5: Is this path still valid?

debian/systemd/intelmq.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
```
kamil-certat, line 1: Should it replace `intelmq-api.service`? Please also name it rather `intelmq-web` to say clear, that it's not the botnet service

setup.py
```diff
@@ -63,7 +71,7 @@
     extras_require={
         'development': TESTS_REQUIRES + DOCS_REQUIRES,
     },
-    packages=find_packages(),
+    packages=find_namespace_packages(),
```
kamil-certat, line 74: NIT: I would stay with a normal packages, this is allowing unintentional extending the package structure by other packages, and can then break the things we have in `__init__` files.

intelmq/app/server.py
```diff
@@ -0,0 +1,96 @@
...
+from intelmq.version import __version__
+import intelmq.app.dependencies
+import intelmq.app.api.exceptions
+
+from intelmq.app.config import Config
+from intelmq.app.api.router import router as api_router
+from intelmq.app.webgui.router import router as web_router
+
+
+class IntelMQUvicornWorker(UvicornH11Worker):
```
kamil-certat, line 26: So... It's just the pure Uvicorn worker, we don't need to subclass it :)

intelmq/app/server.py
```diff
@@ -0,0 +1,96 @@
...
+)
+
+
+@app.on_event("startup")
+def init_app():
+    intelmq.app.dependencies.startup(config)
+
+
+app.add_middleware(CORSMiddleware, allow_origins=config.allow_origins, allow_methods=("GET", "POST"))
+app.include_router(api_router, prefix="/api/v1")
```
kamil-certat, line 62: :x:  This is a breaking change in the API path. We currently use the `v1/api`, and as so, please do not silently change paths. It then breaks any API usage we don't control.

intelmq/bin/intelmqsetup.py
kamil-certat, line 1: :x: I definitely disagree with removing support for webserver/reverse proxy configuration, as well as deprecating disabling installing API and/or IntelMQ Manager.

Please note, IntelMQ is still used in many VM-native deployments. The reverse proxy configuration is there a basic requirement. This is not needed in Docker-based deployments, and because of that it should be the optional behaviour of the `intelmqsetup`. Removing this support creates a big headache for admins of current IntelMQ deployments (including me), and drops the ability to automatically update them.

scripts/intelmq-api-setup-systemd
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
kamil-certat, line 1: This script will most probably not work with the changes

intelmq/app/config.py
```diff
@@ -0,0 +1,81 @@
...
+
+    workers: int = 2
+
+    host: str = "0.0.0.0"
+
+    port: int = 8080
+
+    debug: bool = False
+
+    def __init__(self):
```
kamil-certat, line 44: NIT: This loader looks like a little overkill that could mostly be replaced by a loop ;)

intelmq/__init__.py
```diff
@@ -20,18 +20,22 @@
...
     DEFAULT_LOGGING_PATH = os.path.join(ROOT_DIR, "var/log/")
     VAR_RUN_PATH = os.path.join(ROOT_DIR, "var/run/")
+    VAR_SERVER_PATH = os.path.join(ROOT_DIR, "var/lib/server/")
     VAR_STATE_PATH = os.path.join(ROOT_DIR, "var/lib/bots/")
 
 
 DEFAULT_LOGGING_LEVEL = "INFO"
 HARMONIZATION_CONF_FILE = os.path.join(CONFIG_DIR, "harmonization.conf")
 RUNTIME_CONF_FILE = os.path.join(CONFIG_DIR, "runtime.yaml")
+INTELMQ_CONF_FILE = os.path.join(CONFIG_DIR, "intelmq.yaml")
```
kamil-certat, line 37: NIT: `/etc/intelmq/intelmq.yaml` sounds like a little too much of IntelMQ, maybe name it `config.yaml`?

kamil-certat (changes requested):
Hey, I definitely support merging all projects, I think we're almost there (for me).

I have the following general questions/concerns

1) Confusing names

You have introduced the names "app" and "server" into the code, which I think can be confusing. It looks like a very core part to me (especially with the 'intelmq server start' command - it looks more like starting the botnet to me), but in our case the "very core" are bots, and I'd like to avoid that. Also, the "app" is in one place, then "serve" in others, and they don't map to each other - which makes it harder to find the right part of the code if you don't have the inside knowledge. I spent a long time thinking about how to solve that. What do you think of the name "web"? We can use it instead of "app" and "server" everywhere: in directory structures and commands, and it would fit better and say better what this piece of code is about.

2) Renaming the IntelMQ Manager

I see you have introduced the term "WebUI" into the code instead of IntelMQ Manager. It's functionally correct, IntelMQ Manager is the web user interface, but my experience says that every renaming and difference in names used by users and developers makes the onboarding harder for both, and just leads to more time spent on explanations and reviews (I had a few projects struggling with this...). Not to mention the Google/Microsoft culture of renaming tools every month ;). I don't see any benefit in changing the code to webui, it's exactly the same as the established name of IntelMQ Manager. Could we keep the name "Manager" in the code as well?

3) Broken support for non-Docker deployments and customisation

The one-line startup command is a great thing, and also the right way to work with Docker-like deployments where we communicate with an external reverse proxy via IP:PORT sockets. However, IntelMQ still supports (and is the primary method for) VM-native deployments. There are a few things to keep in mind:

  1) We need to provide SystemD services to run our web service (you started this),
  2) we need to provide a sample reverse proxy configuration and not rely on an external one (the config will then be installed by DEB package and/or manually tuned) - this is not provided (historically we have chosen Apache, so let's stick with it as the default),
  3) The communication between the service and the proxy should go through a Unix socker - it's an important thing, in a dockerised environment you limit who can talk to whom through different bridges (networks). In a VM-native environment, the network stack is (to simplify a bit) one for all on the machine, and so the privileges set on the unix socket are the way to say who can use the connection.
There are a lot of SystemD unit configuration files in the contrib/ directory, and this is a mix of new and old (no longer working). 

Please review the unit files and apache config to make them work again: Apache handles external connections (under `/intelmq/...` for API and `/intelmq-manager` for the Manager)  -> proxy them to the socket -> socket is handled by FastAPI. I can help with testing.

We also need to allow administrators to change the Gunicorn configuration in production deployments. There are three ways to do this: 1) mention using the Gunicorn Env variable to do this, 2) let proxy all configs through our CLI (I really don't want to do this), 3) document running web application without our command line

4) API path changes

These are big & breaking changes that shouldn't be done in a mature software without careful consideration. Changing Manager paths is fine for me, but API can be used without it.

5) Lack of documentation - but this should rather go in a separate PR, this one is big enough :)

6) 

> Changes positions.conf file path to /var/lib/intelmq/server/positions.json. This file is hardly ever edited manually by the user, no need to put it in etc.

I support this change, but also have to challenge your assumption: any kind of automated deployment that modifies bots has to modify it - otherwise IntelMQ Manager won't work. So in my setup, the Ansible scripts generate default locations and modify them without any interaction with the manager :)

intelmq/bin/intelmqsetup.py
kamil-certat: After some private discussion: I agree that - as a principle - modifying Apache/nginx in the setup script goes a little beyond the scope of the IntelMQ itself, and it would be right to let the example Apache config be only in the documentation.

As it's breaking change for existing deployments and may require some manual intervention, I'm not sure how to handle it within DEB packages.

intelmq/__init__.py
gethvi: `/etc/nginx/nginx.conf`
`/etc/apache2/apache2.conf`

I find it intuitive, if I just installed it for the first time I would look exactly for this. :smile:

kamil-certat: Okay, don't let user forgot they use IntelMQ :D

intelmq/app/server.py
gethvi: My argument is that while technically the API is publicly available, it is not public per se, currently the endpoints are not documented for the outside use nor do we explicitly say "this is a public API, feel free to use it". As it stands right now it is more of an internal API used for the Manager only. As such I don't think we need to "support" it for other usage.
It's kind of like if you reverse engineer an internal API of some sort, you can't expect the developer to keep you in mind and don't change the API because of you.

gethvi:
> 1. Confusing names
> 
> You have introduced the names "app" and "server" into the code, which I think can be confusing. It looks like a very core part to me (especially with the 'intelmq server start' command - it looks more like starting the botnet to me), but in our case the "very core" are bots, and I'd like to avoid that. Also, the "app" is in one place, then "serve" in others, and they don't map to each other - which makes it harder to find the right part of the code if you don't have the inside knowledge. I spent a long time thinking about how to solve that. What do you think of the name "web"? We can use it instead of "app" and "server" everywhere: in directory structures and commands, and it would fit better and say better what this piece of code is about.

I see. The logic was:
- `app` is an instance of FastAPI Application and it consists of two separate parts (FastAPI routers really): `api` and `webgui` where loading the webgui (old Manager) is optional.
- `server` was ment to mean the Uvicorn/Gunicorn instance of whichever one runs the `app`

But I can see that the code might not fully reflect this logic right now.


> 2. Renaming the IntelMQ Manager
> 
> I see you have introduced the term "WebUI" into the code instead of IntelMQ Manager. It's functionally correct, IntelMQ Manager is the web user interface, but my experience says that every renaming and difference in names used by users and developers makes the onboarding harder for both, and just leads to more time spent on explanations and reviews (I had a few projects struggling with this...). Not to mention the Google/Microsoft culture of renaming tools every month ;). I don't see any benefit in changing the code to webui, it's exactly the same as the established name of IntelMQ Manager. Could we keep the name "Manager" in the code as well?

The thought behind this change is that now the IntelMQ Manager is not a standalone tool anymore, it doesn't really need it's own name because it is just a part of IntelMQ. The user/administrator does not come in contact with the name Manager anymore. The only place where they come in contact with it is in the config which now provides optional boolean `enable_webgui`. From a new users perspective I find this way more self explaining than `enable_manager`.

Anyone slightly experienced with IntelMQ should be able to put two and two together and figure out that `webgui` probably means the only web interface IntelMQ ever had - the Manager.

The website itself will have just IntelMQ in it's title now.

Technically we are not "renaming" the tool, we are merging it to another bigger tool. From now on it is all just "IntelMQ". And as a part of the bigger tools, the code serves the purpose of providing webgui. :)

contrib/app/api/intelmq-api.service
gethvi: Sorry, I forgot to delete this file.

contrib/app/api/intelmq-api.service
gethvi: Also forgot to delete this, it is replaced by `intelmq/debian/systemd/intelmq.service`.

setup.py
gethvi: Yeah this had it's reason because I couldn't persuade python to package dir with static files (javascript, css, etc) in the python package (and I didn't want to add `__init__.py` everywhere).

I followed setuptools docs on this (switch tab to setup.py):
https://setuptools.pypa.io/en/latest/userguide/datafiles.html#subdirectory-for-data-files

which states _"...In this case, the recommended approach is to treat `data` as a namespace package..."_

PR #2435, opened 2023-12-11
FIX: Ensure closing log files on reload

kamil-certat:
During the reloading process, log handlers are not explicitly closed. This may cause long living open file handlers and keeping cached files in the memory. If log files are big, the RAM cache usage can be very high.


kamil-certat:
BTW: just to clearly justify, why it's needed:

On a long-running machine without the change:
```bash
$ lsof -u intelmq | grep /var/log | wc -l
1338

$ lsof -u intelmq | grep /var/log | grep deleted | wc -l
1129
```
More than 1000 of opened file descriptors to deleted log files, hold by IntelMQ.

On the machine with the change, there is not a single one.

sebix (approved):

Issue #2324, opened 2023-02-13
Netlab 360 Hajime feed seems discontinued


gethvi:
https://data.netlab.360.com/hajime/

No more activity, the graphs from the data looks empty and the data URL requires HTTP auth. I think it's safe to assume the feed is dead.

gethvi:
Done.

Issue #2323, opened 2023-02-13
Netlab 360 EK Magnitute feed seems discontinued


gethvi:
https://data.netlab.360.com/ek/

"Last updated" says years ago and the data links require HTTP auth. I think it's safe to assume the feed is dead.

gethvi:
Done.

Issue #2322, opened 2023-02-13
Netlab 360 DGA public feed is discontinued


gethvi:
As per the website https://data.netlab.360.com/dga/:

_After 6 years of providing free service, we regretfully decide to end the free DGA feeds as the current practice is not sustainable. If your business needs commercial license, please contact netlab@360.cn for pricing information. Thanks._

gethvi:
Done.

Issue #2321, opened 2023-02-13
Sucuri feed seems discontinued


gethvi:
The URL for the data http://labs.sucuri.net/?malware redirects to "About" page.

I couldn't find any other URL where the data is accessible. I think it's safe to assume the feed is dead.

gethvi:
Done.

PR #2442, opened 2023-12-14
FIX: Removes discontinued feeds and bots

gethvi:
Solves #2323 #2324 #2322 #2321 #2320


docs/user/feeds.md
```diff
@@ -742,7 +742,7 @@ parameters:
   port: 61614
   provider: CERT.PL
   server: n6stream-new.cert.pl
-  ssl_ca_certificate:
+  ssl_ca_certificate:
```
sebix:
```suggestion
  ssl_ca_certificate:
```
doesn't look intentional

sebix (changes requested):

docs/user/feeds.md
gethvi: Fixed. But it was (re)generated this way with `scripts/generate-feeds-docs.py`.

PR #2449, opened 2024-01-31
FIX: Let modified_pulses_only can be changed

qux-bbb:
The original logic will cause modified_pulses_only is always False.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,8 @@
   - Remove undocumented and unused attributes of `StompCollectorBot` instances:
     `ssl_ca_cert`, `ssl_cl_cert`, `ssl_cl_cert_key`.
   - Minor fixes/improvements and some refactoring (see also above: *Core*...).
+- `intelmq.bots.collectors.alienvault_otx.collector` (PR#2449 by qux-bbb):
+  - Fix modified_pulses_only is always False.
 
 #### Parsers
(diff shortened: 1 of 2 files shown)


kamil-certat:
Hey, thanks for catching that! Could you please add an entry to the Changelog.md with your fix? A test would be also great, but I think it had to be about the whole feature, so I believe we can go without it eventually.

aaronkaplan:
reviewed. Clear logic bug, thx for catching it @qux-bbb . Pulling that in.

aaronkaplan:
@sebix could you please also double check? It's trivial . thx. if all ok, pls press "merge" :) thx

CHANGELOG.md
```diff
@@ -61,8 +61,9 @@
   - Remove undocumented and unused attributes of `StompCollectorBot` instances:
     `ssl_ca_cert`, `ssl_cl_cert`, `ssl_cl_cert_key`.
   - Minor fixes/improvements and some refactoring (see also above: *Core*...).
-- `intelmq.bots.collectors.shadowserver.collector_reports_api`:
-  - The 'json' option is no longer supported as the 'csv' option provides better performance (PR#2372 by elsif2).
+- `intelmq.bots.collectors.alienvault_otx.collector` (PR#2449 by qux-bbb):
```
kamil-certat, line 67: It looks like there is a merging issue removing previous changelog entry. Could you please restore it?

kamil-certat (approved):

PR #2460, opened 2024-02-05
Documentation update for #2455

### LINKED ISSUE #2455: Shadowserver parser docs missing part on feedname
Shadowserver parser docs says:

> A list of possible values of the feedname parameter can be found in the table below in the column "Report Type".

This is no longer true. I understand that the feedname mapping is now gathered from the schema, but still the user doesn't know what feedname he should use for configuration.

If we want to keep the possibility for "fixed report type" we should provide the user with easily accessible values he can use for `feedname` parameter.

elsif2:
Correction to the user documentation for the Shadowserver bots.

### DIFF
--- a/docs/user/bots.md
+++ b/docs/user/bots.md
@@ -930,8 +930,7 @@ oldest files available!).
 
 The resulting reports contain the following special field:
 
-- `extra.file_name`: The name of the downloaded file, with fixed filename extension. The API returns file names with the
-  extension `.csv`, although the files are JSON, not CSV. Therefore, for clarity and better error detection in the parser, the file name in `extra.file_name` uses `.json` as extension.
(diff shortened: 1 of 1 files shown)


gethvi (approved):

docs/user/bots.md
```diff
@@ -2154,7 +2153,7 @@ correct mapping of the columns:
 
 **`feedname`**
 
-(optional, string) Name of the Shadowserver report, see list below for possible values.
+(optional, string) Name of the Shadowserver report. The value for each report type can be found in the schema `feed_name` field.
```
sebix, line 2156: For example using `curl https://interchange.shadowserver.org/intelmq/v1/schema | jq .[].feed_name`?

PR #2461, opened 2024-02-07
FIX: RIPE expert: handle unknown response for non-ASN types

monoidic:
Currently, the RIPE bot successfully handles "No abuse contact found for" messages in response to ASNs, but not for IPs. An example of this would be the address `2001:470:1:c84::12e`, which is from a range assigned by ARIN.
This PR removes the filter on only checking for these types of messages on ASNs, allowing for IPs to not cause errors in the bot.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,8 @@
 - `intelmq.bots.experts.jinja` (PR#2417 by Mikk Margus Möll):
   - Add optional `socket_perms` and `socket_group` parameters to change
     file permissions on socket file, if it is in use.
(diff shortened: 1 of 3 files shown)


kamil-certat:
Thanks, it looks good to me!. Do you think it would be possible to easily add a test for that case? And a short changelog entry?

monoidic:
> Thanks, it looks good to me!. Do you think it would be possible to easily add a test for that case? And a short changelog entry?

Certainly.

aaronkaplan:
Thank you, @monoidic

PR #2469, opened 2024-03-01
FIX: Do not overwrite feed.documentation if present

gethvi:
As per @sebix's suggestion, overwrite is better to set to False on this.

Thanks!


gethvi:
Tests failing because of: `Connection to interchange.shadowserver.org timed out.`

sebix:
Did they block us?

cc @elsif2

aaronkaplan:
as agreed directly, merging now...

PR #2476, opened 2024-03-07
ENH: Added new expert plugin - extra_tags

manuel-subredu:
* added a new plugin that allows manipulation of extra_tags. Main use case
  is to allow usage of same pipeline for both dev and prod setups but be able
  to differentiate between messages and select a different path for your
  dev stage (ex: route dev messages to local directory for examination rather
  then production database)

## Description

ExtraTags expert plugin aims to ease the testing of an existing pipeline and allow
easy routing of the messages in development phase. Main use case:
 1. use extra_tags plugin right after a parser (or input) and mark the messages as
 beloning to development stage ```{'stage': 'dev'}```
 2. insert a prod-dev-routing decision right before final output and with filter
 plugin route non-production messages to various backends to inspection

## Testing
```bash
(.venv) ➜  intelmq git:(develop) ✗ pytest intelmq/tests/bots/experts/extra_tags -s
=========================================================================================================== test session starts ============================================================================================================
platform darwin -- Python 3.12.0, pytest-8.0.2, pluggy-1.4.0
rootdir: ws/extra-tags/intelmq
configfile: pytest.ini
plugins: cov-4.1.0, requests-mock-1.11.0
collected 5 items

intelmq/tests/bots/experts/extra_tags/test_expert.py ...True
..

---------- coverage: platform darwin, python 3.12.0-final-0 ----------
Coverage XML written to file coverage.xml


============================================================================================================ 5 passed in 2.01s =============================================================================================================
(.venv) ➜  intelmq git:(develop)
```

### DIFF
--- a/intelmq/bots/experts/extra_tags/expert.py
+++ b/intelmq/bots/experts/extra_tags/expert.py
@@ -1,0 +1,39 @@
+# SPDX-FileCopyrightText: 2024 Manuel Subredu
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+# -*- coding: utf-8 -*-
+from intelmq.lib.bot import ExpertBot
+
+
+class ExtraTagsExpertBot(ExpertBot):
+    """Add custom extra tags as needed"""
+    overwrite: bool = False
+    tags: dict = None
+
+    def init(self):
+        self.logger.debug("Found {} tags in config section.".format(len(self.tags.keys())))
+
+    def process(self):
+        event = self.receive_message()
+
+        existing_tags = event.get('extra.tags')
+        if existing_tags is None:
+            event.add('extra.tags', self.tags)
+        else:
+            if self.overwrite is False:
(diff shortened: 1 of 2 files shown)


kamil-certat:
Hey, thanks for the commitment! I've looked through your code, and I think it's somehow similar to the `sieve` bot. Have you tried it before? I feel like it may be very near to accomplish what you want, although I'm not sure if it supports manipulating a dict inside a field.

May I ask you to have a look at the sieve bot, and see if it's either already solve the problem, or would be relatively easily possible to extend it? I think it may be nice to extend it as it has much more features to manipulate an event :)

manuel-subredu:
Mea culpa ! I thought that Sieve is somehow related to mail area and didn't looked into the specifics. You are right, sieve expert includes the functionality I actually needed. I think we can drop this pull request. Sorry for the noise.

kamil-certat:
No problem!

PR #2479, opened 2024-03-17
ignore user-generated docs files

sebix:
and remove already committed files from the tree


kamil-certat (approved):

gethvi:
Thanks!

There is one more step necessary: in the github action (`build-docs.yaml`) we need to generate these two files (by running `scripts/generate-event-docs.py` and `scripts/generate-feeds-docs.py`) before building the docs.

This will require `ruamel.yaml` installed.

sebix:
> There is one more step necessary: in the github action (`build-docs.yaml`) we need to generate these two files (by running `scripts/generate-event-docs.py` and `scripts/generate-feeds-docs.py`) before building the docs.

I expected that the docs build would fail if files were missing, but apparently, mkdocs does not complain about that. Thanks for noticing this, I added this step to the workflow file.

kamil-certat (approved):

PR #2488, opened 2024-04-03
Fix dataplane parser failing to parse AS names with pipe characters

DigitalTrustCenter:
At the time of writing, the sshpwauth.txt feed contains an AS name with a pipe character ('|').
AS262907 - BRASIL TECPAR | AMIGO | AVATO

The existing dataplane parser throws an error on that line, as it splits the data fields on just the pipe character.
From the feed header: "Each field is separated by a pipe symbol ('|') and at least two whitespace characters on either side."
Technically there are way more characters other that space that are whitespace characters, but just space seems to be used.
Hence we can split on '  |  ' instead to fix this line breaking.

### DIFF
--- a/intelmq/bots/parsers/dataplane/parser.py
+++ b/intelmq/bots/parsers/dataplane/parser.py
@@ -143,7 +143,10 @@ class DataplaneParserBot(ParserBot):
         else:
             event = self.new_event(report)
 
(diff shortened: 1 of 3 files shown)


sebix:
Thank you for the excellent PR, including the test case :)

PR #2491, opened 2024-04-04
ENH: smtp_batch invalid row hardening

e3rd:
A tiny change, rows missing a should-have-been field time.observation will now be skipped with a warning, instead of a bot error.

### DIFF
--- a/intelmq/bots/outputs/smtp_batch/output.py
+++ b/intelmq/bots/outputs/smtp_batch/output.py
@@ -237,8 +237,7 @@ class SMTPBatchOutputBot(Bot):
                         print("... failed ...", flush=True)
                         continue
                 else:
-                    # visible both warning and print
-                    self.logger.warning(f"Warning: %s timeout, too big to read from redis", mail_record)
+                    self.logger.warning("Warning: timeout, too big to read from redis: %r", mail_record)
                     self.timeout.append(mail_record)
                     continue
 
@@ -256,7 +255,7 @@ class SMTPBatchOutputBot(Bot):
                     if threshold and row["time.observation"][:19] < threshold.isoformat()[:19]:
                         continue
                 except KeyError:
-                    self.logger.warning(f"Warning: %s row skipped due to time.observation error", mail_record)
+                    self.logger.warning("Warning: row skipped due to time.observation error: %r", mail_record)
                 fieldnames = fieldnames | set(row.keys())
                 keys = set(self.allowed_fieldnames).intersection(row)
                 ordered_keys = []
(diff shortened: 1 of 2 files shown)


kamil-certat (approved):

intelmq/bots/outputs/smtp_batch/output.py
```diff
@@ -253,8 +252,11 @@ def prepare_mails(self):
             fieldnames = set()
             rows_output = []
             for row in lines:
-                if threshold and row["time.observation"][:19] < threshold.isoformat()[:19]:
-                    continue
+                try:
+                    if threshold and row["time.observation"][:19] < threshold.isoformat()[:19]:
+                        continue
+                except KeyError:
+                    self.logger.warning(f"Warning: %s row skipped due to time.observation error", mail_record)
```
sebix: I suggest to mention what the error is (missing field). And format it with repr
```suggestion
                    self.logger.warning("Warning: row skipped because it is missing the field 'time.observation': %r", mail_record)
```

intelmq/bots/outputs/smtp_batch/output.py
```diff
@@ -290,7 +292,7 @@ def prepare_mails(self):
                     try:
                         zf.writestr(filename + ".csv", output.getvalue())
                     except Exception:
-                        self.logger.error(f"Cannot zip mail {mail_record}")
+                        self.logger.error(f"Cannot zip mail %s", mail_record)
```
sebix: remove the `f` modifier as above and format the email address (adds parenthesis automatically)

```suggestion
                        self.logger.error("Cannot zip mail %r", mail_record)
```

intelmq/bots/outputs/smtp_batch/output.py
```diff
@@ -238,8 +238,7 @@ def prepare_mails(self):
                         continue
                 else:
                     # visible both warning and print
-                    print(f"Warning: {mail_record} timeout, too big to read from redis", flush=True)
-                    self.logger.warning(f"Warning: {mail_record} timeout, too big to read from redis")
+                    self.logger.warning(f"Warning: %s timeout, too big to read from redis", mail_record)
```
sebix:
```suggestion
                    self.logger.warning("Warning: %s timeout, too big to read from redis", mail_record)
```
remove the f-string here as well

intelmq/bots/outputs/smtp_batch/output.py
```diff
@@ -238,8 +238,7 @@ def prepare_mails(self):
                         continue
                 else:
                     # visible both warning and print
```
sebix: with the print removed, the statement is likely not true any more

sebix (changes requested):

intelmq/bots/outputs/smtp_batch/output.py
e3rd: Right, I've unified it for every logger command

intelmq/bots/outputs/smtp_batch/output.py
e3rd: Thanks, fixed!

intelmq/bots/outputs/smtp_batch/output.py
e3rd: Good point, fixed

sebix (approved):

PR #2494, opened 2024-04-16
FIX: handle init failure on incorrect current event

kamil-certat:
If event hold in current event wasn't saved correctly, the bot is unable to start and handle any event.

In our case, the file wasn't created at all. I'm not sure yet why (I suspect some trouble by reloading), but such a problem shouldn't stop the bot from working.

### DIFF
--- a/intelmq/bots/outputs/misp/output_feed.py
+++ b/intelmq/bots/outputs/misp/output_feed.py
@@ -68,21 +68,25 @@ class MISPFeedOutputBot(OutputBot):
             self.timedelta = datetime.timedelta(minutes=parse_relative(self.interval_event))
 
         if (self.output_dir / '.current').exists():
-            with (self.output_dir / '.current').open() as f:
-                self.current_file = Path(f.read())
-            self.current_event = MISPEvent()
-            self.current_event.load_file(self.current_file)
+            try:
+                with (self.output_dir / '.current').open() as f:
+                    self.current_file = Path(f.read())
+                self.current_event = MISPEvent()
+                self.current_event.load_file(self.current_file)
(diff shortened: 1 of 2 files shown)


intelmq/bots/outputs/misp/output_feed.py
```diff
@@ -68,21 +68,25 @@ def init(self):
...
+                last_max_time = datetime.datetime.strptime(last_max_time, '%Y-%m-%dT%H:%M:%S.%f')
+                if last_max_time < datetime.datetime.now():
+                    self.min_time_current = datetime.datetime.now()
+                    self.max_time_current = self.min_time_current + self.timedelta
+                    self.current_event = None
+                else:
+                    self.min_time_current = last_min_time
+                    self.max_time_current = last_max_time
+            except:
+                self.logger.exception("Loading current event %s failed. Skipping it.", self.current_event)
```
sebix, line 88: Is it really OK to drop events if there was a failure?

kamil-certat: It's not dropping anything, as long as I didn't missed anything. This bot is making a new MISP event periodically, and writing incoming IntelMQ events in the same event. Every time they are saved to the file with the MISP event. The `current_event` is just a reference to the current _MISP_ event, and if we don't load the _current_ (actually: the last one created), the bot creates a new one and save data there. Old data, if they exist, are not lost.

In my case, the pointer to the current MISP event was pointing to not-existing file. When I still need to understand why, as long as we don't have atomic saving to the disk, it may happen (e.g. crash between saving the pointer file and the actual event file)

PR #2499, opened 2024-05-10
Filter expert: Docs fix, treat false as false for filter_regex and add logging

sebix:
-    Docs: fix syntax in filter expert bot documentation
- filter expert: treat filter_regex false as False
  if the parameter
  ```yaml
  filter_regex: false
  ```
  was set, the bot treated this as true
  with this patch, false is treated as false
- filter expert: add debugging output for comparisons
    if users or devs run the bot in DEBUG log level, there was no useful
    output yet.
    This change adds two log statements, showing which comparision was made

### DIFF
--- a/intelmq/bots/experts/filter/expert.py
+++ b/intelmq/bots/experts/filter/expert.py
@@ -148,10 +148,12 @@ class FilterExpertBot(ExpertBot):
             return self.equalsFilter(event, key, condition)
 
     def equalsFilter(self, event, key, value):
+        self.logger.debug('Equality check: %r (event value) == %r (filter value).', event.get(key), value)
         return (key in event and
                 event.get(key) == value)
 
     def regexSearchFilter(self, event, key):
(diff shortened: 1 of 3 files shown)


intelmq/bots/experts/filter/expert.py
```diff
@@ -64,7 +64,7 @@ def init(self):
             self.filter = False
 
         self.regex = False
-        if self.filter_regex is not None:
+        if self.filter_regex is not None and self.filter_regex is not False:
```
kamil-certat: > If this attribute is empty or evaluates to false

In such case, the `if not self.filter_regex` would be better

kamil-certat (approved):

intelmq/bots/experts/filter/expert.py
sebix: yeah, right, that should cover more cases

sebix: ok, no. this now triggers **10** test fails

PR #2509, opened 2024-07-04
ENH: Caching, advanced mapping and separating events for MISP Feed output bot

kamil-certat:
MISP Feed output bot got new features:

 * caching incoming messages to not re-generate on every run;
 * an ability to select which fields should be mapped and set additional parameters for attributes;
 * an ability to group messages in different MISP events, based on a field.

The bot is fully backward-compatible. By default, the previous behaviour is kept.

In addition, code related to Python 3.6 was removed and the message library was fixed not to modify the original dict instance.

This PR replaces on PR #2505.

### DIFF
--- a/docs/dev/bot-development.md
+++ b/docs/dev/bot-development.md
@@ -197,13 +197,25 @@ The `CacheMixin` provides methods to cache values for bots in a Redis database. 
 - `redis_cache_ttl: int = 15`
 - `redis_cache_password: Optional[str] = None`
 
-and provides the methods:
+and provides the methods to cache key-value pairs:
 
 - `cache_exists`
 - `cache_get`
 - `cache_set`
 - `cache_flush`
 - `cache_get_redis_instance`
+
+and following methods to cache objects in a queue:
+
+- `cache_put`
+- `cache_pop`
+- `cache_length`.
+
+Caching key-value pairs and queue caching are two separated mechanisms. The first is designed
+ for arbitrary values, the second one is focused on temporary storing messages (but can handle other
+ data). You won't see caches from one in the another. For example, if adding a key-value pair using
+ `cache_set`, it does not change the value from `cache_length`, and if adding an element using
+ `cache_put` you cannot use `check_exists` to look for it.
 
 ### Pipeline Interactions
 
--- a/docs/dev/bot-development.md
+++ b/docs/dev/bot-development.md
@@ -211,19 +211,18 @@ and following methods to cache objects in a queue:
 - `cache_rpop`
 - `cache_llen`.
 
-Caching key-value pairs and queue caching are two separated mechanisms. The first is designed
- for arbitrary values, the second one is focused on temporary storing messages (but can handle other
- data). You won't see caches from one in the another. For example, if adding a key-value pair using
- `cache_set`, it does not change the value from `cache_llen`, and if adding an element using
- `cache_lpush` you cannot use `check_exists` to look for it.
+Caching key-value pairs and queue caching are two different mechanisms. The functions in the
+ first list are designed for arbitrary values, while the latter ones are primarily for temporarily
+ storing messages, but can also handle other data types. You won't see caches from one in the other.
+ For example, if adding a key-value pair using `cache_set`, it does not change the value from
+ `cache_llen`, and if adding an element using `cache_lpush` you cannot use `check_exists` to look for it.
 
 When using queue-based caching, you have to serialize object to a format accepted by Redis/Valkey
 as the underlying storage. For example, to store a message in a queue using bot ID as key, you can
 use code like:
 
 ```python
-msmessage = self.receive_message().to_dict(jsondict_as_string=True)
-self.cache_lpush(self.bot_id, json.dumps(message))
+self.cache_lpush(self.bot_id, self.receive_message().to_json(jsondict_as_string=True))
 ```
 
 and to retrieve a message from the cache:
--- a/docs/dev/bot-development.md
+++ b/docs/dev/bot-development.md
@@ -207,15 +207,37 @@ and provides the methods to cache key-value pairs:
 
 and following methods to cache objects in a queue:
 
-- `cache_put`
-- `cache_pop`
-- `cache_length`.
+- `cache_lpush`
+- `cache_rpop`
+- `cache_llen`.
 
 Caching key-value pairs and queue caching are two separated mechanisms. The first is designed
  for arbitrary values, the second one is focused on temporary storing messages (but can handle other
  data). You won't see caches from one in the another. For example, if adding a key-value pair using
- `cache_set`, it does not change the value from `cache_length`, and if adding an element using
- `cache_put` you cannot use `check_exists` to look for it.
+ `cache_set`, it does not change the value from `cache_llen`, and if adding an element using
+ `cache_lpush` you cannot use `check_exists` to look for it.
+
+When using queue-based caching, you have to serialize object to a format accepted by Redis/Valkey
+as the underlying storage. For example, to store a message in a queue using bot ID as key, you can
+use code like:
+
+```python
+msmessage = self.receive_message().to_dict(jsondict_as_string=True)
+self.cache_lpush(self.bot_id, json.dumps(message))
+```
+
+and to retrieve a message from the cache:
+
+```python
+data = self.cache_pop()
+if data is None:
+    return # handle empty cache
+message = json.loads(data)
+# to use it as Message object
+message_obj = MessageFactory.from_dict(
+    message, harmonization=self.harmonization, default_type="Event"
+)
+```
 
 ### Pipeline Interactions
 
--- a/docs/user/bots.md
+++ b/docs/user/bots.md
@@ -4794,9 +4794,30 @@ as not usable for IDS.
 
 **`event_separator`
 
-(optional, string): If set to a field name from IntelMQ event, the bot will group incoming messages
-in separated MISP events, based on the value of this field. The `interval_event` parameter acts
-for all grouping events together.
+(optional, string): If set to a field name from IntelMQ event, the bot will work in parallel on a few
+events instead of saving all incomming messages to a one. Each unique value from the field will
+use its own MISP Event. This is useful if your feed provides data about multiple entities you would
+like to group, for example IPs of C2 servers from different botnets. For a given value, the bot will
+use the same MISP Event as long as it's allowed by the `interval_event`.
+
+**`additional_info`
+
+(optional, string): If set, the generated MISP Event will use it in the `info` field of the event,
+in addition to the standard IntelMQ description with the time frame (you cannot remove it as the bot
+depends of datetimes saved there). If you use `event_separator`, you may want to use `{separator}`
+placeholder which will be then replaced with the value of the separator.
+
+For example, the following configuration can be used to create MISP Feed with IPs of C2 servers
+of different botnets, having each botnet in a separated MISP Events with an appropiate description.
+Each MISP Event will contain objects with the `source.ip` field only, and the events' info will look
+like *C2 Servers for botnet-1. IntelMQ event 2024-07-09T14:51:10.825123 - 2024-07-10T14:51:10.825123*
+
+```yaml
+event_separator: malware.name
+additional_info: C2 Servers for {separator}.
+attribute_mapping:
+  source.ip:
+```
 
 **Usage in MISP**
 
--- a/docs/user/bots.md
+++ b/docs/user/bots.md
@@ -4762,7 +4762,7 @@ hour", string.
 
 **`bulk_save_count`**
 
-(optional, int) If set to a non-0 value, the bot won't refresh the MISP feed immeadiately, but will cache
+(optional, int) If set to a non-0 value, the bot won't refresh the MISP feed immediately, but will cache
 incoming messages until the given number of them. Use it if your bot proceeds a high number of messages
 and constant saving to the disk is a problem. Reloading or restarting bot as well as generating
 a new MISP event based on `interval_event` triggers saving regardless of the cache size.
--- a/docs/user/bots.md
+++ b/docs/user/bots.md
@@ -4760,6 +4760,13 @@ The PyMISP library >= 2.4.119.1 is required, see
 () The output bot creates one event per each interval, all data in this time frame is part of this event. Default "1
 hour", string.
 
+**`bulk_save_count`**
+
+(optional, int) If set to a non-0 value, the bot won't refresh the MISP feed immeadiately, but will cache
+incoming messages until the given number of them. Use it if your bot proceeds a high number of messages
+and constant saving to the disk is a problem. Reloading or restarting bot as well as generating
+a new MISP event based on `interval_event` triggers saving regardless of the cache size.
+
 **Usage in MISP**
 
 Configure the destination directory of this feed as feed in MISP, either as local location, or served via a web server.
--- a/docs/user/bots.md
+++ b/docs/user/bots.md
@@ -4835,7 +4835,7 @@ attribute_mapping:
 The keys refers to which MISP events you want to tag. If you want to tag all of them, use `__all__`.
 If you use `event_separator` and want to add additional tags to some events, use the expected values
 of the separation field. The *list of dicts* defines MISP tags as parameters to create `MISPTag`
-objects from. Each dictonary has to have at least `name`. For all available parameters refer to the
+objects from. Each dictionary has to have at least `name`. For all available parameters refer to the
 [PyMISP documentation](https://pymisp.readthedocs.io/en/latest/_modules/pymisp/abstract.html#MISPTag)
 for `MISPTag`.
 
--- a/docs/user/bots.md
+++ b/docs/user/bots.md
@@ -4736,6 +4736,12 @@ Create a directory layout in the MISP Feed format.
 The PyMISP library >= 2.4.119.1 is required, see
 [REQUIREMENTS.txt](https://github.com/certtools/intelmq/blob/master/intelmq/bots/outputs/misp/REQUIREMENTS.txt).
 
+Note: please test the produced feed before using in production. This bot allows you to do an
+extensive customisation of the MISP feed, including creating multiple events and tags, but it can
+be tricky to configure properly. Misconfiguration can prevent bot from starting or have bad
+consequences for your MISP Instance (e.g. spaming with events). Use `intelmqctl check` command
+to validate your configuration against common mistakes.
+
 **Module:** `intelmq.bots.outputs.misp.output_feed`
 
 **Parameters:**
@@ -4765,7 +4771,7 @@ hour", string.
 (optional, int) If set to a non-0 value, the bot won't refresh the MISP feed immediately, but will cache
 incoming messages until the given number of them. Use it if your bot proceeds a high number of messages
 and constant saving to the disk is a problem. Reloading or restarting bot as well as generating
-a new MISP event based on `interval_event` triggers saving regardless of the cache size.
+a new MISP event based on `interval_event` triggers regenerating MISP feed regardless of the cache size.
 
 **`attribute_mapping`**
 
@@ -4775,6 +4781,10 @@ first-level key represents an IntelMQ field that will be directly translated to 
 dictionary represents additional parameters PyMISP can take when creating an attribute. They can use
 names of other IntelMQ fields (then the value of such field will be used), or static values. If not needed,
 leave empty dict.
+
+For available attribute parameters, refer to the
+[PyMISP documentation](https://pymisp.readthedocs.io/en/latest/_modules/pymisp/mispevent.html#MISPObjectAttribute)
+for the `MISPObjectAttribute`.
 
 For example:
 
@@ -4817,6 +4827,40 @@ event_separator: malware.name
 additional_info: C2 Servers for {separator}.
 attribute_mapping:
   source.ip:
+```
+
+**`tagging`
+
+(optional, dict): Allows setting MISP tags to MISP events. The structure is a *dict of list of dicts*.
+The keys refers to which MISP events you want to tag. If you want to tag all of them, use `__all__`.
+If you use `event_separator` and want to add additional tags to some events, use the expected values
+of the separation field. The *list of dicts* defines MISP tags as parameters to create `MISPTag`
+objects from. Each dictonary has to have at least `name`. For all available parameters refer to the
+[PyMISP documentation](https://pymisp.readthedocs.io/en/latest/_modules/pymisp/abstract.html#MISPTag)
+for `MISPTag`.
+
+Note: setting `name` is enough for MISP to match a correct tag from the global collection. You may
+see it lacking the colour in the MISP Feed view, but it will be retriven after importing to your
+instance.
+
+Example 1 - set two tags for every MISP event:
+
+```yaml
+tagging:
+  __all__:
+    - name: tlp:red
+    - name: source:intelmq
+```
+
+Example 2 - create separated events based on `malware.name` and set additional family tag:
+
+```yaml
+event_separator: malware.name
+tagging:
+  __all__:
+    - name: tlp:red
+  njrat:
+    - name: njrat
 ```
 
 **Usage in MISP**
--- a/docs/user/bots.md
+++ b/docs/user/bots.md
@@ -4866,7 +4866,7 @@ tagging:
 
 ** `flat_events`
 
-(optional, bool): instead of creating an object for every incomming IntelMQ message, it will add
+(optional, bool): instead of creating an object for every incoming IntelMQ message, it will add
 attributes directly to the MISP event. Useful if your want to export just a list of data, e.g.
 C2 domains, without having to group some attributes together. By default set to `False`.
 
--- a/docs/user/bots.md
+++ b/docs/user/bots.md
@@ -4770,9 +4770,9 @@ a new MISP event based on `interval_event` triggers saving regardless of the cac
 **`attribute_mapping`**
 
 (optional, dict) If set, allows selecting which IntelMQ event fields are mapped to MISP attributes
-as well as attribute parameters (like e.g. a comment). The expected format is a *dictonary of dictionaries*:
+as well as attribute parameters (like e.g. a comment). The expected format is a *dictionary of dictionaries*:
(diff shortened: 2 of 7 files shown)


kamil-certat:
Yeah, finally green :tada:

sebix:
Are you using it in production?

kamil-certat:
We started using it on staging, found some pain points, and now I'll test it & hopefully promote to prod in a few days - so, not yet, but soon ;)

sebix:
There's not much I can check here without reading the MISP code and docs and setting up an instance.

Maybe @Rafiot, can you have a glimpse?

docs/user/bots.md
```diff
@@ -4595,6 +4595,44 @@ The PyMISP library >= 2.4.119.1 is required, see
 () The output bot creates one event per each interval, all data in this time frame is part of this event. Default "1
 hour", string.
```
aaronkaplan, line 4767: what happens when set to 0 ?

docs/user/bots.md
```diff
@@ -4595,6 +4595,44 @@ The PyMISP library >= 2.4.119.1 is required, see
...
+  feed.name:
+    comment: event_description.text
+  destination.ip:
+    to_ids: False
+```
+
+would create a MISP object with three attributes `source.ip`, `feed.name` and `destination.ip`
+and set their values as in the IntelMQ event. In addition, the `feed.name` would have a comment
+as given in the `event_description.text` from IntelMQ event, and `destination.ip` would be set
+as not usable for IDS.
```
aaronkaplan: clear and understandable. Cool

docs/user/bots.md
```diff
@@ -4595,6 +4595,44 @@ The PyMISP library >= 2.4.119.1 is required, see
...
+
+**`event_separator`
+
+(optional, string): If set to a field name from IntelMQ event, the bot will group incoming messages
+in separated MISP events, based on the value of this field. The `interval_event` parameter acts
+for all grouping events together.
```
aaronkaplan: this is not 100% clear for me yet:

will it collect events until (... until when?...) and then sort and group by this viel?
I think rephrasing this slightly makes it clearer in the documentaiton.

intelmq/bots/outputs/misp/output_feed.py
```diff
@@ -5,37 +5,40 @@
...
     """Generate an output in the MISP Feed format"""
+
     interval_event: str = "1 hour"
+    bulk_save_count: int = None
     misp_org_name = None
     misp_org_uuid = None
-    output_dir: str = "/opt/intelmq/var/lib/bots/mispfeed-output"  # TODO: should be path
+    output_dir: str = (
+        "/opt/intelmq/var/lib/bots/mispfeed-output"  # TODO: should be path
+    )
```
aaronkaplan: why the tuple?

intelmq/bots/outputs/misp/output_feed.py
```diff
@@ -65,80 +64,201 @@ def init(self):
...
+            # For proper handling of nested fields
+            message_obj = MessageFactory.from_dict(
+                message, harmonization=self.harmonization, default_type="Event"
+            )
+            key = message_obj.get(self.event_separator) or DEFAULT_KEY
+
+        if key in self.current_events:
+            event = self.current_events[key]
+        else:
+            event = self._generate_new_event(key)
```
aaronkaplan: don't we want to support at add a default tag to an event, such as "FEED:Source=intelmq" ?
Of course, this should be configured by the user.

intelmq/bots/outputs/misp/output_feed.py
```diff
@@ -65,80 +64,201 @@ def init(self):
...
+                )
+                self.current_events = {}
+
+        if not self.current_files or self.max_time_current < datetime.datetime.now():
             self.min_time_current = datetime.datetime.now()
             self.max_time_current = self.min_time_current + self.timedelta
+            self.current_events = {}
 
-    def process(self):
+    def _load_event(self, file_path: Path, key: str):
```
aaronkaplan, line 159: missing function docstring ?

intelmq/bots/outputs/misp/output_feed.py
```diff
@@ -65,80 +64,201 @@ def init(self):
...
+            # Always create the first event so we can keep track of the interval.
+            # It also ensures cleaning the queue after startup in case of awaiting
+            # messages from the previous run
+            self._generate_feed()
+        elif cache_size >= self.bulk_save_count:
+            self._generate_feed()
+
+        self.acknowledge_message()
+
+    def _generate_new_event(self, key):
```
aaronkaplan: missing function docstring?

intelmq/bots/outputs/misp/output_feed.py
```diff
@@ -65,80 +64,201 @@ def init(self):
...
+            self.output_dir / f"{self.current_events[key].uuid}.json"
+        )
+        with (self.output_dir / ".current").open("w") as f:
+            if not self.event_separator:
+                f.write(str(self.current_files[key]))
+            else:
+                json.dump({k: str(v) for k, v in self.current_files.items()}, f)
+        return self.current_events[key]
+
+    def _add_message_to_feed(self, message: dict):
```
aaronkaplan: missing func docstring?

intelmq/bots/outputs/misp/output_feed.py
```diff
@@ -65,80 +64,201 @@ def init(self):
...
+        else:
+            event = self._generate_new_event(key)
+
+        obj = event.add_object(name="intelmq_event")
+        if not self.attribute_mapping:
+            self._default_mapping(obj, message)
+        else:
+            self._custom_mapping(obj, message)
+
+    def _default_mapping(self, obj: "MISPObject", message: dict):
```
aaronkaplan: docstring?

intelmq/bots/outputs/misp/output_feed.py
```diff
@@ -65,80 +64,201 @@ def init(self):
...
             except NewAttributeError:
                 # This entry isn't listed in the harmonization file, ignoring.
-                pass
+                if object_relation != "__type":
+                    self.logger.warning(
+                        "Object relation %s not exists in MISP definition, ignoring",
+                        object_relation,
+                    )
+
+    def _extract_misp_attribute_kwargs(self, message: dict, definition: dict) -> dict:
```
aaronkaplan, line 275: docstring? not just # comment

intelmq/bots/outputs/misp/output_feed.py
```diff
@@ -65,80 +64,201 @@ def init(self):
...
+                value in self.harmonization["event"] or
+                value.split(".", 1)[0] in self.harmonization["event"]
+            ):
+                result[parameter] = message.get(value)
+            else:
+                result[parameter] = value
+        return result
 
-        feed_output = self.current_event.to_feed(with_meta=False)
+    def _custom_mapping(self, obj: "MISPObject", message: dict):
```
aaronkaplan: same

intelmq/bots/outputs/misp/output_feed.py
```diff
@@ -65,80 +64,201 @@ def init(self):
...
+                obj.add_attribute(
+                    object_relation,
+                    value=message[object_relation],
+                    **self._extract_misp_attribute_kwargs(message, definition),
+                )
+                # In case of manual mapping, we want to fail if it produces incorrect values
 
-        with self.current_file.open('w') as f:
-            json.dump(feed_output, f)
+    def _generate_feed(self, message: dict = None):
```
aaronkaplan: same

intelmq/lib/mixins/cache.py
```diff
@@ -1,17 +1,29 @@
...
 """
 
+import json
 from typing import Any, Optional
 import redis
 import intelmq.lib.utils as utils
 
 
 class CacheMixin:
+    """Provides caching possibilities for bots
```
aaronkaplan: quick line (link to documentation on the mixins) ?

aaronkaplan (dismissed):
minor changes requested pls.

aaronkaplan:
My main request is:

1) give the possibility (or document it) how to add default tags to events (such as "FEED:Source=intelmq") to any event which gets generated via this feed. Adding a default tag helps in sorting out automatically added events and attributes in case something goes wrong.
2) make sure the users understand in the documentation how feeds work. Linking to the MISP documentation and what each parameter when including the MISP feed actually means.
For example, did you know that if  you set "caching:enabled" in a MISP feed, then that means the feed events and attributes don't get added to the `events` (and `attributes`) table in MISP. 
It's totally non-intuitive. 
Why am I saying this? I had to figure it out the hard way myself.
3) basically we need a *tutorial* on docs.intelmq.org  on how to run this properly!
It's a very cool feature and addition but you can quickly shoot yourself in the foot if you don't know how it works and what every parameter in MISP means.

intelmq/bots/outputs/misp/output_feed.py
kamil-certat: It's not a tuple ;) It's a Python way to put the string into next line (or split into multiple strings) to avoid line-length limit - although it's my fault. The rules we enforce in the repository are very weak, so weak that my almost every time wants tot reformat a lot, and sometimes I forget reverting changes not related to my code.

BTW, I think more and more that we would profit from enforcing black/ruff instead of weak pycodestyle.

intelmq/bots/outputs/misp/output_feed.py
kamil-certat: It's an interesting request, could you please elaborate how do you see it? I was focused on what we need now (so I didn't include a few different things I was thinking about), but if you have a specific idea you think it would be important to include now - let's share.

intelmq/bots/outputs/misp/output_feed.py
kamil-certat: Not at all. If you wish to document `_load_event` with `This function loads an event from file`, feel free, but I disagree with enforcing a docstring requirement for every function. It should happen when the function is complex, otherwise it's useless. Over-commenting is not a good practice. If there is anything you don't understand in the function, please let me know.

intelmq/bots/outputs/misp/output_feed.py
kamil-certat: Disagree it's needed here.

intelmq/bots/outputs/misp/output_feed.py
kamil-certat: Disagree it's needed here.

intelmq/bots/outputs/misp/output_feed.py
kamil-certat: Disagree it's needed here.

intelmq/bots/outputs/misp/output_feed.py
kamil-certat: Disagree it's needed here.

intelmq/bots/outputs/misp/output_feed.py
kamil-certat: Disagree it's needed here.

kamil-certat:
Thanks for the review!

re 1: It's a good idea, I'll look into it. I was concentrated on what we actually need, and I do think there are more things we can improve in the bot.
re 2: It's a good point, I will point to the MISP documentation. I also had to explore what we really can do, the PyMISP documentation wasn't really helpful :D 
re 3: yeah, I agree it needs a tutorial (and I've planned it), but it's a separated thing, I won't include it here. My plan is to write a blog post once we establish the final integration internally, and then we could include it in the documentation as well. I expect to publish it around September, hopefully. There is a queue of what I'd like to write about, and a limited time ;)

docs/user/bots.md
kamil-certat: What do you think about the current version?

> will it collect events until (... until when?...) and then sort and group by this viel?

This field is responsible only for separating using the field. Collecting & holding is a different setting, that can be used together, but you don't have to. Let's have a look at the new description

docs/user/bots.md
kamil-certat: I've looked in the code, and it looks like in such case, this would be a ValueError. However, you can use "0 hours" what should result in creating a new MISP event for every IntelMQ event.

kamil-certat:
@aaronkaplan 

Could you have a look again? I have implemented tagging as well as rewritten the documentation, added config examples, implemented validation in the `check` method and refactored code, including renaming methods to be more clear what they do or adding doc strings where I found it useful. I have also added descriptions to complex test cases explaining what should happen.

I do admit that the configuration of the bot is complex. I did my best to marriage flexibility and readability, but I think in the future we may eventually redesign this configuration, based on feedback. I'm also open for any suggestions.

intelmq/bots/outputs/misp/output_feed.py
aaronkaplan: I am also using black more and more...

sebix:
@aaronkaplan you requested some changes to this PR and @kamil-certat implemented various changes based on your feedback. Can you please re-check?

sebix:
@aaronkaplan this PR is waiting for your approval

aaronkaplan:
Still need to test against a misp server

sebix:
rebased on develop. there were quite some conflicts to fix.

sebix:
now the JSON parser test fails, the output differs to the expected one.

The PR contains changes in `intelmq/lib/message.py`. I don't know yet if this is a rebase error or an error in the code.

sebix:
> Still need to test against a misp server

As discussed yesterday, I'll take this over

intelmq/lib/mixins/cache.py
```diff
@@ -51,6 +67,20 @@ def cache_set(self, key: str, value: Any, ttl: Optional[int] = None):
         if self.redis_cache_ttl:
             self.__redis.expire(key, self.redis_cache_ttl)
 
+    def cache_put(self, value: dict) -> int:
+        # Returns the length of the list after pushing
+        size = self.__redis.lpush(self.bot_id, json.dumps(value))
+        return size
```
sebix:
```suggestion
    def cache_lpush(self, key: str, value: Any) -> int:
        # Returns the length of the list after pushing
        size = self.__redis.lpush(key, value)
        return size
```

I appreciate the new function, but why is the name of the list not a parameter, but fixed?
Also, why is the value a dict?

The new functions are also missing in the docs in `docs/dev/bot-development.md`

intelmq/lib/mixins/cache.py
```diff
@@ -51,6 +67,20 @@ def cache_set(self, key: str, value: Any, ttl: Optional[int] = None):
...
+
+    def cache_length(self) -> int:
+        return self.__redis.llen(self.bot_id)
```
sebix:
```suggestion
    def cache_llen(self, key: str) -> int:
        return self.__redis.llen(key)
```

The new function is missing in the docs in `docs/dev/bot-development.md`.

intelmq/lib/mixins/cache.py
```diff
@@ -51,6 +67,20 @@ def cache_set(self, key: str, value: Any, ttl: Optional[int] = None):
...
+
+    def cache_pop(self) -> dict:
+        data = self.__redis.rpop(self.bot_id)
+        if data is None:
+            return None
+        return json.loads(data)
```
sebix:
```suggestion
    def cache_rpop(self, key: str) -> dict:
        data = self.__redis.rpop(key)
        return data
```

The new function is missing in the docs in `docs/dev/bot-development.md`.

docs/user/bots.md
```diff
@@ -4760,6 +4766,103 @@ The PyMISP library >= 2.4.119.1 is required, see
...
+as given in the `event_description.text` from IntelMQ event, and `destination.ip` would be set
+as not usable for IDS.
+
+**`event_separator`
+
+(optional, string): If set to a field name from IntelMQ event, the bot will work in parallel on a few
+events instead of saving all incoming messages to a one. Each unique value from the field will
+use its own MISP Event. This is useful if your feed provides data about multiple entities you would
+like to group, for example IPs of C2 servers from different botnets. For a given value, the bot will
+use the same MISP Event as long as it's allowed by the `interval_event`.
```
sebix: I think this parameter name is misleading and unintuitive. The name "separator" is most often used to specify the element that separates joined data.

From the description, I gather it is actually a grouping or aggregation key.

docs/user/bots.md
```diff
@@ -4760,6 +4766,103 @@ The PyMISP library >= 2.4.119.1 is required, see
...
+For example:
+
+```yaml
+attribute_mapping:
+  source.ip:
+  feed.name:
+    comment: event_description.text
+  destination.ip:
+    to_ids: False
+```
```
sebix, line 4798: An empty value causes an exception:
```
  File "intelmq/bots/outputs/misp/output_feed.py", line 262, in _extract_misp_attribute_kwargs
    for parameter, value in definition.items():
                            ^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'items'
```

```suggestion
```yaml
attribute_mapping:
  source.ip: {}
  feed.name:
    comment: event_description.text
  destination.ip:
    to_ids: False
```

sebix (changes requested):
Found problems in the docs and in the changes of the cache mixin

kamil-certat:
I'm working on a few improvements / additional requirements right now, a new code should come today or tomorrow. I'll then address comments

sebix:
Documentation for the new parameter `flat_events` is missing

intelmq/lib/mixins/cache.py
kamil-certat: The idea was to create a message caching for the bot which would make it easy to use, completely abstract from the underlying implementation, not a simple wrapper over the Redis (see e.g. `cache_set`). This is why they do not follow exact the same naming as Redis library, and do additional things.

I have now mixed feelings in which direction it should go. I see two general options:
* renaming my new methods to something more clear, like `cache_dict`, `get_cached_dict`
* go with the simple wrapping over Redis library

I both cases, I would consider also making a key optional (and using bot ID as the default). If going with the first option, then I'd also handle transforming messages to the JSON and back. 

@sebix what do you think is better?

docs/user/bots.md
kamil-certat: `grouping_key` sounds much better, I'll change it

kamil-certat:
I'm aware of that, I'll adjust documentation tomorrow

kamil-certat:
I also have to add matching unit tests ;)

intelmq/lib/mixins/cache.py
sebix: > * renaming my new methods to something more clear, like `cache_dict`, `get_cached_dict`

These names do not indicate in any way that a FIFO queue is used, but instead make the impression that a dictionary is the cache.
What about `queue_push` and `queue_pop`?

>     * go with the simple wrapping over Redis library

It would be more generic and slightly lower-level. (However, redis/valkey are not that low-level anyway :) )

> I both cases, I would consider also making a key optional (and using bot ID as the default). If going with the first option, then I'd also handle transforming messages to the JSON and back.

I understand your desire for simplicity, but I have doubts about ambiguity. With an optional key, some possible interfaces are:
- `func(key: Optional[str], value: Any)`: Requires calling it with `None` as the first parameter: `func(None, foobar)`. No big difference to using `self.bot_id`: `func(self.bot_id, foobar)`
- `func(value: Any, key: Optional[str] = None)`: The order of parameters is the complete opposite of anything else in IntelMQ and elsewhere.
- `func(key_or_value: Any, value_if_key: Optional[Any])` allowing to call `func(value)` and `func(key, value)`: Too complicated. Not very KISSy :kiss:

kamil-certat: > I understand your desire for simplicity, but I have doubts about ambiguity. With an optional key, some possible interfaces are:

I'd never ever thought about option 3, and the option 1 is not simplifying anything :sweat_smile: So the 2 is the only one I would ever consider

But as my original idea seems to introduce too many questions, I'll move forward with your original idea of wrapping Redis library and let the developer use it flexible.

kamil-certat:
BTW, all methods were documented in bot-development :thinking:

sebix:
> BTW, all methods were documented in bot-development 🤔

Oh, sorry. I don't know where I was looking.

docs/user/bots.md
```diff
@@ -4772,6 +4772,7 @@ hour", string.
 incoming messages until the given number of them. Use it if your bot proceeds a high number of messages
 and constant saving to the disk is a problem. Reloading or restarting bot as well as generating
 a new MISP event based on `interval_event` triggers regenerating MISP feed regardless of the cache size.
+To ensure saving on reload without any delay, you should also set `_sighup_delay` parameter.
```
sebix: That's not a parameter, but an internal variable

docs/user/bots.md
```diff
@@ -4760,6 +4766,146 @@ The PyMISP library >= 2.4.119.1 is required, see
 () The output bot creates one event per each interval, all data in this time frame is part of this event. Default "1
 hour", string.
 
+**`bulk_save_count`**
+
+(optional, int) If set to a non-0 value, the bot won't refresh the MISP feed immediately, but will cache
+incoming messages until the given number of them. Use it if your bot proceeds a high number of messages
+and constant saving to the disk is a problem. Reloading or restarting bot as well as generating
+a new MISP event based on `interval_event` triggers regenerating MISP feed regardless of the cache size.
+To ensure saving on reload without any delay, you should also set `_sighup_delay` internal variable.
```
sebix: What do you want to say with this last sentence?
`_sighup_delay` is not a parameter. It's a switch that developers can use to indicate bot-specific behavior. It is true by default.

Why does the user need to do anything here?

docs/user/bots.md
```diff
@@ -4760,6 +4766,146 @@ The PyMISP library >= 2.4.119.1 is required, see
...
+    source.fqdn:
+      comment: malware.name
+      type: domain
+      category: "Network activity"
+      to_ids: true
+  tagging:
+    __all__:
+      - name: tlp:amber
+  # ensure saving on reload
+  _sighup_delay: false
```
sebix: Here, you do the opposite of what is written above, disabling it instead of activating it.

And still, this is not something a user should need to do. If the bot needs to disable this flag (maybe depending on other parameters), do it in `init()`.

docs/dev/bot-development.md
```diff
@@ -197,14 +197,48 @@ The `CacheMixin` provides methods to cache values for bots in a Redis database.
...
 
+and following methods to cache objects in a queue:
+
+- `cache_lpush`
+- `cache_rpop`
+- `cache_llen`.
+
+Caching key-value pairs and queue caching are two separated mechanisms. The first is designed
+ for arbitrary values, the second one is focused on temporary storing messages (but can handle other
+ data). You won't see caches from one in the another. For example, if adding a key-value pair using
```
sebix:
```suggestion
Caching key-value pairs and queue caching are two different mechanisms. The functions in the first list are designed
 for arbitrary values, while the latter ones are primarily for temporarily storing messages, but can also handle other data
 types. You won't see caches from one in the other. For example, if adding a key-value pair using
```

docs/dev/bot-development.md
```diff
@@ -197,14 +197,48 @@ The `CacheMixin` provides methods to cache values for bots in a Redis database.
...
+ `cache_set`, it does not change the value from `cache_llen`, and if adding an element using
+ `cache_lpush` you cannot use `check_exists` to look for it.
+
+When using queue-based caching, you have to serialize object to a format accepted by Redis/Valkey
+as the underlying storage. For example, to store a message in a queue using bot ID as key, you can
+use code like:
+
+```python
+msmessage = self.receive_message().to_dict(jsondict_as_string=True)
+self.cache_lpush(self.bot_id, json.dumps(message))
```
sebix: Why not use `message.to_json` directly?
```suggestion
self.cache_lpush(self.bot_id, self.receive_message().to_json(jsondict_as_string=True))
```

sebix: Or: `Message.serialize`

docs/dev/bot-development.md
```diff
@@ -197,14 +197,48 @@ The `CacheMixin` provides methods to cache values for bots in a Redis database.
...
+
+```python
+data = self.cache_pop()
+if data is None:
+    return # handle empty cache
+message = json.loads(data)
+# to use it as Message object
+message_obj = MessageFactory.from_dict(
+    message, harmonization=self.harmonization, default_type="Event"
+)
```
sebix, line 238: Untested:

```suggestion
message_obj = MessageFactory.unserialize(
    data, harmonization=self.harmonization, default_type="Event"
)
```
I think the two optional parameters are unnecessary too.

sebix (changes requested):

docs/user/bots.md
kamil-certat: This is a good idea, I'll do it so in `init()`. I used wrong wording in the docs. The thing is: if you keep the `_sighup_delay` enabled, then after receiving the reloading signal, the bot waits for the next incoming message to actually do the reloading. This means, the accumulated (cached) events won't be saved. This can be very unfortunate delay

docs/dev/bot-development.md
kamil-certat: I'm almost sure the default_type was necessary, but I'll check it again.

docs/user/bots.md
kamil-certat: Actually, I'm stupid, I have already solved it by setting the right defaults... But I'll do it conditionally if using bulk saving

docs/dev/bot-development.md
kamil-certat: Using the first. I have copied the code from the bot, where the JSON representation is used later.

docs/dev/bot-development.md
kamil-certat: The first is recommended, the second is necessary to match the dumping code.

1) if the harmonization is not provided, the code will _read it from the file_, and not use any already loaded version. Reading files and processing JSON is in Python quite expensive operation, it's much better to provide the ready harmonization.
2) the second depends on setting `with_type` when serializing the message. As I didn't set it on serialization, I have to provide type on deserialization. I could do it the other way around, there is no difference - but one or another has to be set.

sebix: Ad 1: True, we should probably cache that file -> different topic
Ad 2: `MessageFactory` is the preferred way of convering Messages to JSON back and forth. I agree, that both ways need to be compatible with each other. Let's refactor this doc section later when improving the overall docs.

sebix (approved):

PR #2513, opened 2024-07-09
ENH: Add ability to copy more fields from the report

kamil-certat:
Sometimes collectors sets more context information, e.g. the source file name. Currently, parsers usually remove those non-standard information. With this change, we let the user decide to copy more information to the event.

In addition, the currently copied fields were documented.

Real use case: the HTTP collector sets filename when extracting data from an archive. This filename contains an information we want to include in the final event, which we cannot retrieve in any other way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,8 +32,8 @@
 - `intelmq.bots.parsers.shadowserver._config`:
   - Fetch schema before first run (PR#2482 by elsif2, fixes #2480).
 - `intelmq.bots.parsers.dataplane.parser`: Use `  |  ` as field delimiter, fix parsing of AS names including `|` (PR#2488 by DigitalTrustCenter).
-- all parsers: add `copy_custom_fields` parameter allowing copying additional fields from the report, e.g. `extra.file_name`.
-  (PR# by Kamil Mankowski).
+- all parsers: add `copy_collector_provided_fields` parameter allowing copying additional fields from the report, e.g. `extra.file_name`.
+  (PR#2513 by Kamil Mankowski).
 
 #### Experts
 - `intelmq.bots.experts.sieve.expert`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@
 - `intelmq.bots.parsers.shadowserver._config`:
   - Fetch schema before first run (PR#2482 by elsif2, fixes #2480).
 - `intelmq.bots.parsers.dataplane.parser`: Use `  |  ` as field delimiter, fix parsing of AS names including `|` (PR#2488 by DigitalTrustCenter).
+- all parsers: add `copy_custom_fields` parameter allowing copying additional fields from the report, e.g. `extra.file_name`.
+  (PR# by Kamil Mankowski).
 
 #### Experts
 - `intelmq.bots.experts.sieve.expert`:
--- a/intelmq/lib/bot.py
+++ b/intelmq/lib/bot.py
@@ -1082,6 +1082,7 @@ class ParserBot(Bot):
     _default_message_type = 'Report'
 
     default_fields: Optional[dict] = {}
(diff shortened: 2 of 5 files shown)


intelmq/lib/bot.py
```diff
@@ -1245,6 +1246,13 @@ def process(self):
                         for key, value in self.default_fields.items():
                             event.add(key, value, overwrite=False)
 
+                if self.copy_custom_fields:
+                    for key in self.copy_custom_fields:
+                        if key not in report:
+                            continue
+                        for event in events:
+                            event.add(key, report.get(key), overwrite=False)
+
```
sebix: When this code is in `intelmq.lib.bot.ParserBot.process`, it will not be active in all parsers as the above documentation suggests.

Adding it to `intelmq.lib.message.Report.__init__` (and passing the parameter in `intelmq.lib.bot.Bot.new_event` will cover all cases AFAIK.

sebix (changes requested):
I'm not convinced by the parameter name. Custom fields sound very generic and is ambiguous, also it lacks any mention of _report_ or _event_.

What about `replicate_report_fields(_to_event)`? It's longer, but it speaks by itself.

CHANGELOG.md
```diff
@@ -32,6 +32,8 @@
 - `intelmq.bots.parsers.shadowserver._config`:
   - Fetch schema before first run (PR#2482 by elsif2, fixes #2480).
 - `intelmq.bots.parsers.dataplane.parser`: Use `  |  ` as field delimiter, fix parsing of AS names including `|` (PR#2488 by DigitalTrustCenter).
+- all parsers: add `copy_custom_fields` parameter allowing copying additional fields from the report, e.g. `extra.file_name`.
+  (PR# by Kamil Mankowski).
```
sebix:
```suggestion
- `intelmq.lib.bot.ParserBot`: add `copy_custom_fields` parameter allowing copying additional fields from the report, e.g. `extra.file_name` (PR#2513 by Kamil Mankowski).
```

intelmq/lib/bot.py
kamil-certat: Hmm, you mean some parsers do not use the `process`? Then they do not use the `default_fields`, right?

kamil-certat:
@sebix I'm open to name changes, but hmm, I'm not convinced with your suggestion either. Actually, my original idea was `copy_fields` (as direct similarity to `default_fields`) or `copy_report_fields`. However, I realised it would then create a confusion that those are _all_ copied (not parsed) fields, which is not true - creating an event from the report has a hardcoded list of fields to replicate. This is how `additional` came to play :D

What would you say about:
- `extended_report_fields`
- `copy_additional_report_fields`
- `replicate_extra_report_fields` (although this is too narrow, with a different harmonisation it could be more than `extra.*`)
- `extendend_report_fields_to_copy` (actually, this suggests we want a list, while `copy_...` sounds like Y/N question)

...or any other? I think that the last sounds currently the best

intelmq/lib/bot.py
sebix: `process` can be overwritten

kamil-certat: Of course, it could be, but I didn't know it's a broad practice - I've looked at the source code, and you're right, a lot of parsers rely on overriding the whole `process`. I'll move the code to the `new_event`, and I think we should consider moving support of `default_fields` there as well as this is something I'd expect all parsers to support (but it's a separated thing).

sebix:
Some remarks from @th-certbund:

sebix:
Some remarks by @th-certbund:
- "custom" is also very generic
- "report" is ambiguous too
- What about "copy_collector_provided_fields" or similar?

aaronkaplan (approved):
looks good, I read it and it's understandable.

aaronkaplan:
thank you @kamil-certat !!

Issue #2486, opened 2024-03-21
Allow sieve reuse data from the event


kamil-certat:
Currently, only modify bot can access data from the event and set them back in other field. It would be useful to let Sieve bot do the same, e.g. to rewrite some fields.

PR #2514, opened 2024-07-15
ENH: Sieve variable support

monoidic:
This fixes issue #2486

### DIFF
--- a/intelmq/bots/experts/sieve/expert.py
+++ b/intelmq/bots/experts/sieve/expert.py
@@ -14,6 +14,7 @@ import os
 import re
 import traceback
 import operator
+import json
 
 from datetime import datetime, timedelta, timezone
 from typing import Callable, Dict, Optional, Union
@@ -24,7 +25,7 @@ from intelmq import HARMONIZATION_CONF_FILE
 from intelmq.lib import utils
 from intelmq.lib.bot import ExpertBot
 from intelmq.lib.exceptions import MissingDependencyError
-from intelmq.lib.message import Event
+from intelmq.lib.message import Message
 from intelmq.lib.utils import parse_relative
 from intelmq.lib.harmonization import DateTime
 
@@ -65,7 +66,9 @@ class SieveExpertBot(ExpertBot):
             harmonization_config = utils.load_configuration(HARMONIZATION_CONF_FILE)
             SieveExpertBot._harmonization = harmonization_config["event"]
 
-        self.sieve = SieveModel(self.file, self.logger)
+        self.metamodel = self.init_metamodel()
+        self.model = self.read_sieve_file(self.file, self.metamodel)
+        self.variables = {}
 
     @staticmethod
     def check(parameters):
@@ -105,7 +108,7 @@ class SieveExpertBot(ExpertBot):
 
     def process(self) -> None:
         event = self.receive_message()
-        procedure = self.sieve.process(event)
+        procedure = self.model_process(event)
 
         # forwarding decision
         if procedure == Procedure.DROP:
@@ -122,14 +125,6 @@ class SieveExpertBot(ExpertBot):
             self.send_message(event, path=path)
 
         self.acknowledge_message()
-
-
-class SieveModel:
-    def __init__(self, file_path: str, logger) -> None:
-        self.metamodel = self.init_metamodel()
-        self.sieve = self.read_sieve_file(file_path, self.metamodel)
-        self.logger = logger
-        self.variables = {}
 
     _string_op_map = {
         "==": operator.eq,
@@ -181,9 +176,9 @@ class SieveModel:
         str,
         Callable[
             [
-                "SieveModel",
+                "SieveExpertBot",
                 object,
-                Event,
+                Message,
             ],
             bool,
         ],
@@ -245,11 +240,11 @@ class SieveModel:
             # apply custom validation rules
             metamodel.register_obj_processors(
                 {
-                    "SingleStringMatch": SieveModel.validate_string_match,
-                    "MultiStringMatch": SieveModel.validate_string_match,
-                    "SingleNumericMatch": SieveModel.validate_numeric_match,
-                    "MultiNumericMatch": SieveModel.validate_numeric_match,
-                    "SingleIpRange": SieveModel.validate_ip_range,
+                    "SingleStringMatch": SieveExpertBot.validate_string_match,
+                    "MultiStringMatch": SieveExpertBot.validate_string_match,
+                    "SingleNumericMatch": SieveExpertBot.validate_numeric_match,
+                    "MultiNumericMatch": SieveExpertBot.validate_numeric_match,
+                    "SingleIpRange": SieveExpertBot.validate_ip_range,
                 }
             )
 
@@ -274,13 +269,13 @@ class SieveModel:
                 f"Could not parse sieve file {filename!r}, error in ({e.line}, {e.col}): {e}"
             )
 
-    def process(self, event):
+    def model_process(self, event):
         procedure = Procedure.CONTINUE
         self.variables.clear()
-        if not self.sieve:  # empty rules file results in empty string
+        if not self.model:  # empty rules file results in empty string
             return procedure
 
-        for statement in self.sieve.statements:
+        for statement in self.model.statements:
             procedure = self.process_statement(statement, event)
             if procedure == Procedure.KEEP:
                 self.logger.debug(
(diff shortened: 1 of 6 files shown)


sebix:
Wow, very neat again!

@kamil-certat as you have a use-case of this feature, do you want to have a look at the diff?

intelmq/bots/experts/sieve/expert.py
```diff
@@ -41,219 +40,309 @@
...
-        'Expression': lambda self, match, event: self.match_expression(match, event),
+    _date_op_map = {":before": operator.lt, ":after": operator.gt}
+
+    _cond_map: Dict[
+        str,
+        Callable[
+            [
+                "SieveModel",
+                object,
+                Event,
```
kamil-certat: NIT: it could be any _Message_. Although it's rare, it is possible to process Report object as well

intelmq/bots/experts/sieve/expert.py
```diff
@@ -41,219 +40,309 @@
...
+        for path in paths:
+            self.send_message(event, path=path)
+
+        self.acknowledge_message()
+
+
+class SieveModel:
+    def __init__(self, file_path: str, logger) -> None:
+        self.metamodel = self.init_metamodel()
+        self.sieve = self.read_sieve_file(file_path, self.metamodel)
```
kamil-certat: NIT: I'd ask to rename `sieve` here or in the bot class - just for sake of confusing, those are two variables with the same name, relatively close to each other, and with a different meaning. Maybe name it in the bot as `sieve_model`?

intelmq/bots/experts/sieve/expert.py
```diff
@@ -263,48 +352,40 @@ def process_conjunction(self, conj, event) -> bool:
 
     def process_condition(self, cond, event) -> bool:
         name = cond.match.__class__.__name__
-        return cond.neg ^ self._cond_map[name](self, cond.match, event)
+        ret = self._cond_map[name](self, cond.match, event)
+        if cond.neg:
+            ret = not ret
+        return ret
```
kamil-certat: Did you decide to rewrite because of readability, or there is a change I don't see? For sake of readability, I'd think about `return not ret if cond.neg else ret`. In cases like this I find it a little clearer to keep this in a one line. But it's personal preference.

intelmq/bots/experts/sieve/expert.py
```diff
@@ -263,48 +352,40 @@ def process_conjunction(self, conj, event) -> bool:
...
-            return op in {'!=', '!~'}
+            return op in {"!=", "!~"}
 
-        lhs = event[key]
-        if not isinstance(lhs, str) and op not in ('==', '!='):
-            if isinstance(lhs, dict):
-                lhs = dumps(lhs)
-            else:
-                lhs = str(lhs)
+        return self._string_op_map[op](event[key], value)
```
kamil-certat: I'd ask for a short explanation why:
* `process_single_string_match`
* `process_multi_string_match`
* `process_multi_numeric_match`
were changed. 

The `value(s).value(s)` -> `value(s)` change looks clear, it's been moved upper to the operator maps, but I'm unsure about dropped conversion and special cases for numeric match, where is it handled now? (I might have missed it)

intelmq/bots/experts/sieve/expert.py
```diff
@@ -419,33 +511,33 @@ def validate_ip_range(ip_range) -> None:
...
         """
-        valid_types = {'Integer', 'Float', 'Accuracy', 'ASN'}
-        position = SieveExpertBot.get_linecol(num_match.value, as_dict=True)
+        valid_types = {"Integer", "Float", "Accuracy", "ASN"}
+        position = SieveModel.get_linecol(num_match.value, as_dict=True)
 
         # validate harmonization type (event key)
         try:
-            type = SieveExpertBot._harmonization[num_match.key]['type']
+            type = SieveExpertBot._harmonization[num_match.key]["type"]
```
kamil-certat, line 532: I'd think about explicit setting the harmonisation in the `SieveModel` class. Now it's a little strange that the `SieveModel` class, just used by `SieveExpertBot`, then silently have a backref to `SieveExpertBot` - and thus to a value that by default is `None`. If you try to draw how classes depend on each other, I would be surprised seeing a circle between classes not inheriting on each other.

intelmq/tests/bots/experts/sieve/test_expert.py
```diff
@@ -1719,33 +1854,41 @@ def test_mixed_if_and_actions(self):
...
-        expected['extra.list_empty'] = True
+        expected["extra.list_empty"] = True
 
         self.input_message = event
         self.run_bot()
         self.assertMessageEqual(0, expected)
 
-    def test_extra_dict(self):
-        self.sysconfig['file'] = os.path.join(os.path.dirname(__file__), 'test_sieve_files/test_extra_dict.sieve')
+    def test_variables(self):
```
kamil-certat, line 1888: (!) What has happened to the`test_extra_dict` test case?

kamil-certat (changes requested):
I love the change, as well as the refactoring and formatting against black/ruff. I have some NIT comments and one or two clarification questions, but also a one big one: what has happened to the test `test_extra_dict`? Have you found it duplicated (if so, please remove test files and say what duplicates it)?

Please also update the bot documentation to describe variables as well as the changelog.

And as a discussion, I have two things I'm unsure, but I don't find it blocking:
1) a variable can be in place of typed values, like `Bool`. Shouldn't we then ensure in code that the variable has the correct value?
2) variables look like extremely promising for the future, but I also ask myself - wouldn't it be also nice just allow syntax like `ifd.field1 = idf.field2`? The current use case looks like only supporting this.
3) I'd think about extending the validation - I haven't seen if there is any place where we would try to statically validate that the requested variable was previously defined

intelmq/bots/experts/sieve/expert.py
monoidic: The removed code in `process_multi_numeric_match` was after an unconditional `return` and hence completely dead code.
The rest of the changes are in order to uniformly add `resolve_value` to all match functions via `_cond_map`, to allow for the use of variables in those matches.

The changes in `process_single_string_match` appear to have been due to me starting development from an older commit and accidentally discarding the newer commits.

intelmq/bots/experts/sieve/expert.py
monoidic: Perhaps splitting the bot in two was not a great idea. I re-merged them, though kept the previous `SieveModel.process` separate from `SieveExpertBot.process` as `model_process`

intelmq/tests/bots/experts/sieve/test_expert.py
monoidic: Accidentally discarded the commi which added this.

monoidic:
> what has happened to the test test_extra_dict?

Accidentally discarded newer commits to the Sieve bot. I noticed it in the commit diff ~a week ago, but got distracted and forgot to address it...

> a variable can be in place of typed values, like Bool. Shouldn't we then ensure in code that the variable has the correct value?

This has actually been addressed in the code. The `resolve_value` method which resolves variables (if needed) has a parameter, `expected_types`, expecting a type or tuple thereof. If specified, it raises an exception if the variable's type is not as expected.
Or do you mean that an expected type could somehow be specified in the Sieve file itself as well?

> variables look like extremely promising for the future, but I also ask myself - wouldn't it be also nice just allow syntax like `ifd.field1 = idf.field2`? The current use case looks like only supporting this.

At the moment, yes.
The parsing end would technically accept both `append extra.list $elem` as well as `append $list $elem` at the moment (with the former appending a value given in a variable to a list in the event, and the latter working purely with variables), though the processing end does not handle variables as keys yet.
Allowing event keys to be used like that could be an option, but seems like a slight downgrade feature-wise (e.g any temporary variables which are not meant to end up in the event itself would have to exist under random keys of the event and be cleared out afterwards) while requiring all the same new validation that's been added for variables anyway.
Though for some cases, it would be more ergonomical.

> I'd think about extending the validation - I haven't seen if there is any place where we would try to statically validate that the requested variable was previously defined

I'd expect `check()` to be the place for this.

kamil-certat (approved):
Thanks for the explanation! It sounds sufficient for me.

Before it gets merged, I'd ask you to verify / add a test that `check()` would reject a sieve file with variable used without being declared beforehand

sebix:
Merging this, based on @kamil-certat's positive review.

For the last request on a check for undeclared variables, I opened https://github.com/certtools/intelmq/issues/2526

PR #2523, opened 2024-08-16
Shadowserver: Fix parameter 'reports' behaviour if empty string

### LINKED ISSUE #2521: shadowserver documentation: if reports is "", things don't work
It was confusing for a user to have "reports": "" in the config but then he would not get any reports.
Instead the reports key must not be present at all, to get all reports. That's confusing and might have to be better explained in the documentation.

sebix:
if the parameter reports is an empty string (or only contains whitespace), behave as if the parameter is not set and select all reports

fixes #2521


aaronkaplan (approved):
thank you!

PR #2531, opened 2024-10-15
Fix to avoid schema download if not configured #2530.

### LINKED ISSUE #2530: ShadowServer parser attempts to download schema even if not configured
The schema update call is included in the standard IntelMQ's crontab file in deb distribution. However, the parser does not check if there is any ShadowServer bot configured in the runtime. In a controlled environment it causes unnecessary alerts/errors.

The parser should follow the behaviour of other bots and check if there is any configured parser before attempting to download new schema.

elsif2:
The code has been updated to check for configured bots before running the schema update.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@
 #### Parsers
 - `intelmq.bots.parsers.shadowserver._config`:
   - fix error message formatting if schema file is absent (PR#2528 by Sebastian Wagner).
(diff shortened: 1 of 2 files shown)


sebix:
Look OK for me.

@kamil-certat can you test it and then merge?

kamil-certat (approved):
After testing on our instance, it looks solving the problem, thank you!

PR #2532, opened 2024-10-18
use PyMISP, ExpandedPyMISP is deprecated

rvyhnal:
MISP related bots try to import the deprecated ExpandedPyMISP instead of PyMISP which produces the deprecation message in logs. 
As ExpandedPyMISP was introduced for the transition to Python3 quite some time ago, please consider merging this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 #### Collectors
 - `intelmq.bots.collectors.shadowserver.collector_reports_api.py`:
   - Fixed behaviour if parameter `types` value is empty string, behave the same way as not set, not like no type.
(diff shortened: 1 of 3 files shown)


sebix:
Sorry for the late reaction; I missed the PR.

What is the minimum required pymisp version now? Are the `REQUIREMENTS.txt` in both files' directories still correct?

@Rafiot (author of both bots) can you please check if that's correct?

Rafiot:
I'd suggest to bump the requirements to the most recent release (they have wheels, and are for sure supporting python 3.9 to 3.13) but it's not required as the API hasn't changed.

Regarding the changes PyMISPExpanded -> PyMISP, yes, it is correct.

sebix:
> Regarding the changes PyMISPExpanded -> PyMISP, yes, it is correct.

Thank you!

PR #2536, opened 2024-11-20
Fixed behaviour if parameter `types` value is empty string, behave th…

elsif2:
This fixes the issue of the _types_ list being empty when the runtime config is generated with the web manager:

```
   parameters:
     types: ''
```

A similar fix was made for _reports_ previously.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@
 
 ### Bots
 #### Collectors
+- `intelmq.bots.collectors.shadowserver.collector_reports_api.py`:
+  - Fixed behaviour if parameter `types` value is empty string, behave the same way as not set, not like no type.
 
 #### Parsers
 - `intelmq.bots.parsers.shadowserver._config`:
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,8 @@
 
 ### Bots
 #### Collectors
+- `intelmq.bots.collectors.shadowserver.collector_reports_api.py`:
+  - Fixed behaviour if parameter `types` value is empty string, behave the same way as not set, not like no type.
```
sebix, line 23:
```suggestion
  - Fixed behaviour if parameter `types` value is empty string, behave the same way as not set, not like no type (PR#2536 by Jason).
```

sebix (approved):

sebix:
Thanks for the fix.

Unfortunately, the actual problem is much bigger than that of single parameters and requires a generic solution. Multiple users are falling into this trap regularly and it hurts the usability of IntelMQ massively.

kamil-certat (approved):

Issue #2515, opened 2024-07-16
Introduce strong formatting & linting aka black/ruff


kamil-certat:
IntelMQ has a very weak formatting/linting rules what is not a typical thing in modern development, where projects tend to provide strict & automated formatting to keep the code style consistent.

Especially in Python world, there is a general movement last years to implement the [black](https://github.com/psf/black) formatter, and currently most black-compatible [ruff](https://github.com/astral-sh/ruff) formatter, which also replace some other tools, like isort.

I'd like to introduce ruff with import sorting enabled in IntelMQ. This would provide a consistent code style without much developer action required as black/ruff enforces the style automatically. The code style is compliant with PEP8, and in addition lowers the number of git conflicts, removes unused imports, and can be fully configured to support any of our style configurations.

PR #2540, opened 2024-11-30
Makefile: add pycodestyle

sebix:


kamil-certat (approved):
What do you think about eventually adding stronger formatter (aka https://github.com/certtools/intelmq/issues/2515) already?

sebix:
> What do you think about eventually adding stronger formatter (aka #2515) already?

I'm not yet sure about the implications, I would need to have a look at it.

PR #2551, opened 2025-01-13
contrib: remove prettyprint

sebix:
was introduced by b0447aea601cc066cbbc3dc3a11e2bea609f3bdb long ago

better use the command line tool `jq` for that purpose, it's more powerful


aaronkaplan:
Well, I still do like the script. Even though it can just internally call `jq` now...

kamil-certat:
You can still have the script copied, it's just not really usable in the repository :)

kamil-certat (approved):

sebix:
> You can still have the script copied, it's just not really usable in the repository :)

@aaronkaplan What's your take on that?

aaronkaplan (changes requested):
My take on this is that we should document the jq command (and link to the jq command for those who don't know it)

Then we can safely delete the script.

sebix:
Added it to the FAQ page. If we have a tutorial in the future, it can be added there too.

PR #2552, opened 2025-01-13
contrib eventdb: new to_json script

sebix:
Extract data from the IntelMQ EventDB and convert it to JSON to use it with intelmqctl

### DIFF
--- a/contrib/eventdb/to_json.py
+++ b/contrib/eventdb/to_json.py
@@ -1,0 +1,52 @@
+#!/usr/bin/python3
+
+# SPDX-FileCopyrightText: 2025 Institute for Common Good Technology
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+from argparse import ArgumentParser
+from datetime import datetime
+import json
+from sys import exit, stderr
+from pprint import pprint
+
+from psycopg2 import connect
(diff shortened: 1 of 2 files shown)


contrib/eventdb/to_json.py
```diff
@@ -0,0 +1,48 @@
...
+import json
+from sys import exit, stderr
+from pprint import pprint
+
+from psycopg2 import connect
+from psycopg2.extras import RealDictCursor
+
+try:
+    with open('/etc/intelmq/eventdb-serve.conf') as fody_config:
+        conninfo = json.load(fody_config)['libpg conninfo']
```
kamil-certat: This is not a standard IntelMQ path nor the format (but a Fody one), could you please make it a configurable parameter as well as a syntax example, so one could use it without the Fody in the same environment? E.g. in my setup Fody is configured in a container, separated from the main IntelMQ itself.

kamil-certat (changes requested):

contrib/eventdb/to_json.py
sebix: I added a parameter `--dsn` to pass a standard libpg dsn string that takes precedence over the fody configuration, which now serves as a fallback.

kamil-certat (approved):

PR #2570, opened 2025-02-26
doc: fix parameter description shadowserver api

sebix:
The parameter `reports` is not required according to the source code

unset parameters means all reports are processed

please verify @elsif2

### DIFF
--- a/docs/user/bots.md
+++ b/docs/user/bots.md
@@ -960,7 +960,7 @@ The resulting reports contain the following special field:
 
 **`reports`**
 
-(required, string/array of strings) An array of strings (or a list of comma-separated values) of the mailing lists you want to process.
+(optional, string/array of strings) An array of strings (or a list of comma-separated values) of the mailing lists you want to process.
 
 **`types`**
 


elsif2:
That is correct.

sebix:
Is the current state, not the proposed change, correct?

elsif2:
Yes, it is the current state.

sebix:
Interesting, then I don't understand the code.

Here, the default for `report` is None:
https://github.com/certtools/intelmq/blob/adb2d156d642ea4b804e26b12e81d1b0d21ec425/intelmq/bots/collectors/shadowserver/collector_reports_api.py#L43
and the default for `_report_list` is an empty list:
https://github.com/certtools/intelmq/blob/adb2d156d642ea4b804e26b12e81d1b0d21ec425/intelmq/bots/collectors/shadowserver/collector_reports_api.py#L50
This sets `_report_list` if `report` is set:
https://github.com/certtools/intelmq/blob/adb2d156d642ea4b804e26b12e81d1b0d21ec425/intelmq/bots/collectors/shadowserver/collector_reports_api.py#L59-L64
Otherwise, `_report_list` is left at the empty list.

In `_report_list`, `_report_list` is used and an empty list is handled:
https://github.com/certtools/intelmq/blob/adb2d156d642ea4b804e26b12e81d1b0d21ec425/intelmq/bots/collectors/shadowserver/collector_reports_api.py#L104-L105

What requires `reports` then?

elsif2:
It isn't required. 

> The parameter reports is not required according to the source code

Is correct.

sebix:
Okay, but why do the docs then need to say "required"?

elsif2:
I think there is a misunderstanding.  I was agreeing with your change.

sebix:
> I think there is a misunderstanding. I was agreeing with your change.

Okay, then let's merge it :)

PR #2572, opened 2025-03-03
Update pycodestyle

kamil-certat:
Newer Ubuntu runners have pycodestyle in the version that supports Python 3.8+ operators


sebix (approved):

kamil-certat:
I think this would have to be force-merged as the expected jobs seem not to be triggered with changes just in the CI file

sebix:
> I think this would have to be force-merged as the expected jobs seem not to be triggered with changes just in the CI file

@aaronkaplan Can you please do that?

PR #2579, opened 2025-03-05
lib/test: show warnings and errors in case of mismatch

sebix:
Show the warnings and errors of a bot if the amount does not match the expected number


sebix:
Hm, the tests of 9 bots are failing now

kamil-certat:
It looks like at least some of the failing tests are about negative test cases

sebix:
> It looks like at least some of the failing tests are about negative test cases

It was actually just bad copy-pasting :grimacing:

PR #2598, opened 2025-04-11
intelmqctl: On stop wait longer for bots to be stopped

### LINKED ISSUE #2595: intelmqctl stop bots are still running
Hi,

working more with intelmq these days I noticed when executing `intelmqctl stop` sometimes some bots are still reported as running afterwards in the output (not that big of an issue) and the exit-code is `!= 1` (bigger issue, since my wrapper script (using systemd for restarting and most important for starting when booting the server) reacts on this).

I noticed when running `intelmqctl status` after the `intelmqctl stop`, the bots actually are reported as stopped. Looking deeper into the code responsible for stopping the bots, I noticed intelmq(ctl) uses the following prodecure for stopping the whole botnet:
1. Iterate over the bots in the botnet (https://github.com/certtools/intelmq/blob/aadc8870d95f3d690ae60982b2ea01600daf2a54/intelmq/bin/intelmqctl.py#L563-L564)
2. Send them the `SIGTERM` signal (https://github.com/certtools/intelmq/blob/aadc8870d95f3d690ae60982b2ea01600daf2a54/intelmq/lib/processmanager.py#L197-L199)
3. Wait for `0.75` Seconds (https://github.com/certtools/intelmq/blob/aadc8870d95f3d690ae60982b2ea01600daf2a54/intelmq/bin/intelmqctl.py#L567)
4. Check if the bots are running (https://github.com/certtools/intelmq/blob/aadc8870d95f3d690ae60982b2ea01600daf2a54/intelmq/bin/intelmqctl.py#L568-L571) -> determines the exit code

So to me it looks like on our server it takes too long until all the bots are finally stopped (when executing `intelmqctl status` the bots are stopped after all). In our case we're speeking about 16 bots on a server with 4 GiB RAM and 2 cores (not that impressive specs, but so far we're not dealing with massive amounts of data and half of the bots are really just for testing purposes).

With this in mind, does my analysis make sense to you (as people knowing intelmq much better than I do)?

So far my approach would be simply increasing the time `intelmqctl stop` sleeps until checking on the bots (not generally, but adding this as a parameter to the CLI). Am I missing a simpler solution here?

Lukas-Heindl:
fixes certtools/intelmq#2595

retry multiple times on `intelmqctl stop` to check if bots really stopped, since the bots might take longer to stop. Using retry in constrast to increasing the sleep_time keeps the delay short in case the bots did already stop.

Note: I intend to make at least the amount of retries performed configurable so that if the current total `sleep_time` of 4.75 seconds does not suffice for some users they are able to increase it.

### DIFF
--- a/intelmq/bin/intelmqctl.py
+++ b/intelmq/bin/intelmqctl.py
@@ -567,12 +567,14 @@ Get some debugging output on the settings and the environment (to be extended):
         # only aliasing the list to ease reading the following
         stopped_but_still_running_bots = bots
 
-        # parameters:
+        retries = getattr(self._parameters, 'stop_retry_limit', 5)
+
+        # parameters (default):
         # - sleep 0.75 s with an increment of 0.1
         # - at most 5 tries
         # => sleep-ing at most 4.75 seconds
         sleep_time = 0.75 # in seconds
-        for _ in range(5):
+        for _ in range(retries):
             # give the bots some time to terminate
             time.sleep(sleep_time)
             # update the botnet_status
--- a/intelmq/bin/intelmqctl.py
+++ b/intelmq/bin/intelmqctl.py
@@ -584,7 +584,7 @@ Get some debugging output on the settings and the environment (to be extended):
                     stopped_but_still_running_bots.remove(bot_id)
 
             # check if all bots are stopped -> no need to wait further
-            if len(stopped_but_still_running_bots) == 0:
+            if not stopped_but_still_running_bots:
                 break
             # the longer the bots need to terminate the longer we wait to check
             # again to avoid long-term load on the system
--- a/intelmq/bin/intelmqctl.py
+++ b/intelmq/bin/intelmqctl.py
@@ -563,12 +563,36 @@ Get some debugging output on the settings and the environment (to be extended):
         for bot_id in bots:
             self.bot_stop(bot_id, getstatus=False)
 
-        retval = 0
-        time.sleep(0.75)
-        for bot_id in bots:
-            botnet_status[bot_id] = self.bot_status(bot_id)[1]
-            if botnet_status[bot_id] not in ['stopped', 'disabled']:
-                retval = 1
+        # shallow copy of the list suffices
+        # only aliasing the list to ease reading the following
+        stopped_but_still_running_bots = bots
+
+        # parameters:
+        # - sleep 0.75 s with an increment of 0.1
+        # - at most 5 tries
+        # => sleep-ing at most 4.75 seconds
+        sleep_time = 0.75 # in seconds
+        for _ in range(5):
+            # give the bots some time to terminate
+            time.sleep(sleep_time)
+            # update the botnet_status
+            for bot_id in stopped_but_still_running_bots:
+                botnet_status[bot_id] = self.bot_status(bot_id)[1]
+                if botnet_status[bot_id] in ['stopped', 'disabled']:
+                    stopped_but_still_running_bots.remove(bot_id)
+
+            # check if all bots are stopped -> no need to wait further
+            if len(stopped_but_still_running_bots) == 0:
+                break
+            # the longer the bots need to terminate the longer we wait to check
+            # again to avoid long-term load on the system
+            # but stop at 5 seconds to avoid waiting too long until rechecking
+            # the status
+            sleep_time = min(5, sleep_time+0.1)
+
+        retval = 1
+        if len(stopped_but_still_running_bots) == 0:
+            retval = 0
 
         self.log_botnet_message('stopped', group)
         return retval, botnet_status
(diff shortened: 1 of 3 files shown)


sebix:
the 3.10 pytest fail is unrelated (see #2602)

Lukas-Heindl:
I put quite some comments in there. Looking at the other code, I see much less comments. If I should shrink the comments just let me know.

sebix:
> If I should shrink the comments just let me know.

Please not, inline comments are great :)

Lukas-Heindl:
Alright then. The tests (that shall pass) pass. Any additional feedback? (also notify if not, so I can squash the commits into one commit)

Lukas-Heindl:
Is there something still blocking the merge? (please don't feel pressured, I just want to avoid this got lost in the whole fixing the pytest pipelines issue)

Lukas-Heindl:
Friendly reminder for this PR.

sebix:
> Is there something still blocking the merge? (please don't feel pressured, I just want to avoid this got lost in the whole fixing the pytest pipelines issue)

A code review and testing (+ changelog entry, but that's minor). Most of us do that work honorary, so it's sometimes hard to find the time.

Lukas-Heindl:
Alright 👍

I know finding times in the free time for such things is always hard. Personally I also always say contributing to OSS is not supposed to make your live more stressfull, instead it is supposed to be fun. So I didn't want to bother you, it's really mostly about avoiding this PR getting lost (I know that edge is sometimes sharp)

intelmq/bin/intelmqctl.py
```diff
@@ -563,12 +563,38 @@ def botnet_stop(self, group=None):
...
+        # => sleep-ing at most 4.75 seconds
+        sleep_time = 0.75  # in seconds
+        for _ in range(retries):
+            # give the bots some time to terminate
+            time.sleep(sleep_time)
+            # update the botnet_status
+            for bot_id in stopped_but_still_running_bots:
+                botnet_status[bot_id] = self.bot_status(bot_id)[1]
+                if botnet_status[bot_id] in ['stopped', 'disabled']:
+                    stopped_but_still_running_bots.remove(bot_id)
```
sebix: You iterate over a list while modifying it. That's not working in the way you might expect:

```
>>> d = list(range(5))
>>> for x in d:
...     print(x)
...     if x == 2:
...         d.remove(x)
...         
0
1
2
4
```

That skips an element.

You need to iterate over a copy.

intelmq/bin/intelmqctl.py
```diff
@@ -563,12 +563,38 @@ def botnet_stop(self, group=None):
...
+
+            # check if all bots are stopped -> no need to wait further
+            if len(stopped_but_still_running_bots) == 0:
```
sebix:
```suggestion
            if not stopped_but_still_running_bots:
```

sebix (changes requested):

intelmq/bin/intelmqctl.py
Lukas-Heindl: Oh that's a very good catch 👍

What do you think about using `filter()` like I did in caf75c34302ba2412a8cd85f1b8c2d86bcaf49c7?

It should (kinda by design) avoid this mistake.

sebix: A list comprehension would be much shorter and more readable.

Lukas-Heindl: Oh I agree that's much nicer (and also saves the `list(` call after applying `filter(`) -> see 3d1d5e5567bfe44e8bfcd15e45cd5b3b406a2305 (btw I'd then squash some commits after the final review but before the merge)

sebix (changes requested):
The code looks OK now, thanks.

But your PR introduces a bug:

Expected behaviour:
```
intelmq@localhost:~> intelmqctl -t json stop 
{
    "deduplicator-expert": "disabled",
    "feodo-tracker-browse-collector": "disabled",
    "feodo-tracker-browse-parser": "stopped",
    "file-output": "stopped",
    "postgres-output": "stopped",
    "spamhaus-drop-collector": "disabled",
    "spamhaus-drop-parser": "stopped",
    "taxonomy-expert": "stopped"
}
```
Current behaviour:
```
intelmq@localhost:~> intelmqctl -t json stop 
{}
```

Lukas-Heindl:
Oh yes. In the last iteration I removed the update to `botnet_status` accidentally. It is back in so the issue should be fixed now.

(I'm only updating the `bot_id`s in `stopped_but_still_running_bots` as the others are not expected to change in the meantime (as far as I thought this through). But we can easily iterate over all bots for this update.)

sebix:
rebased on develop and added a changelog entry

sebix:
Thank you for your contribution and especially for your patience!

The patch will be included in the next IntelMQ release coming end of week.

Lukas-Heindl:
Thanks for reviewing and improving the quality of the contribution.

Having the changes so soon (with the new release, so without installing from some other branch or so) is also quite nice 👍

PR #2606, opened 2025-04-23
asn, fake expert: default paths, asn: add checks

### LINKED ISSUE #2566: ASN Lookup expert: Missing sane default value for database path
https://github.com/certtools/intelmq/blob/adb2d156d642ea4b804e26b12e81d1b0d21ec425/intelmq/bots/experts/asn_lookup/expert.py#L28

I suggest that the default path is `/var/lib/intelmq/bots/asn_lookup/ipasn.dat` as was the case some time ago.

Also, the update script could check first if the path exists (and isn't a None or empty string) and even if it is writeable before downloading the data.

sebix:
asn, fake expert: use state path for database default
asn: Check for database file existence and writability fixes #2566

### DIFF
--- a/intelmq/bots/experts/asn_lookup/expert.py
+++ b/intelmq/bots/experts/asn_lookup/expert.py
@@ -8,8 +8,8 @@ import io
 import re
 import sys
 import bz2
-import pathlib
 import requests
+from pathlib import Path
 
 from intelmq import VAR_STATE_PATH
 from intelmq.lib.bot import ExpertBot
@@ -41,6 +41,8 @@ class ASNLookupExpertBot(ExpertBot):
             self.logger.error("Read 'bots/experts/asn_lookup/README' and "
                               "follow the procedure.")
             self.stop()
+        if not Path(self.database).is_file:
+            raise ValueError('Database file does not exist or is not a file.')
 
     def process(self):
         event = self.receive_message()
@@ -113,6 +115,12 @@ class ASNLookupExpertBot(ExpertBot):
         if pyasn is None:
             raise MissingDependencyError("pyasn")
 
+        for database_path in set(bots.values()):
+            if not Path(database_path).is_file:
+                raise ValueError('Database file does not exist or is not a file.')
+            elif not os.access(database_path, os.W_OK):
+                raise ValueError('Database file is not writeable.')
+
         try:
             if verbose:
                 print("Searching for the latest database update...")
@@ -160,7 +168,7 @@ class ASNLookupExpertBot(ExpertBot):
             prefixes = pyasn.mrtx.parse_mrt_file(archive, print_progress=False, skip_record_on_error=True)
 
         for database_path in set(bots.values()):
-            database_dir = pathlib.Path(database_path).parent
+            database_dir = Path(database_path).parent
             database_dir.mkdir(parents=True, exist_ok=True)
             pyasn.mrtx.dump_prefixes_to_file(prefixes, database_path)
 
--- a/intelmq/bots/experts/asn_lookup/expert.py
+++ b/intelmq/bots/experts/asn_lookup/expert.py
@@ -36,12 +36,10 @@ class ASNLookupExpertBot(ExpertBot):
         try:
             self._database = pyasn.pyasn(self.database)
         except OSError:
-            self.logger.error("pyasn data file does not exist or could not be "
-                              "accessed in %r.", self.database)
-            self.logger.error("Read 'bots/experts/asn_lookup/README' and "
-                              "follow the procedure.")
+            raise ValueError(f"pyasn data file does not exist or could not be accessed in {self.database!r}. ",
+                             "Please see https://docs.intelmq.org/latest/user/bots/#asn-lookup")
             self.stop()
-        if not Path(self.database).is_file:
+        if not Path(self.database).is_file():
             raise ValueError('Database file does not exist or is not a file.')
 
     def process(self):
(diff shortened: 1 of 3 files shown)


sebix:
rebased to fix the tests

intelmq/bots/experts/asn_lookup/expert.py
```diff
@@ -40,6 +41,8 @@ def init(self):
             self.logger.error("Read 'bots/experts/asn_lookup/README' and "
                               "follow the procedure.")
             self.stop()
+        if not Path(self.database).is_file:
```
kamil-certat: An idea - maybe add also a check in the `check` method that would verify if the file is not a file or is not writable (also check from L121)? Not existing file should not cause a failure in check (it could be like this for a new bot), but could be an info

kamil-certat (approved):

intelmq/bots/experts/asn_lookup/expert.py
sebix: The line is actually wrong. `is_file` is a method, not a member, therefore `is_file` always evaluates to true.

> An idea - maybe add also a check in the check method that would verify if the file is not a file or is not writable (also check from L121)?

I updated the existence check and added a type check

> Not existing file should not cause a failure in check (it could be like this for a new bot), but could be an info

I can follow you, but on the other hand, if the database file does not exist at the time of checking, the bot can also not operate with this current state and needs intervention by the admin to get into a working state.

kamil-certat: Ahh, good that you caught it!

> I can follow you, but on the other hand, if the database file does not exist at the time of checking, the bot can also not operate with this current state and needs intervention by the admin to get into a working state.

This is also true, so a failure may be expected. I'd anyway keep it more as warning as an error as it could be handled automatically, and is a desired state on the first run - until manual or automated action take place - and the bot configuration itself might be correct.

intelmq/bots/experts/asn_lookup/expert.py
```diff
@@ -66,12 +67,15 @@ def process(self):
...
+        database_path = Path(parameters.get('database', ''))
+        if not database_path.exists():
+            return [["error", f"File given as parameter 'database' ({database_path!s}) does not exist."]]
+        elif not database_path.is_file():
+            return [["error", f"Parameter 'database' ({database_path!s}) exists, but is not a file."]]
         try:
             pyasn.pyasn(parameters['database'])
         except Exception as exc:
-            return [["error", "Error reading database: %r." % exc]]
+            return [["error", f"Error reading database ({database_path!s}): %r." % exc]]
```
kamil-certat: Why mixing f-string and %? Let's pack `exc` inside the f-string

intelmq/bots/experts/asn_lookup/expert.py
```diff
@@ -66,12 +67,15 @@ def process(self):
 
     @staticmethod
     def check(parameters):
-        if not os.path.exists(parameters.get('database', '')):
-            return [["error", "File given as parameter 'database' does not exist."]]
+        database_path = Path(parameters.get('database', ''))
+        if not database_path.exists():
+            return [["error", f"File given as parameter 'database' ({database_path!s}) does not exist."]]
```
kamil-certat: Light suggestion: make it as warning
Strong suggestion: "...not exists. You may need to trigger first downloading manually. See: https://docs.intelmq.org/latest/user/bots/#asn-lookup" 

Otherwise we create an error message for eventually expected state, and people will think the bot is buggy

intelmq/bots/experts/asn_lookup/expert.py
```diff
@@ -66,12 +67,15 @@ def process(self):
 
     @staticmethod
     def check(parameters):
-        if not os.path.exists(parameters.get('database', '')):
-            return [["error", "File given as parameter 'database' does not exist."]]
+        database_path = Path(parameters.get('database', ''))
```
kamil-certat, line 71: What would you think about adding a one more check - how old the file is? And issue a warning if it's older than a week (I think it should be updated more often, right?)? Something like "The ASN database is older than 7 days and may be not up-to-date any more. Is automated database update working?"

kamil-certat (changes requested):

intelmq/bots/experts/asn_lookup/expert.py
kamil-certat: I recall we had once a problem with broken ASN update cronjob, and it resulted in wrong assigning and wrong addressing some events - so a check for that might help someone :)

sebix: Good idea, added it.

intelmq/bots/experts/asn_lookup/expert.py
sebix: > I'd anyway keep it more as warning as an error as it could be handled automatically, and is a desired state on the first run - until manual or automated action take place - and the bot configuration itself might be correct.

I disagree with you that the configuration is correct if it's not able to operate.

A better solution, solving your and my concerns, could be to run the auto update method at initialization if the database doesn't exist.
However, I'd like to postpone that feature in favor of getting the changes with improved checks and default values done.

sebix:
Hehe, the added check already works:

```
FAILED intelmq/tests/bots/experts/asn_lookup/test_expert.py::TestASNLookupExpertBot::test_static_bot_check_method - ValueError: checks is [['warning', 'Database (/home/runner/work/intelmq/intelmq/intelmq/tests/bots/experts/asn_lookup/ipasn.dat) is older than one week. Check the auto update, see: https://docs.intelmq.org/latest/user/bots/#asn-lookup.']]
```

I believe the line raising that exception is actually a bug.

EDIT: Addressed by #2622

kamil-certat (approved):

Issue #2586, opened 2025-03-20
[smtp_batch] Additional field for grouping


Lukas-Heindl:
Hi,

I have a usecase where I'd like to send batched E-Mails to someone with collected events for which I want to use the smtp_batch output bot. Then in this case it would be nice to break down the events for one contact further into multiple E-Mails. Would it be possible to add some option such that the *group-by* is not only done by `source.abuse_contact` but also by a field determined by the bot configuration?
Eventually it would also be nice to use this field also included in the *group-by* in e.g. the subject, but that's just for the future.

---

I'm willing to submit this as a PR myself, I just wanted to create a place for input and also most important to check beforehand whether this has any chance of being merged.

So far `smtp_batch` inserts each event as value into *redis* with the `bot_id + source.abuse_contact` as key:
https://github.com/certtools/intelmq/blob/a5f2f9c3499a02b0ac9aaf38a4e8ca13fd5cf769/intelmq/bots/outputs/smtp_batch/output.py#L77-L78

It then later uses `redis.keys()` to retrieve all keys with a certain prefix (`bot_id`):
https://github.com/certtools/intelmq/blob/a5f2f9c3499a02b0ac9aaf38a4e8ca13fd5cf769/intelmq/bots/outputs/smtp_batch/output.py#L224

So the most straight forward way would be to simply extend the redis-key with the value of the field that should be included in the logical *group-by*. But we'd need to be careful, the retrieved redis-key (here `mail_record`) is still used to obtain the destination E-Mail address:
https://github.com/certtools/intelmq/blob/a5f2f9c3499a02b0ac9aaf38a4e8ca13fd5cf769/intelmq/bots/outputs/smtp_batch/output.py#L282
In order to solve this I see two options
1. separate the values used to derive the redis-key with a special character like `|` -> in this case we'd need to make sure this character doesn't occur in any of the values -> replacing (but probably modifying the values is a bad idea) or escaping (a bit more complex) needed
[...]

PR #2608, opened 2025-04-24
smtp_batch: proper exit code when non-interactive

Lukas-Heindl:
This is only related to #2586 as that issue covers more than this PR. For the rest see the commit message.

### DIFF
--- a/intelmq/bots/outputs/smtp_batch/output.py
+++ b/intelmq/bots/outputs/smtp_batch/output.py
@@ -167,8 +167,12 @@ class SMTPBatchOutputBot(Bot):
                 sys.exit(0)
             elif i == "all":
                 count = 0
+                exit_code = 0
                 for mail in mails:
-                    if self.build_mail(mail, send=True):
+                    succ = self.build_mail(mail, send=True)
+                    if not succ:
+                        exit_code = 1
+                    else:
                         count += 1
                         print(f"{mail.to} ", end="", flush=True)
                         try:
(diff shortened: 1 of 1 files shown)


Lukas-Heindl:
From my side this would be already finished (if the tests pass, but I expect them to do so with that little changes). Is there anything left to do for me?

sebix:
The tests are broken, #2602 would fix that.

Do you mean _smtp_batch: proper exit code when **~~non~~**-interactive_?

Lukas-Heindl:
No I think it's right this way. When the bot is interactive, I assume the exit code is not that important since a human is reading the output anyhow (also how to set the exit code if the first command (e.g. sending the first debug mail failed) but the second command succeded?).

So this change is supposed to handle the case when the bot is run via cli, but e.g. as cron job (so not interactively but with the `--send` flag). In this case no human reads the output usually and the exit code can help in triggering an alert something has gone wrong (triggering a human to look into the issue).

> tests

So do I understand this correctly, are all PRs suspended at the moment, until that test-case is fixed?

Lukas-Heindl:
Now with the tests fixed, can you trigger the workflows on this?

sebix:
> No I think it's right this way. When the bot is interactive, I assume the exit code is not that important since a human is reading the output anyhow (also how to set the exit code if the first command (e.g. sending the first debug mail failed) but the second command succeded?).
> 
> So this change is supposed to handle the case when the bot is run via cli, but e.g. as cron job (so not interactively but with the `--send` flag). In this case no human reads the output usually and the exit code can help in triggering an alert something has gone wrong (triggering a human to look into the issue).

Thanks for the explanation. I confused (non-)interactive with the bot run mode (continuous/scheduled). Your logic makes sense of course :)

> (also how to set the exit code if the first command (e.g. sending the first debug mail failed) but the second command succeded?).

My 2 cents on this - but I'm not a user of this bot: If any error occurred, the exit code would be non-zero (or the maximum of all potential exit codes). But that's neither important nor urgent to do.

sebix (approved):

PR #2610, opened 2025-04-29
smtp_batch: add feature for grouping and templating

Lukas-Heindl:
Like discussed in #2586, here is the thing. I already did some manual testing regarding templating the subject string. For this configured
1. add `extra.my_template_value` to the events (I did this with the modify bot for now)
2. add that key (`extra.my_template_value`) to the new `additional_grouping_keys` parameter
3. use `subject: "Prefix {{ extra_my_template_value }}"`
4. activate the templating capabilites for the subject line with `templating: {subject: true}`

But probably some tests in the test suite will be needed for this (but I'd like to defer this one after the first feedback round)

Any comments?

Maybe issues to discuss:
- how values get hashed generically with `hash_arbitrary`
- passing the `template_data` along with each `mail` as a tuple
- really a warning if jinja2 couldn't be imported? (the user might not want jinja2) -- also whether to add this optional requirement to the `requirements.txt`?

~Also I didn't get the testing environment setup correctly until now, so it would be nice if someone could trigger the pipelines yet already so I can check the outputs regarding code-style.~

### DIFF
--- a/intelmq/bots/outputs/smtp_batch/output.py
+++ b/intelmq/bots/outputs/smtp_batch/output.py
@@ -8,7 +8,7 @@ import os
 import sys
 from tempfile import NamedTemporaryFile
 import time
-from typing import Any, Generator, Optional
+from typing import Any, Iterable, Optional, Dict, List
 import zipfile
 from base64 import b64decode
 from collections import OrderedDict
@@ -50,13 +50,13 @@ class Mail:
     to: str
     path: str
     count: int
-    template_data: dict[str,Any]
+    template_data: Dict[str, Any]
 
 
 class SMTPBatchOutputBot(Bot):
     # configurable parameters
     additional_grouping_keys: Optional[list] = []  # refers to the event directly
-    templating: Optional[dict[str, bool]] = {'subject': False, 'body': False, 'attachment': False}
+    templating: Optional[Dict[str, bool]] = {'subject': False, 'body': False, 'attachment': False}
     alternative_mails: Optional[str] = None
     bcc: Optional[list] = None
     email_from: str = ""
@@ -126,8 +126,6 @@ class SMTPBatchOutputBot(Bot):
             raise MissingDependencyError('envelope', '>=2.0.0')
         if jinja2 is None:
             self.logger.warning("No jinja2 installed. Thus, the templating is deactivated.")
-        if self.additional_grouping_keys and len(self.additional_grouping_keys) > 0:
-            self.additional_grouping_keys_trans = {self.fieldnames_translation.get(k, k) for k in self.additional_grouping_keys}
         self.set_cache()
         self.key = f"{self._Bot__bot_id}:"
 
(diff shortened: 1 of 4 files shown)


Lukas-Heindl:
Feel free to have a look, but it will be some time (estimate: 1-2 weeks) until I can have a look at the failing pipelines.
My way to go would be
1. looking into what makes them fail
   - code style should pass obviously
   - failing tests here might indicate a not fully backwards compatible change
2. feedback (of course any time, also before/while 1.)
3. Write additional tests covering the new features

intelmq/bots/outputs/smtp_batch/output.py
```diff
@@ -283,31 +327,40 @@ def prepare_mails(self):
...
             mail = Mail(mail_record, email_to, path, count)
-            self.build_mail(mail, send=False)
-            if count:
-                yield mail
+            # build_mail only used to output metadata of the mail -> send=False -> return None
+            self.build_mail(mail, send=False, template_data=template_data)
+            yield (mail, template_data if jinja2 and self.templating and any(self.templating.values()) else {})
 
-    def build_mail(self, mail, send=False, override_to=None):
+    def build_mail(self, mail, send=False, override_to=None, template_data:dict[str, Any]={}):
```
e3rd: Integrate `template_data` into the `Mail` dataclass instead of passing tuples.

Add annotations in the new code for the non-trivial objects (they did not exist when I wrote it originally), ex, `def build_mail(self, mail: Mail,`

e3rd (changes requested):
Seems good, thanks! :)

Lukas-Heindl:
Finally the tests are working. (Sorry for the frequent pushes)

So now from my point of view only some tests for the new features are missing. I'd like to keep the old/current tests as "basic" config with the "basic" set of features and one new test (different config is needed for this anyhow) for the more advanced features (one run which includes the new grouping and the templating).

Lukas-Heindl:
The test should be easily fixed now. But I noticed another issue (for which we do not have a test yet:
https://github.com/certtools/intelmq/blob/2f15e20422d350f6ac4cfa641d89712f7f306d7c/intelmq/bots/outputs/smtp_batch/output.py#L77-L78
When `source.abuse_contact` is multiple addesses it is important to use the address stored in the key and not the `soruce.abuse_contact` field.

The fix I thought about would be to simply add a number to the redis key serving as index which indicates which address of the `source.abuse_contact` field should be used in case there are multiple ones.

Any other ideas / comments on this approach?

Lukas-Heindl:
Ok I noted at least with the default `harmonization.yml`, `source.abuse_contact` shouldn't be a list. Still the old version also handled the list case. What do you think should we continue supporting that case? (and if yes what do you think about my current approach?)

Apart from that issue this is finished from my point of view (maybe cleanup the commit history even more). Any comments / changes required?

e3rd:
(sorry I'm late again, trying to get to you in the next days)

e3rd:
Whoa, it took me so long. Thanks for having investigated so much but let's stop supporting list and lets merge!

sebix:
> Whoa, it took me so long. Thanks for having investigated so much but let's stop supporting list and lets merge!

Thank you for the review, I rebased on develop, added changelog entries and will merge after the checks

Issue #1591, opened 2020-07-29
Package contrib tools


ghost:
Include the contrib tools in packages

- [ ] python packages
- [ ] debian packages
- [x] rpm packages

aaronkaplan:
Not clear, what you mean with this one. --verbose please.

wagner-certat:
There are a few "tools" in terms of executables in contrib:
- `bash-completion` (only missing in pypi packages)
- `check_mk`
- `development-tools`
- `elasticsearch`
- `eventdb`
- `feeds-config-generator`
- `malware_name_mapping`
- `prettyprint`
- `systemd`
- `vagrat`

those could be offered in separate optional packages

wagner-certat:
Further, the logcheck rules which are currently shipped, can be moved to a contrib package

sebix:
Solved by #2614

Issue #1772, opened 2021-03-02
Remove logrotate and use built-in log file rotate by default


wagner-certat:
Since #1637 IntelMQ can rotate log files by itself. Currently (optional) logrotate rules are used, but that is neither perfect nor a future-proof way (syslog and systemd journals are the more modern options).

I propose to get rid of the logrotate dependency.

- [ ] Come with meaningful default values for log rotation
- [ ] Remove logrotate from contrib and the packages
- [ ] Document that change in NEWS as well

aaronkaplan:
I get the point that logrotate rules are maybe not really so cool. However, I wonder if adding logrotation to intelmq itself isn't a bit like the systemd discussion: systemd tends to do everything. Like, add a DNS recursor to intelmq - why? because it might be more efficient, etc...
So... Not sure if it is good from an architectural design point to add this functionality but rather show how different *existing* logrotation approaches may be integrated / harnessed.

It breaks a bit the KISS and do one thing only principle...
See what I mean?

wagner-certat:
It's not about *adding* new functionality, but about changing the defaults. My intention here is that IntelMQ should have sane defaults so that users won't come into troubles (e.g. running out of disk space) and do not need to take additional measures to not get problems (setting up logrotate). In any case, a customization of the logging behavior is possible. Want to use syslog? Then change the logging handler to syslog. Want to use logrotate? Then disable rotation and set-up logrotate.

wagner-certat:
Revised suggestion:

For all *new* IntelMQ-installations via pip (any non-deb/rpm-installation), use the integrated log rotation by default. The log rotation can, at any time, be turned off, the logging documentation should show how this can be done. For all *existing* non-deb/rpm-installation, nothing changes (IntelMQ can differentiate between those cases because of the state-file).
For deb/rpm-installations, logrotate is automatically set-up. (this is the current behaviour).

A probably relevant reason, why I proposed this in the first-place: For pip-installations, we can't easily install other system-packages like logrotate and change it's configuration. For deb/rpm-packages, such behaviour is expected and supported.

PR #2614, opened 2025-05-07
pkg: package complete contrib dir in new separate package

sebix:
server all contrib/ contents in a new package `intelmq-contrib`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ Please refer to the [NEWS](NEWS.md) for a list of changes which have an affect o
 
 ### Packaging
 - Replace `/opt/intelmq` example paths in bots with variable `VAR_STATE_PATH` for correct paths in LSB-path setups like with packages (PR#2587 by Sebastian Wagner).
+- New deb-package `intelmq-contrib` with all `contrib/` scripts and documentation (PR#2614 by Sebastian Wagner).
 
 ### Tests
 - `intelmq.tests.lib.test_pipeline.TestAmqp.test_acknowledge`: Skip on all Python versions when running on CI (PR#2602 by Sebastian Wagner).
--- a/debian/intelmq-contrib.install
+++ b/debian/intelmq-contrib.install
@@ -1,0 +1,9 @@
+contrib/eventdb/ /usr/share/intelmq/contrib/
+contrib/example-extension-package/ /usr/share/intelmq/contrib/
+contrib/feeds-config-generator/ /usr/share/intelmq/contrib/
+contrib/logcheck/ /usr/share/intelmq/contrib/
+contrib/logrotate/ /usr/share/intelmq/contrib/
+contrib/malware_name_mapping/ /usr/share/intelmq/contrib/
+contrib/prettyprint/ /usr/share/intelmq/contrib/
+contrib/systemd/ /usr/share/intelmq/contrib/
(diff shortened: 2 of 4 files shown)


kamil-certat:
What is the reason about that?

sebix:
contrib contain several useful scripts and all users of packages would need to download, install and regularly update them purely manually. For example the malware name mapping and eventdb scripts. packaging them (separatly) makes it way easier for them to use, and also increases the visibility of these scripts.

sebix:
this will fix https://github.com/certtools/intelmq/issues/1591

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ Please refer to the [NEWS](NEWS.md) for a list of changes which have an affect o
 
 ### Packaging
 - Replace `/opt/intelmq` example paths in bots with variable `VAR_STATE_PATH` for correct paths in LSB-path setups like with packages (PR#2587 by Sebastian Wagner).
+- New deb-package `intelmq-contrib` with all `contrib/` scripts and documentation (PR#2614 by Sebastian Wagner).
```
kamil-certat, line 40: I don't see the documentation in the packaged files :thinking: man pages generated from mkdocs would be something cool

debian/intelmq-contrib.install
```diff
@@ -0,0 +1,9 @@
+contrib/eventdb/ /usr/share/intelmq/contrib/
+contrib/example-extension-package/ /usr/share/intelmq/contrib/
+contrib/feeds-config-generator/ /usr/share/intelmq/contrib/
+contrib/logcheck/ /usr/share/intelmq/contrib/
+contrib/logrotate/ /usr/share/intelmq/contrib/
```
kamil-certat, line 5: Before approval, could you please clarify two things to me?

The two contrib directories: `logrotate` and `malware_name_mapping` looks to me like something that eventually should be included in the main package, shouldn't it? I thought those two things should be installed by defaults (or maybe there are separated files?). For the rest, I have no doubts (tbh, I could even just skip packaging them at all)

CHANGELOG.md
sebix: > I don't see the documentation in the packaged files 🤔

all contents of contrib/ are part of the package in the same structure, including their READMEs.

debian/intelmq-contrib.install
sebix: > The two contrib directories: logrotate and malware_name_mapping looks to me like something that eventually should be included in the main package, shouldn't it? I thought those two things should be installed by defaults (or maybe there are separated files?).

Regarding logrotate:
Maybe it's more complex. If the administrator chooses not to use logrotate, but systemd logging?
The logrotate rules had a semi-official status. Although they are in contrib (unofficial stuff), we used them in the packages (basically "official").
This PR actually doesn't remove the logrotate config from the main package. It is only added additionally to the contrib package, duplicating the content. I removed that line.
See also #1772 

Regarding the malware name mapping scripts:
The maintenance and future of these scripts and the mapping itself are unclear, so I would have my concerns if we included them in the main package.
Follow-up question: Where do we draw the line between the main package and contributed stuff, without bloating the main package with small (potentially broken) scripts?

> For the rest, I have no doubts (tbh, I could even just skip packaging them at all)

The rest is not of importance to me either, but it was cheap to add them as well.

kamil-certat (approved):

PR #2615, opened 2025-05-08
bin: fix link to orphaned queues FAQ

sebix:


sebix:
The test fail is unrelated and caused by https://github.com/pymssql/pymssql/issues/937
As that only affects Python 3.8, where no wheel for pymssql is available and Python 3.8 is EOL anyway, let's do that: https://github.com/certtools/intelmq/pull/2617

kamil-certat (approved):

PR #2617, opened 2025-05-14
Drop support for Python 3.8

sebix:
it's EOL since 2024-10 and it's no longer in use by a supported plattform

Upgrade code with pyupgrade --py39-plus

The trigger for the change is https://github.com/certtools/intelmq/pull/2615 and https://github.com/pymssql/pymssql/issues/937

### DIFF
--- a/.github/workflows/scripts/setup-full.sh
+++ b/.github/workflows/scripts/setup-full.sh
@@ -15,10 +15,6 @@ sudo chown -R elasticsearch:elasticsearch /etc/default/elasticsearch
 sudo systemctl start elasticsearch
 
 sudo apt update
-if [ $python_version == '3.8' ]; then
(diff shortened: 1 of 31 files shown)


kamil-certat (approved):
You've missed the Python requirements in Debian package: https://github.com/certtools/intelmq/blob/2f15e20422d350f6ac4cfa641d89712f7f306d7c/debian/control#L28C1-L28C26

And please prepare similar changes for other components

@aaronkaplan I think you have to approve/change in the Repo config that CI for 3.8 is no longer requiered

sebix:
> You've missed the Python requirements in Debian package: https://github.com/certtools/intelmq/blob/2f15e20422d350f6ac4cfa641d89712f7f306d7c/debian/control#L28C1-L28C26

Thanks, added that change.

PR #2618, opened 2025-06-19
Added Shodan Alert API Collector

sebix:
Created in cooperation with Malawi CERT


sebix:
test will succeed with https://github.com/certtools/intelmq/pull/2619

intelmq/bots/collectors/shodan/collector_alert.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+
+    api_key: Optional[str] = None
+
+    def init(self):
+        if shodan is None:
+            raise ValueError("Library 'shodan' is needed but not installed.")
+
+        self.set_request_parameters()
+        if tuple(int(v) for v in pkg_resources.get_distribution("shodan").version.split('.')) <= (1, 8, 1):
```
kamil-certat: 1. Looks like `pkg_resources` is not imported
2. It's deprecated anyway, and will be gone soon, please use replacements

kamil-certat (changes requested):

intelmq/bots/collectors/shodan/collector_alert.py
kamil-certat: 3. Let's add this check to `check` method

sebix: IntelMQ uses `pkg_resources` a lot, so to get rid of it, we will need to change a few more code lines.

For the sake of consistency and not to introduce deviant code, I'd like to keep `pkg_resources` here.

sebix: > Let's add this check to `check` method

Good idea, but the static check method can't call the non-static `self.set_request_parameters()`  which processes and provides the proxy parameters :/

kamil-certat: But you don't need proxy to check the library version, just extract checking to a static method. Or is there a kind of import/patching race condition?

kamil-certat (approved):
If possible, I'd still suggest adding the check method with lib version check :)

sebix:
I don't yet know what causes the tests fail about the `check` method. Maybe I miss the forest for the trees.

EDIT: Yep, was exactly that :see_no_evil:

aaronkaplan (approved):
Initial review of the code: looks good , thanks!!

BUT... the tests fail.... and the worked again....

aaronkaplan (approved):

PR #2623, opened 2025-07-15
enh: collector_mail_attach Decrypt GPG attachments

e3rd:
Our partner sends us a GPG-encrypted attachment. This PR solves this.

Before merging, let's discuss. Should it be part of a collector bot, or should it be rather a GPG parser?

Current pipeline (current PR)

```mermaid
graph LR;
a["mail collector attach"]--> b["generic csv parser"];
```

Another possibility

```mermaid
graph LR;
a["mail collector attach"]--> c["gpg parser"] --> b["generic csv parser"];
```

Another possibility

```mermaid
graph LR;
a["mail collector attach"]--> b["generic csv parser allowing GPG decryption"];
```

Another possibility

```mermaid
graph LR;
a["mail collector attach"]--> b["gpg parser, inheriting from generic csv parser"];
```


sebix:
A separate OpenPGP parser (first variant) would strictly adhere to the rule "one bot one job", _but_ the current code is not designed for nor tested with two parsers in a row.
As you did it now, adding the decryption to the collector, is in line with other post-collection processing, such as extracting from archives, so that's fine for me.

Forth variant makes less sense than the third. In third, the CSV parser could use an "OpenPGP Mixin".

e3rd:
Great! Thanks for having investigated, sebix! :) I am then ready to be merged – or is there anything I could do better?

intelmq/bots/collectors/mail/collector_mail_attach.py
```diff
@@ -28,11 +29,19 @@ class MailAttachCollectorBot(MailCollectorBot):
     mail_user: str = "<user>"
     rate_limit: int = 60
     subject_regex: str = "<subject>"
+    decrypt: bool = False
+    """ Decrypt the attachment with GPG. """
```
kamil-certat: Would you mind changing those strings to comments? Or do you use an IDE that catches it as hint?

intelmq/bots/collectors/mail/collector_mail_attach.py
```diff
@@ -63,6 +72,13 @@ def process_message(self, uid, message):
                     raw_reports = ((attach_filename, attach['content'].read()), )
 
                 for file_name, raw_report in raw_reports:
+                    if self.decrypt:
+                        gpg = self._gpg().decrypt(raw_report, passphrase=self.gpg_passphrase)
+                        if gpg.ok:
+                            raw_report = gpg.data
+                        else:
+                            self.logger.error('Could not decrypt attachment %s: %s', file_name, gpg.status)
```
kamil-certat: Add `.` at the end of the log (...there are tests for that and they would fail if the log without `.` at the end have been triggered :D)

kamil-certat (approved):
I agree with the recommendation from @sebix I like the change, it looks good - however I'd slightly suggest adding negative test to ensure proper handling GPG errors

intelmq/bots/collectors/mail/collector_mail_attach.py
e3rd: ok, thanks

intelmq/bots/collectors/mail/collector_mail_attach.py
e3rd: In vscode I'm using right now, I see them as hints.

image

In the future, they might be more easily grabbed for autodocs to the manager

image

The docstrings are more easily parseable than comments. But ok, I stick with the current usage and convert them to comments now.

kamil-certat: If you see them in IDE, it's okay to me

kamil-certat:
Ah, unfortunately, you need to add license files for not code test files (see example https://github.com/certtools/intelmq/blob/develop/docs/static/images/Logo_Intel_MQ.png.license)

e3rd:
tests added, license solved :) I'm done now

.reuse/dep5
```diff
@@ -33,3 +33,15 @@ License:   AGPL-3.0-or-later
...
+Copyright: 2025 Edvard Rejthar, CSIRT.cz <edvard.rejthar+intelmq@nic.cz>
+License: AGPL-3.0-or-later
+
+Files: intelmq/tests/bots/collectors/mail/gpg_wrong_attachment.eml
+Copyright: 2025 Edvard Rejthar, CSIRT.cz <edvard.rejthar+intelmq@nic.cz>
+License: AGPL-3.0-or-later
+
+Files: intelmq/tests/bots/collectors/mail/gpg_ring/**
+Copyright: 2025 Edvard Rejthar, CSIRT.cz <edvard.rejthar+intelmq@nic.cz>
+License: AGPL-3.0-or-later
```
sebix: missing newline at end of file

intelmq/bots/collectors/mail/collector_mail_attach.py
```diff
@@ -28,11 +31,19 @@ class MailAttachCollectorBot(MailCollectorBot):
     mail_user: str = "<user>"
     rate_limit: int = 60
     subject_regex: str = "<subject>"
+    decrypt: bool = False
```
sebix: As OpenPGP is not the only e-mail encryption method (S/MIME is also widespread), I suggest naming the parameter `decrypt_openpgp` or similar.

```suggestion
    decrypt_openpgp: Optional[bool] = False
```

intelmq/bots/collectors/mail/collector_mail_attach.py
```diff
@@ -28,11 +31,19 @@ class MailAttachCollectorBot(MailCollectorBot):
...
+    """ The private key passhrase """
+
+    gpg_home: str = ""
+    """ Change the GPG home directory """
 
     def init(self):
         super().init()
         if self.attach_regex is None:
-            raise InvalidArgument('attach_regex', expected='string')
+            raise InvalidArgument("attach_regex", expected="string")
```
sebix: I don't see how this and other formatting changes are related to the decryption. It makes the diff harder to read and the blames harder to interpret and use.

intelmq/bots/collectors/mail/collector_mail_attach.py
```diff
@@ -80,5 +108,13 @@ def process_message(self, uid, message):
         self.logger.info("Email report read.")
         return seen
 
+    @cache
+    def _gpg(self):
+        try:
+            from gnupg import GPG
+        except ImportError:
+            raise MissingDependencyError("python-gnupg", ">=0.5")
```
sebix: I suggest doing the check in `init` (for quick user-feedback). If decyrption is activated, and the library is not installed, raise the exception.

sebix (changes requested):
Thanks for this extensive work, including nice test cases.

I'm missing the documentation though (in `docs/user/bots.md`).

intelmq/bots/collectors/mail/collector_mail_attach.py
e3rd: my bad!

intelmq/bots/collectors/mail/collector_mail_attach.py
e3rd: good point

intelmq/bots/collectors/mail/collector_mail_attach.py
e3rd: What about `decrypt_gpg` to be in line with other params? (If you mean it, decrypt_openpgp is alright.)

.reuse/dep5
e3rd: huh, magic

e3rd:
All should be done :)

intelmq/bots/collectors/mail/collector_mail_attach.py
sebix: The encryption standard is called OpenPGP. gnupg/`gpg` is the program we are using to decrypt it.

So I think `opengpg` is a better suffix, as the important information is the standard's name, not the name of the program that the bot uses internally.

Analogously, S/MIME is the standard, and `gpgsm` is one of the programs for en-/decryption.

e3rd: Seems legit. Should I then use opengpg for other params too? (Which is longer)

intelmq/bots/collectors/mail/collector_mail_attach.py
kamil-certat: (but we should move to ruff and standardize formatting anyway, I always have to revert formatting changes :P)

intelmq/bots/collectors/mail/collector_mail_attach.py
kamil-certat: I'd switch only `_passphrase` as `gpg_home` is related only to the gnupg itself. In my opinion `_openpgp` sounds generally better, but `_gpg` is also acceptable :sweat_smile:

e3rd: So we have now three attributes:

* decrypt_openpgp
* openpgp_passphrase
* gpg_home

Good luck for the users who are not accustomed to PGP nomenclature maze :D . But I admit it's right now.

sebix: On the first glance, it looks awful, but glancing at it for five minutes, it all makes sense xD
Luckily, they are grouped in the docs.

If the parameters were hierarchical, they looked more sensible:

- decrypt
  - openpgp
    - enabled: yes
    - passphrase: xxx
    - gpg_home: yyy

The more I think of it, `openpgp_passphrase` is also ambiguous, as OpenPGP allows for asymmetric and symmetric encryption, so it could be the passphrase for symmetric decryption.
(If the parameter was also called `gnupgp_passphrase`, that's again ambiguous, as that could also mean a passphrase for unlocking gnupg itself, eg, the keyring or whatever).

So, finally `openpgp_key_passphrase`?
Advantage: If in the future someone adds a parameter to select a specific key, it would be called `openpgp_key_id` or `openpgp_key_file`, etc.

I hope that's my last thought on this topic :sweat_smile:

e3rd: Hey, I might make them hierarchical, that would be the best!

```
"decrypt.openpgp": bool
"decrypt.openpgp.key_passphrase": str
"decrypt.openpgp.gpg_home": Path
```

Isn't it the best?

IMO, the key cannot be chosen, at least by GnuPG. It either is in the keyring, or not. (For encrypting, we choose keys, not for decrypting.)

sebix:
Waiting for input on the parameter scheme from @aaronkaplan or @kamil-certat

aaronkaplan (approved):

PR #2634, opened 2025-08-12
http mixin: only send auth when strings not empty

### LINKED ISSUE #2590: HTTP: If username and password are empty strings (default), Authorization is still used
if `http_username` and `http_password` are set to an empty string, as it's the default with IntelMQ Manager, then the collector does send an Authorization header:
```
collector: Runtime configuration: parameter 'http_password' loaded with value 'HIDDEN'.
collector: Runtime configuration: parameter 'http_username' loaded with value ''.
....
os-collector: Request headers: {'User-Agent': 'python-requests/2.32.3', 'Accept-Encoding': 'gzip, deflate, br, zstd', 'Accept': '*/*', 'Connection': 'keep-alive', 'Authorization': 'Basic Og=='}.
```

For endpoints which evaluate the header also, if not required, this triggers a 403.

sebix:
Only set HTTP header 'Authorization' if username or password are set and are not both empty string as they are by default in the Manager

fixes https://github.com/certtools/intelmq/issues/2590

### DIFF
--- a/intelmq/lib/mixins/http.py
+++ b/intelmq/lib/mixins/http.py
@@ -65,9 +65,12 @@ class HttpMixin:
         # tls certificate settings
         if self.ssl_client_cert is not None:
             self.__session.cert = self.ssl_client_cert
-        # auth settings
-        if self.http_username is not None:
+        # auth settings. username or password must exist (if only one, then this is likely an error, but we should try without auth) and not be an empty string
+        if self.http_username and self.http_password:
             self.__auth = (self.http_username, self.http_password)
+        elif self.http_username and self.http_password:
+            # only one, but not both are given
+            self.logger.warning("Either 'http_username' or 'http_password' are given, but for HTTP Authentication, both must be set.")
         self.__session.auth = self.__auth
         # headers settings
         if self.http_header is not None:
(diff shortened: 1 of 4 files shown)


intelmq/lib/mixins/http.py
```diff
@@ -65,8 +65,8 @@ def setup(self):
         # tls certificate settings
         if self.ssl_client_cert is not None:
             self.__session.cert = self.ssl_client_cert
-        # auth settings
-        if self.http_username is not None:
+        # auth settings. username or password must exist (if only one, then this is likely an error, but we should try without auth) and not be an empty string
+        if self.http_username or self.http_password:
```
aaronkaplan: would it work only with a pwd? (no username?)

aaronkaplan (changes requested):
not quite sure.... I think http auth need *both* username and pwd

intelmq/lib/mixins/http.py
sebix: Not sure if I understand. What change do you request here?

aaronkaplan: Well a pure password in http auth without username is meaningless.. no? How do HTPT servers react?

sebix: Changed to `and` and added a warning if only one is given

aaronkaplan: thanks :)

sebix:
Now also added a positive check for the Authorization header.

sebix:
@aaronkaplan this PR is waiting for your approval

kamil-certat (approved):
This is actually a workaround for Manager issue, just btw

sebix:
> This is actually a workaround for Manager issue, just btw

The manager bug triggers this, but it could also be a result of a bad manual edit of the config files

PR #2638, opened 2025-08-14
Add autostart package

sebix:
adds a systemd timer and service and a shell script to call `intelmqctl
start` on system boot and every five minutes

raise debian compat level 10 for better and automatic dh-systemd support
https://wiki.debian.org/Teams/pkg-systemd/Packaging#Using_debhelper_with_dh_systemd


sebix:
Is already available in the unstable repository

aaronkaplan (approved):
looks good, easy. Clear

PR #2647, opened 2025-09-18
enh: collector_mail_attach allow_empty

e3rd:
Adding a tag. I keep receiving empty attachments from a partner. Since this is a valid behaviour and there is no reason for a bot to stop, I've added the parameter.


sebix:
I was thinking about the default value of the new parameter and about the behaviour itself.

I think I agree with you, that normally an empty attachment is an error. An error in parsing, the bot configuration or by the sender and needs admin attention. In the parsers, we also don't silently accept a misformatted report.

docs/user/bots.md
```diff
@@ -494,6 +494,10 @@ The fields can be used by parsers to identify the feed and are not automatically
 
 (optional, string) The OpenPGP private key passhrase.
 
+**`allow_empty`**
+
+(optional, boolean) Allow the attachment to be empty. If False, an error is raised and bot stays stopped.
```
sebix:
```suggestion
(optional, boolean) Allow the attachment to be empty. If False (default), an error is raised and bot stays stopped.
```

intelmq/bots/collectors/mail/collector_mail_attach.py
```diff
@@ -88,6 +91,11 @@ def process_message(self, uid, message):
                         else:
                             self.logger.error('Could not decrypt attachment %s: %s.', file_name, gpg.status)
                             continue
+
+                    if not raw_report and self.allow_empty:
+                        self.logger.info("Email report read (empty).")
```
sebix: Also log the action (ignored):
```suggestion
                        self.logger.info("Empty email report ignored.")
```

intelmq/tests/bots/collectors/mail/lib.py
```diff
@@ -59,3 +60,7 @@ def messages(self, *args, **kwargs):
 class MockedGpgAttachmentImbox(MockedImbox):
     def messages(self, *args, **kwargs):
         yield 0, deepcopy(EMAIL_GPG_ATTACHMENT)
+
+class MockedEmptyTextAttachmentImbox(MockedImbox):
+    def messages(self, *args, **kwargs):
+        yield 0, deepcopy(EMAIL_TEXT_ATTACHMENT_EMPTY)
```
sebix: missing newline at the end

sebix (changes requested):

e3rd:
>  I agree with you, that normally an empty attachment is an error. 

I would have not known either but this was the default behaviour. Thanks, I'll change!

e3rd:
If merging, can you squash by the button so that I don't have to rebase here? (I still may.)

sebix (approved):

PR #2650, opened 2025-09-19
ECS expert

sebix:


aaronkaplan:
I assume you mean the  Elastic Common Schema ?

sebix:
Yes. This is an old unfinished code from 2021. It still needs some more work, especially unit tests

PR #2667, opened 2025-11-06
Drop privileges before creating temporary directory

creideiki:
Otherwise, when the tests are run as root, the temporary directory gets created as root, but the cleanup code runs with dropped privileges, so it cannot remove it.


creideiki:
This makes no practical difference in GitHub CI, since its runners are regular users. However, we want to run tests in our internal CI as well, and there, the runners are root.

kamil-certat (approved):
LGTM

PR #2672, opened 2025-11-30
install docs: fix bug and simplify

sebix:
one command for Debian referenced a path for Ubuntu, which caused a 404

remove the necessity for gnupg (which is not installed by default)


sebix:
Just stumbled over that again. Currently, the install docs on Debian are broken.

As this is a small fix and rather important, I'm now merging that myself.

PR #2687, opened 2026-02-17
Fix Modify bot regex matching for numeric fields; add regression test

### LINKED ISSUE #2686: Modify bot: regex not matching without ()
In a Modify bot configuration, a condition like

```json
"if": {
    "field": "^12\\.[34]"
```

does not match values like "12.3" while

```json
"if": {
    "field": "^(12\\.[34])"
```
does.

This looks like a bug with regex handling.

sofiathefirst2323:
Fixes #2686
## Summary
Fix ModifyExpertBot regex matching so regex rules are always applied to the string version of the field value. This prevents cases where regex matching can behave differently depending on value type.

## What changed
- ModifyExpertBot: always run `rule.search(str(event[name]))` for regex conditions.
- Added a regression test to ensure regex without parentheses matches same as with parentheses.

## Testing
- pytest -q intelmq/tests/bots/experts/modify/test_expert.py

### DIFF
--- a/intelmq/bots/experts/modify/expert.py
+++ b/intelmq/bots/experts/modify/expert.py
@@ -71,18 +71,10 @@ class ModifyExpertBot(ExpertBot):
             if name not in event:
                 return None
             if is_re_pattern(rule):
-                if isinstance(event[name], (int, float)):
                     match = rule.search(str(event[name]))
                     if match is None:
                         return None
-                    else:
-                        matches[name] = match
-                else:
-                    match = rule.search(event[name])
-                    if match is None:
-                        return None
-                    else:
-                        matches[name] = match
+                    matches[name] = match
             else:  # rule is boolean, int, float, etc
                 if event[name] != rule:
                     return None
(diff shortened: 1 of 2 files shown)


sebix:
@sofiathefirst2323 Thanks for your initiative!

The test failures on Python 3.12 and 3.13 are not related to this PR, see #2685 
However, the codestyle failure is:
`intelmq/bots/experts/modify/expert.py:74:21: E117 over-indented`

In the part of the code you are working on, I see that the current behaviour is: if the field in the event is of type integer or float (which can be the case for a version number), the matching behaviour is slightly different. I assume that was there for a reason, but it may as well be obsolete now. We need to make sure that we are not unintentionally breaking compatibility here (and if so, that we make it intentionally).

We may lack test coverage for comparisons between integer/float types.
Could you please add a unit test for the exact example _th_ showed (which could be added to the existing `types.conf`) that demonstrates that the current code fails, so we have a complete end-to-end test?

Issue #2666, opened 2025-11-03
Error on running the smtp_batch output bot with intelmq 3.5.0


Lukas-Heindl:
Hi,

I checked out the recent version (3.5.0) sadly with this I get
```bash
$ intelmq.bots.outputs.smtp_batch.output --cli smtp-batched-output
smtp-batched-output: SMTPBatchOutputBot initialized with id smtp-batched-output and intelmq 3.5.0 and python 3.11.2 (main, Apr 28 2025, 14:11:48) [GCC 12.2.0] as process 945100. Standalone mode: False.
smtp-batched-output: Bot is starting.
smtp-batched-output: Loading source pipeline and queue 'smtp-batched-output-queue'.
smtp-batched-output: Connected to source queue.
smtp-batched-output: No destination queues to load.
smtp-batched-output: Bot initialization completed
Traceback (most recent call last):
  File "/opt/intelmq/venv/bin/intelmq.bots.outputs.smtp_batch.output", line 7, in <module>
    sys.exit(BOT.run())
             ^^^^^^^^^
  File "/opt/intelmq/venv/lib/python3.11/site-packages/intelmq/bots/outputs/smtp_batch/output.py", line 139, in run
    [setattr(instance, k, v) for k, v in vars(parsed_args).items()]
  File "/opt/intelmq/venv/lib/python3.11/site-packages/intelmq/bots/outputs/smtp_batch/output.py", line 139, in <listcomp>
    [setattr(instance, k, v) for k, v in vars(parsed_args).items()]
     ^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: property 'bot_id' of 'SMTPBatchOutputBot' object has no setter
smtp-batched-output: Bot stopped.
smtp-batched-output: Bot stopped.
```

I think this might be related to the addition of `bot_id` as property in https://github.com/certtools/intelmq/commit/d3864555b7579ef8be08939bcb1db88bea4e5b33 / https://github.com/certtools/intelmq/pull/2509

But I'm not too familiar with the internals of the bots, so I'm not sure if writing a setter for `bot_id` is really the right fix (or somehow ignore that field in the `[setattr(instance, k, v) for k, v in vars(parsed_args).items()]`)

kamil-certat:
Hey,

[...]

PR #2692, opened 2026-03-04
fixup: smtp_batch: make compatible with introduction of bot_id

Lukas-Heindl:
The `bot_id`was introduced as property in #2509 which broke the output bot
This issue/incompatibility was noted in #2666.

Ideally we would had a test which would have spotted this issue prior to releasing the last version. But looking at the current tests, I'm unsure if we can cover this correctly (running an executable is not that nice for testing).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ Please refer to the [NEWS](NEWS.md) for a list of changes which have an affect o
 #### Experts
 
 #### Outputs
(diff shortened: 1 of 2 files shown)


sebix:
> running an executable is not that nice for testing

maybe not nice, but doable. it doesn't need to run always, e.g. marked as exotic, then it's normally not executed locally, but definitely in the CI.

PR #2714, opened 2026-07-07
Handle bracketed URL netloc sanitization

### LINKED ISSUE #2382: Regression on parsing invalid URLs
As a continuation of #2377, we have a regression on parsing invalid URLs. Previously, the `urllib` was mach more liberal in processing URLs, now it rejects much more cases.

We use it for sanitize the URLs, and `html_parser` is an example of bot that uses the liberal behavior in tests:

https://github.com/certtools/intelmq/blob/61c45acfb8cc60e1419abe7c57691561ef9ee072/intelmq/tests/bots/parsers/html_table/test_parser_column_split.py#L47

https://github.com/certtools/intelmq/blob/61c45acfb8cc60e1419abe7c57691561ef9ee072/intelmq/tests/bots/parsers/html_table/test_parser_column_split.py#L73-L80

In patched Python versions (e.g. 3.11.4), this URL is rejected. We need to either decide against allowing such URLs, or redesign our sanitization. 

Temporally, the test is skipped to unlock other work.

0xTaoZ:
## Summary
- restore URL validation for legacy feed URLs whose network location contains bracketed markers rejected by modern urllib
- keep the existing `urlsplit` path for normal URLs and use a small fallback only when `urlsplit` raises `ValueError`
- unskip the HTML table parser regression test for issue #2382 and add direct URL harmonization coverage

Closes #2382.

## Validation
- `git diff --check`
- `INTELMQ_TEST_EXOTIC=1 .venv/bin/python -m pytest -q -o addopts='' intelmq/tests/bots/parsers/html_table/test_parser_column_split.py intelmq/tests/lib/test_harmonization.py intelmq/tests/lib/test_message.py`
- `.venv/bin/python -m compileall -q intelmq/lib/harmonization.py intelmq/tests/lib/test_harmonization.py intelmq/tests/bots/parsers/html_table/test_parser_column_split.py`

## Risk
Low scope: the fallback only runs when `urllib.parse.urlsplit()` raises `ValueError`, while ordinary URL handling continues through the existing parser-backed path.

### DIFF
--- a/intelmq/lib/harmonization.py
+++ b/intelmq/lib/harmonization.py
@@ -1085,6 +1085,15 @@ class URL(String):
     """
 
     @staticmethod
+    def _has_netloc(value: str) -> bool:
+        try:
+            result = parse.urlsplit(value)
+        except ValueError:
+            match = re.match(r'^[A-Za-z][A-Za-z0-9+.-]*://([^/?#]+)', value)
+            return bool(match and match.group(1))
+        return result.netloc != ""
+
+    @staticmethod
     def is_valid(value: str, sanitize: bool = False) -> bool:
         if sanitize:
             value = URL.sanitize(value)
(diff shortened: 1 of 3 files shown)


sebix:
See my comment in #2382

